PR #3, opened 2017-07-19
Dump former MessageKit repo contents

jessesquires:
This is a dump from the previously-deleted MessageKit repo. This may or may not be useful. I'll let you guys decide 😄 

Includes:

- A bit more basic infra setup
- Example project setup that already imports the framework
- travis ci initial setup (for testing example + lib)
- codecov initial setup
- more conventional directory naming (inline with SPM naming conventions)


Sources/MessagesDataSource.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    public let placeholderImage: UIImage
+
+    public init(image: UIImage? = nil, highlightedImage: UIImage? = nil, placeholderImage: UIImage) {
+        self.image = image
+        self.highlightedImage = highlightedImage
+        self.placeholderImage = placeholderImage
+    }
+
+    public func image(highlighted: Bool) -> UIImage {
```
SD10, line 49: This is to remind me to make the change currently on master

PR #5, opened 2017-07-22
v0.1.0

SD10:


Sources/MessageBubble.swift
```diff
@@ -0,0 +1,36 @@
...
+import Foundation
+
+public struct MessageBubble {
+    
+    public let image: UIImage
+    
+    public let highlightedImage: UIImage
+    
+    public func image(highlighted: Bool) -> UIImage {
+        return highlighted ? image : highlightedImage
```
SD10: This will probably be removed/repurposed considering we most likely won't use images for bubbles. The return values are flipped though 😛

Sources/MessageData.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    case audio(Data)
+    
+    case location(CLLocation)
+    
+    case photo(UIImage)
+    
+    case video(file: NSURL, thumbnail: UIImage)
+    
+    case system(String)
```
bojanstef: What is `MessageData.system`? or rather; what is the difference with `.text` and `.system`

towry: I think a little code comment will be helpful for people to understand.

SD10: I'm actually not sure because this is code leftover from the old repo. The only thing to be concerned about for the moment is `MessageData.text`. v1 should be basic messages and then expand to Media data types from there.

@towry eventually the code will be fully commented by v1. right now I'm just hacking things together 😅

SD10:
Some things are shaping up but the code is still very rough. A lot is subject to change. I'm struggling over using a UICollectionViewController opposed to a regular UIViewController. I'd like my collectionView to be typed as `MessagesCollectionView`.

Type casting everywhere :sob:

Sources/MessageData.swift
MacMeDan: IMO I would suggest removing anything that does not make any since. I would rather not have things then to start out with bloat right off the bat. But that might just be me.

Sources/MessageBubble.swift
MacMeDan: So why don't we remove this now? I feel like this a key point in this projects development where we have an opportunity to remove any cruft.

Sources/MessageType.swift
```diff
@@ -0,0 +1,36 @@
...
+
+public protocol MessageType {
+
+    var sender: Sender { get }
+
+    var messageId: String { get }
+
+    var sentDate: Date { get }
+
+    var data: MessageData { get }
```
MacMeDan, line 35: I don't know why we would not have the `sentDate` as a part of the `messageData`. Can we think of a reason we would want this?

MacMeDan (commented):
Remove anything that is not understood or that is not fully implemented.

SD10:
v0.1.0 Features:
- [x] Message Bubbles Incoming/Outgoing
- [x] Message Avatars
- [x] Basic Text Messages Support
- [x] Resizing Messages Based on Text Length
- [x] Input Bar With Send Button
- [x] Move Input Bar When Showing Keyboard

Sources/MessageCollectionViewCell.swift
```diff
@@ -0,0 +1,89 @@
...
+    // MARK: - Methods
+    
+    func setupSubviews() {
+        addSubview(messageContainerView)
+        addSubview(avatarImageView)
+    }
+    
+    override func apply(_ layoutAttributes: UICollectionViewLayoutAttributes) {
+        guard let attributes = layoutAttributes as? MessagesCollectionViewLayoutAttributes else { return }
+        print(attributes.avatarSize)
```
ArtSabintsev: For print statements, you should probably forward them to a print statement that only works in `DEBUG` mode.

SD10: @ArtSabintsev I left those by accident 👍 Thanks. I would never allow them in a release

towry: I guess the print statements is temporary here, I would add a comment to such thing as to be removed. Even in debug mode, other people may not want to see all kind print messages in the console. But it is good to decorate print statement in a DEBUG util in case the code being merged to master branch or we can use swift lint tool to check that.

SD10: All the code will be cleaned up before any merge to master. Having this PR open is just to show the public that there is work being done.

Sources/MessagesCollectionViewLayoutAttributes.swift
```diff
@@ -0,0 +1,70 @@
...
+        return copy
+    }
+    
+    override open func isEqual(_ object: Any?) -> Bool {
+        guard let attributes = object as? MessagesCollectionViewLayoutAttributes else { return false }
+        
+        return attributes.messageFont == messageFont && attributes.avatarSize == avatarSize &&
+            attributes.avatarToEdgePadding == avatarToEdgePadding && attributes.avatarBottomPadding == avatarBottomPadding &&
+            attributes.avatarToContainerPadding == avatarToContainerPadding && attributes.messageLeftRightPadding == messageLeftRightPadding &&
+            attributes.direction == direction
```
towry: Better formatting.

```swift
return attributes.messageFont == messageFont &&
    attributes.avatarSize == avatarSize &&
    attributes.avatarToEdgePadding == avatarToEdgePadding &&
    attributes.avatarBottomPadding == avatarBottomPadding &&
    attributes.avatarToContainerPadding == avatarToContainerPadding &&
    attributes.messageLeftRightPadding == messageLeftRightPadding &&
    attributes.direction == direction
```

Sources/MessageType.swift
lzyraining: I also get confused with naming. Here "MessageData" is the enum of different message category, while the protocol is holding the message content. It's better to name the protocol as "MessageData" and the enum as "MessageType".

Sources/MessageData.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    case attributedText(NSAttributedString)
+    
+    case audio(Data)
+    
+    case location(CLLocation)
+    
+    case photo(UIImage)
+    
+    case video(file: NSURL, thumbnail: UIImage)
```
lzyraining: Will the video type apply a UIButton to as its internal property? Or is it autoplay when video is sent?

Sources/AvatarSizeCalculator.swift
```diff
@@ -0,0 +1,37 @@
...
+ */
+import Foundation
+
+class AvatarSizeCalculator {
+
+    func avatarSizeFor(messageType: MessageType, with layout: MessagesCollectionViewFlowLayout) -> CGSize {
+
+        guard let messagesCollectionView = layout.collectionView as? MessagesCollectionView else { return .zero }
+        guard let isOutgoingMessage = messagesCollectionView.messagesDataSource?.isFromCurrentSender(message: messageType) else { return .zero }
+
```
lzyraining: It always returns an outgoingMessage. Incoming message won't enter next line.

Sources/MessageCollectionViewCell.swift
```diff
@@ -0,0 +1,89 @@
...
+        switch attributes.direction {
+        case .incoming, .outgoing:
+            let avatarX = attributes.avatarToEdgePadding
+            let avatarY = contentView.frame.height - attributes.avatarBottomPadding - attributes.avatarSize.height
+            avatarImageView.frame = CGRect(x: avatarX, y: avatarY, width: avatarWidth, height: avatarHeight)
+            
+            let containerX = contentView.frame.width - attributes.messageLeftRightPadding - containerWidth
+            let containerY: CGFloat = 0
+            messageContainerView.frame = CGRect(x: containerX, y: containerY, width: containerWidth, height: containerHeight)
+        }
```
lzyraining: The logic from the AvatarSizeCalculator is blocking incoming avatar. And there is no difference between incoming and outgoing case. Recommend to add a todo list to remove this constraint.

Sources/MessagesViewController.swift
```diff
@@ -1,28 +1,146 @@
...
+    
+    func setupSubviews() {
+        view.addSubview(messagesCollectionView)
+        view.addSubview(messageInputBar)
+    }
+    
+    func setupConstraints() {
+
+        messagesCollectionView.translatesAutoresizingMaskIntoConstraints = false
+        view.addConstraint(NSLayoutConstraint(item: messagesCollectionView, attribute: .top, relatedBy: .equal, toItem: topLayoutGuide, attribute: .bottom, multiplier: 1, constant: 0))
```
nunogoncalves: Sorry, just here sniffing around. 
Why not using anchors? Minimum targets?

Sources/AvatarSizeCalculator.swift
```diff
@@ -0,0 +1,37 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+import Foundation
+
+class AvatarSizeCalculator {
+
+    func avatarSizeFor(messageType: MessageType, with layout: MessagesCollectionViewFlowLayout) -> CGSize {
```
nunogoncalves: If I may, 
`func avatarSize(for messageType: MessageType, with layout: MessagesCollectionViewFlowLayout)` would make the calling site read better. :)

Sources/MessageContainerSizeCalculator.swift
```diff
@@ -0,0 +1,55 @@
...
+            // TODO: Heights for other types of message data
+            fatalError("Currently .text(String) is the only supported message type")
+        }
+    }
+
+}
+
+fileprivate extension String {
+    
+    func height(considering width: CGFloat, font: UIFont) -> CGFloat {
```
nunogoncalves: Just another tiny suggestion: 

```swift 
func height(considering width: CGFloat, and font: UIFont)
```

would produce a calling site reading experience a bit smoother.
 
```swift
height(considering: width, and: UIFont.systemFont...)
```

Also, this is nice for views with number of lines = 0. If you want to have a label with max number of lines = 2 for example, this won't do it as it will only consider the text. That way I would prefer the sizeThatFits from the label.

SD10:
Hi @ everyone, first and foremost sorry for the delay! I ran into a nasty bug that ate up most of my time. 

I'm sure that most of you have gotten bored looking for some way to contribute, but let's not take this PR too seriously. It was primarily to show people this repo is active.

It's a WIP so you may see some hacky code or half implemented solutions. It doesn't mean I have the intention of leaving them this way 😅

Sources/MessagesViewController.swift
SD10: Yes, minimum targets. I struggled with the decision to support iOS 8. I figured why not unless it becomes a hassle.

Sources/MessageCollectionViewCell.swift
SD10: It's half implemented, not done yet. This branch is really outdated.

Sources/MessageData.swift
SD10: These aren't commented out yet but v1 will only support basic types like `text(String)`. More details to come when I write up a Vision.md

Sources/MessageType.swift
SD10: I had similar concerns. I'm considering:
`MessageType` -> `MessageRepresentable`
`MessageData` -> `MessageType`

Example/Sources/ConversationViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        return messages.count
+    }
+    
+    func messageForItem(at indexPath: IndexPath, in collectionView: UICollectionView) -> MessageType {
+        return messages[indexPath.section]
+    }
+    
+    func avatarForMessage(_ message: MessageType, at indexPath: IndexPath, in collectionView: UICollectionView) -> Avatar {
+        let image = isFromCurrentSender(message: message) ? #imageLiteral(resourceName: "Steve-Jobs") : #imageLiteral(resourceName: "Tim-Cook")
+        return Avatar(placeholderImage: image)
```
lzyraining, line 102: Why not let Sender own the Avatar? To make the API more user friendly. We don't need to map the avatar to message manually.

 ```
func avatarForMessage(_ message: MessageType, at indexPath: IndexPath, in collectionView: UICollectionView) -> Avatar {
        return message.sender.avatar
}
```

SD10: Hey @lzyraining thanks for the feedback!

Deciding how to expose an API that allows for easy customization will be a difficult task. I hope to release a really basic v0.1.0 beta in the next few hours and will be looking for feedback/suggestions.

As for your suggestion, I don't think `Sender` is a good fit for the responsibility of Avatar because some people may not even want Avatars. I'd like to keep this types responsibility minimal. I've even considered making it a protocol.

lzyraining: @SD10 I agree that it's difficult to build easy customized API. Thanks for the optional type in swift, we can simple handle who don't want Avatars by putting a question mark.

I like the idea using protocol for some models. Swift is basically protocol oriented programming language. It's amazing that protocol allows inheritance and extension to make customize and default values easy. But my concern is that we are build a framework for all potential use. From a user's perspective, I don't want to build all the models which are declared as protocol in the library. I just want to it easy to use.

SD10: @lzyraining Thanks, I had considered making `Avatar` optional here as well. It will take a bit of time to decide on a nice API. The great thing is we don't have to worry about API stability at this moment 😃 

Thanks for raising the concerns of having to build the models required by protocols. I was originally thinking that people would probably already have a `User` model that they could conform `Sender` to -- but I see this could also be difficult and intrusive.

PR #7, opened 2017-07-24
v0.1.0: Minor code changes.

LucianoPAlmeida:
Just some minor changes

### DIFF
--- a/Sources/Avatar.swift
+++ b/Sources/Avatar.swift
@@ -40,10 +40,9 @@ public struct Avatar {
     
     public func image(highlighted: Bool) -> UIImage {
         
-        switch highlighted {
-        case true:
+        if highlighted {
             return highlightedImage ?? image ?? placeholderImage
-        case false:
+        } else {
             return image ?? placeholderImage
         }
         
--- a/Sources/MessagesViewController.swift
+++ b/Sources/MessagesViewController.swift
@@ -53,10 +53,7 @@ open class MessagesViewController: UIViewController {
         setupSubviews()
         setupConstraints()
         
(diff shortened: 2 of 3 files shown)


SD10:
Hey @LucianoPAlmeida thanks for the PR.

 Let's hold off on PR's on the moment though. I have a massive amount of hacked together code locally and dealing with merge conflicts could be troublesome 😅

Sources/Avatar.swift
```diff
@@ -40,10 +40,9 @@ public struct Avatar {
     
     public func image(highlighted: Bool) -> UIImage {
         
-        switch highlighted {
```
SD10, line 43: I'm one of those developers that doesn't like if statements for no good reason :P I try to avoid them at all costs!

Sources/MessagesViewController.swift
```diff
@@ -74,6 +71,13 @@ open class MessagesViewController: UIViewController {
         view.addSubview(messageInputBar)
     }
     
+    func setupMessagesCollectionView() {
```
SD10, line 74: Good idea. I'll definitely do things like this as some gardening.

Sources/Sender.swift
```diff
@@ -43,3 +43,10 @@ extension Sender: Equatable {
         return left.id == right.id
     }
 }
+
+// MARK: - Hashable Conformance
+extension Sender: Hashable {
```
SD10, line 48: This is probably a bit premature but not a bad idea. I'd probably take things a bit further by XORing all the underlying hash values.

towry:
I guess it's better to focus on the code reviews on the first PR to see what we can improve before having so many different code changes.

SD10:
Thanks @LucianoPAlmeida but closing this 🎉 👍

Issue #19, opened 2017-07-31
[Bug] Layout is off when in landscape mode


pinchih:
Not sure if MessageKit will support landscape mode in the future, but it's not displayed correctly at this point. ( v0.1.0 )

![simulator screen shot jul 31 2017 3 37 28 pm](https://user-images.githubusercontent.com/12094516/28801568-01aa4922-7607-11e7-8d63-4e91da80542c.png)
# #

SD10:
Definitely a priority 👍 Thank you

SD10:
This should be taken care of by `bcbdf4b` on `v0.2.0`. 
Also, referencing #12 since this was discussed there.

PR #20, opened 2017-08-01
Fix for #19

lzyraining:
screen shot 2017-07-31 at 11 23 09 pm

Fixing the overlapping when change device orientation.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,6 @@
+line_length:
+  warning: 150
+  ignores_comments: true
+
+disabled_rules:
+  identifier_name
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -26,39 +26,39 @@ import UIKit
 import MessageKit
 
 class ConversationViewController: MessagesViewController, MessagesDataSource, MessagesDisplayDataSource {
-    
+
     var messages: [MessageType] = []
-    
+
     override func viewDidLoad() {
         super.viewDidLoad()
-        
+
         addSampleData()
-        
+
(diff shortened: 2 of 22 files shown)


SD10:
@lzyraining what device are you using?

lzyraining:
@SD10 iPhone 7 plus simulator

SD10:
@lzyraining Good catch. Can this be done in `viewWillTransition(to:coordinator:)`? 
Possibly without reloading the collection view as well?

lzyraining:
@SD10 The collection view reload is not necessary. Thanks for pointing to that. I have removed it. The key point is to invalidate the collectionLayout before subviews is layout.

Using `viewWillTransition(to:coordinator:)` seems to cause two troubles to me. I need to call `coordinator.animate(alongsideTransition: , completion: )`. It looks like during the transition it can not detect that bounds has been changed. I need to force call `messageLayout.invalidateLayout()`(Maybe I did something wrong). Another thing is that I can not control the animation.

SD10:
@lzyraining I think I've solved this in the `MessagesCollectionViewLayout` itself by adding an invalidation context. I'd really rather leave all invalidation logic to the flow layout itself. I'll push some changes to `v0.2.0` later tonight and it would be great if you could see if this fixes the problem.

SD10:
@lzyraining Does this layout bug still occur for you on the latest `v0.2.0`?

lzyraining:
@SD10 Layout looks good now. But avatar is wrong placed sometimes when scroll the table.
screen shot 2017-08-01 at 3 20 56 pm

SD10:
@lzyraining This is hard for me to reproduce. I think it's more likely to be a separate issue.

PR #21, opened 2017-08-01
MessageInputBar improvement

andreaantonioni:
* Implement `InputBarView` as textView adding the support for a custom placeholder
* Disable `sendButton` when there's no text to send
* Change `sendButton` font to improve UI
* Implement auto-clear input bar when a message is sent

### DIFF
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -63,26 +63,26 @@ class ConversationViewController: MessagesViewController, MessagesDataSource, Me
                    "Pellentesque venenatis, ante et hendrerit rutrum" +
                    "Quam erat vehicula metus, et condimentum ante tellus augue."
 
-        messages.append(MockMessage(text: msg2, sender: sender2, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg4, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg5, sender: sender3, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg1, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg3, sender: sender1, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg3, sender: sender1, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg2, sender: sender2, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg2, sender: sender2, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg1, sender: currentSender(), id: NSUUID().uuidString))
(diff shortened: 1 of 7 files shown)


Example/Sources/ConversationViewController.swift
```diff
@@ -63,26 +63,26 @@ class ConversationViewController: MessagesViewController, MessagesDataSource, Me
...
-        messages.append(MockMessage(text: msg4, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg5, sender: sender3, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg4, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg5, sender: sender3, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg4, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg5, sender: sender3, id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg1, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg1, sender: currentSender(), id: NSUUID().uuidString))
-        messages.append(MockMessage(text: msg3, sender: sender1, id: NSUUID().uuidString))
+        messages.append(MockMessage(text: msg2, sender: sender2, messageId: NSUUID().uuidString))
```
SD10: You might have to rebase this / fix the parameter name. The latest `v0.2.0` uses `id`. 
I briefly changed it to `messageId` before disabling the SwiftLint error.

Sources/InputTextView.swift
```diff
@@ -0,0 +1,100 @@
...
+            }
+
+            placeholder.draw(in: rect, withAttributes: attributes)
+        }
+    }
+
+    fileprivate func commonInit() {
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(textDidChange),
+                                               name: NSNotification.Name.UITextViewTextDidChange,
```
SD10, line 88: Nit, but can we just use `.UITextViewTextDidChange`?

Sources/InputTextView.swift
```diff
@@ -0,0 +1,100 @@
...
+                                               object: self)
+    }
+
+    func textDidChange(notification: Notification) {
```
SD10: Any specific reason why we moved from using the delegate to the `NotificationCenter`?

SD10, line 92: Nvm, I see now. Quick fingers 😛

SD10:
@andreaantonioni There's definitely some good in this PR, but `v0.2.0` is already far behind regarding fixing some bugs for the `InputTextView` and layout orientation changes.

You may have to re-submit this PR to an updated `v0.2.0` later tonight.

For all watching, if you're going to submit a PR it would be best to give me a heads up beforehand so that we can coordinate what you want to work on 😅

PR #24, opened 2017-08-03
MessageInputBar TextView Improvement

iJasonWade:
Added delegate methods so the placeholder text disappears when the user begins to type. The text is added back when the text field resigns active and the text is empty, also added a check in the send button action to prevent sending empty messages.

### DIFF
--- a/Sources/MessagesViewController.swift
+++ b/Sources/MessagesViewController.swift
@@ -22,190 +22,192 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-
 import UIKit
 
-
 open class MessagesViewController: UIViewController {
-	
+
 	// MARK: - Properties
-	
+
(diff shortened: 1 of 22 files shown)


Sources/MessagesViewController.swift
```diff
@@ -175,6 +176,25 @@ extension MessagesViewController: UICollectionViewDataSource {
 
 }
 
+// MARK: - UITextFieldDelegate Methods
+
+extension MessagesViewController: UITextViewDelegate {
```
SD10, line 181: @iJasonWade Thank you for the PR but this is pretty brittle considering the current app uses "New Message" as text and not placeholder text etc. There are a few minor UI details like this that need to be cleaned up.

I also already have the delegate for `textInputView` set inside of `MessageInputBar`. I'm trying to keep things out of the `MessagesViewController` when possible 🤔

I would also prefer discussing PRs first. We recently made our Slack channel public for this reason. 
It looks like I should open an issue for this.

iJasonWade: I understand. I did join the slack room so I will jump into the conversation there.

PR #26, opened 2017-08-04
Avatar

MacMeDan:
Avatar class implementation. Added the ability to show initials on an avatar. Defaults to values for everything. Added Tests and a `Playground` so others can play with it.


SD10:
@MacMeDan Thanks! I still have to review this but I only wish you could've submitted this to `v0.2.0` instead of `master` 😭 Any way we can do this? It may be better for me to review it first.

Sources/Avatar.swift
```diff
@@ -24,28 +24,93 @@
...
+    
+    open func setBackground(color: UIColor) {
+        backgroundColor = color
+    }
+    
+    open func getImage() -> UIImage {
+        return imageView.image ?? UIImage()
+    }
+    
+    open func roundCorners(by: CGFloat?) {
```
SD10, line 108: I think adding an internal parameter `radius` would be great 😃

Sources/Avatar.swift
```diff
@@ -24,28 +24,93 @@
...
+        initalsLabel.font = UIFont.systemFont(ofSize: size)
+        initalsLabel.textColor = color
+    }
+    
+    open func setBackground(color: UIColor) {
+        backgroundColor = color
+    }
+    
+    open func getImage() -> UIImage {
+        return imageView.image ?? UIImage()
```
SD10, line 105: Any specific reason why this doesn't return an optional?

Tests/AvatarTests.swift
```diff
@@ -0,0 +1,80 @@
...
+    
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    func testNoParams() {
+        let avatar = Avatar()
+        XCTAssertEqual(avatar.initals, "?")
+        XCTAssertEqual(avatar.initalsLabel.text, "?")
+        XCTAssertEqual(avatar.layer.cornerRadius, 0.0)
```
SD10, line 26: Looks like this is failing the tests on iOS 9+. 
Avatar has a default size of `30 x 30` resulting in a `cornerRadius` of `15.0`

.swiftlint.yml
```diff
@@ -1,6 +1,6 @@
-line_length:
-  warning: 150
-  ignores_comments: true
 
 disabled_rules:
-  identifier_name
+- identifier_name
+- trailing_whitespace
+- line_length
```
SD10, line 5: I do like the `trailing_whitespace` rule 🤔 
Same with `line_length` but this is so trivial, let's leave it for now.

SD10 (changes requested):
Great work, thank you for writing unit tests 😅
If you could please review the comments and fix the failing unit tests ❤️

Sources/Avatar.swift
```diff
@@ -24,28 +24,93 @@
...
-
-    public let highlightedImage: UIImage?
-
-    public let placeholderImage: UIImage
-
-    public init(image: UIImage? = nil, highlightedImage: UIImage? = nil, placeholderImage: UIImage) {
-        self.image = image
-        self.highlightedImage = highlightedImage
-        self.placeholderImage = placeholderImage
+open class Avatar: UIView {
```
SD10, line 27: Do you think `AvatarView` would be a better name? Allowing people to identify its a `UIView` subclass?

SD10:
I'm just going to push this off to `v0.3.0`. Sorry for the nightmare of opening an closing PR's.

Sources/Avatar.swift
MacMeDan: Ya that is what I had it called originally and then changed it for simplicity but I am not married to the idea I think I will go back to avatar view.

.swiftlint.yml
MacMeDan: So if you enable `trailing_whitespace` it throws a warning for blank line between methods. I like to separate each method with one empty line to make it easier to read. That is why i disabled it. 

screen shot 2017-08-05 at 8 12 26 pm

SD10: You can still have empty lines but just no spaces on the empty line 🤔

fbartho: For what it's worth, there's an option that helps with this in Xcode Preferences->Text Editing->Editing -- make sure the bottom of the two checkboxes in the following screenshot is checked: https://cl.ly/0d3C2E2B3x0A

SD10:
Closing in favor of #31

PR #31, opened 2017-08-06
AvatarView

MacMeDan:
Created an AvatarView which allows for initials to be set for an avatar that does not have a picture. Also added a `Playground` for people to play with it and understand it capabilities. Tests added because that is what you are supposed to do when you add a feature. Code review suggestions also integrated into this pull request.


.swiftlint.yml
```diff
@@ -1,6 +1,6 @@
-line_length:
-  warning: 150
-  ignores_comments: true
 
 disabled_rules:
-  identifier_name
+- identifier_name
+- trailing_whitespace
+- line_length
+- variable_name
```
SD10: I still think your SwiftLint might be outdated. 
`variable_name` has been deprecated. This SwiftLint doesn't compile for me 🤔
Other than this, LGTM

SD10 (approved):

PR #32, opened 2017-08-06
InputTextView

andreaantonioni:
Implement a subclass of `UITextView` called `InputTextView` which supports a placeholder


SD10:
screen shot 2017-08-06 at 9 19 56 pm

@andreaantonioni Good work, but there seems to be a bug regarding the placeholder text not disappearing when typing 🤔 It could be a stateful bug regarding the flag. Small race condition? You'll have to investigate for me.

MacMeDan:
and Tests. 😅

SD10:
It's hard for me to require tests considering I haven't added any myself. 
We also badly need this feature. @MacMeDan You can play the bad cop here 😁
@andreaantonioni If you're able to include tests that would be great.

SD10:
@andreaantonioni If you're able to do any testing we'll do it in a separate PR. Thank you for this! 💯

MacMeDan:
@SD10 I was not trying to require tests of anyone 😅 . Merely suggesting them and agree does not have to be done in this pr or even by the same contributor. Any one could add them 😁 in fact it would be a great starter task for anyone that would like to contribute. 😜

andreaantonioni:
Ahahaha @MacMeDan is right, new features should always be delivered with tests!
I'll write them asap 👍🏻

MacMeDan:
Heck I am even cool with just one. (Secretly just want @SD10 to be the only one who has not written any) 😈

PR #47, opened 2017-08-10
Focusing on the "why" 

kyledotmp3:
Project managers are always telling us to "focus on the why", so that's what my initial change to the README did. 

Developer UX is forgotten far too often, and I'd like to remind the MessageKit community. Rather than focusing on creating a super-robust and customizable framework, let's create a super easy-to-use Message-UI that is **EVEN EASIER** than JSQ was! 

Swifty Methods! Code so easy to use that documentation is hardly even necessary! Let's do it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Why?
 
-UX is important, but people often forget about Developer UX. MessageKit is a framework geared towards making a chat interface easy for both users and developers.
+UX is important-- but people often forget about Developer UX. MessageKit is a framework geared towards making a chat interface easy for both users and developers.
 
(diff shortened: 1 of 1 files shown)


SD10:
@kbendelow Thank you for your PR! 🎉 Just touching on a few things -- 
I don't think the `What is it?` section is a good fit the readme. It's kind of repetitive and is almost better suited as the why. As for the `Why?` section and explaining "Developer UX", I think this could be better talked about elsewhere. It may be easier for people to contribute if I provide some structure for the document first.

I'm going to close this for now but thank you for your work.

Issue #33, opened 2017-08-06
Collection View Content Size Issue


gungorbasa:
Whenever I try to send a new message, collection view adds huge empty space along side with message as well. Here is the pictures of behavior.

<img width="311" alt="screen shot 2017-08-05 at 11 53 17 pm" src="https://user-images.githubusercontent.com/2487031/29004897-512e4094-7a84-11e7-851a-635688a69eb5.png">
<img width="311" alt="screen shot 2017-08-05 at 11 53 37 pm" src="https://user-images.githubusercontent.com/2487031/29004896-512e1376-7a84-11e7-8c63-7f5ce8ca0f00.png">

SD10:
@gungorbasa Thank you for reporting this. It looks like this is caused by using a hardware keyboard. The `contentOffset` for the `collectionView` is set back to normal by `handleKeyboardWillHide`, which doesn't get triggered in this scenario.

PR #53, opened 2017-08-10
Bug: Collection View Content Size Issue #33 is fixed.

gungorbasa:
Collection view edge inset problem is fixed. (While user uses hardware keyboard)


SD10:
Duplicate of #54

Issue #46, opened 2017-08-10
Improve README and add MessageKit banner


SD10:
### Improve README and add MessageKit banner

The README could use some work 😅. 
I'll definitely work on updating it but this is also a good starter task.

We also could use a banner for the README. 
Something like the [JSQMessagesViewController banner](https://github.com/jessesquires/JSQMessagesViewController) would be great.

I'm open to other designs, I'm just not a good artist so I have no suggestions. 
Overall, I'd like to continue using the current **MessageKit** logo.

mding5692:
Sorry mind posting the logo just in case somebody doesn't remember

mding5692:
Just made some banners for fun right now, any comments on color scheme and other critiques? 

Basically copied over the JSQMessageViewController banner. 

I chose Orange as background colour (happy to change) because its a warmer and much more welcoming and happy colour and feels comfortable. 

![messagekitbanner1](https://user-images.githubusercontent.com/9067177/29157387-c986f264-7d74-11e7-8714-28e01951492c.png)
![messagekitbanner2](https://user-images.githubusercontent.com/9067177/29157388-c98720cc-7d74-11e7-8c20-a8ff5500dac7.png)
![messagekitbanner3](https://user-images.githubusercontent.com/9067177/29157389-c988d96c-7d74-11e7-91e3-47332a87cf4d.png)
![messagekitbanner4](https://user-images.githubusercontent.com/9067177/29157390-c98afe90-7d74-11e7-8e80-071ebc3b915b.png)

Noticed the subtitle font size is a bit small

SD10:
@mding5692 Thanks for doing this! I definitely think we should continue to use a green color so it matches with our organization logo. You can increase the subtitle font size if you want or even remove the subtitle altogether. I can't critique more than that, better than anything I could've done. Maybe someone else has some input?

mding5692:
@SD10 haha no problem, I do this Sketch-related stuff for fun anyways.
Heres green colour: 

![messagekitbanner2](https://user-images.githubusercontent.com/9067177/29158666-e2474c5e-7d79-11e7-987e-d5fc1887e22d.png)

Note that I dont actually know what colour green was used, so this is just roughly the same colour as JSQMessageViewController banner background colour

omaralbeik:
I'm not a designer, but here is what I got 😅 
![1](https://user-images.githubusercontent.com/8127757/29166152-f2ad1160-7dcd-11e7-9743-c45e61167daa.jpg)
[...]

Issue #22, opened 2017-08-01
Slack Invite. 


MacMeDan:
Hey Everyone,
 We have a slack channel secured for helping coordinating features and dev in real time. Most conversations/discussions will happen here on github and we want to keep it that way. But at times tho it can be difficult to get things merged in when there are conflicts so we have created this slack group. 

Click the link to join. Be respectful and encouraging. 

https://join.slack.com/t/messagekit/shared_invite/enQtNTU5NTQ0NTQ5OTg3LTY5NDdmZDk1NzU1YjBjZjY1MTExNzkwMjcxZTkyYzhjNzdhMTM2YzFiNzE4YmRlYjI2MjdkMWRhNGEwMjg2NDE

This link will expire a week from 9/7/2017 please message me if it past that time and you would like to join.

xjli1972:
@MacMeDan : The Slack invitation link is no longer valid, could you please generate a new one?

SD10:
@xjli1972 Updated 👍

Michele404:
Update

nathantannar4:
@Michele404 is it not working? Im not sure how to interpret your comment

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Michele404:
Yes, its not working

nathantannar4:
https://join.slack.com/t/messagekit/shared_invite/enQtNTU5NTQ0NTQ5OTg3LTY5NDdmZDk1NzU1YjBjZjY1MTExNzkwMjcxZTkyYzhjNzdhMTM2YzFiNzE4YmRlYjI2MjdkMWRhNGEwMjg2NDE

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

MacMeDan:
I updated the link in the original message.

PR #58, opened 2017-08-11
Update README.

MacMeDan:
Check it out. @omaralbeik If we could get the banner added to this that would be awesome. Let me know if anything is missing or we would like to add. Please also read through it to make sure it makes sense.  #46 

Has instructions for getting added to slack channel which makes #22 closeable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,8 @@ Welcome to MessagesKit.
 * [**What's Next**](#What's-Next)
 * [**Apps using this library**](#Apps-using-this-library)
 
-[**License**](#font-installation)
+[**License**](#font-License)
 
-[**Installation Options**](#font-installation)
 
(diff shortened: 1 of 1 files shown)


SD10:
@MacMeDan Thank you so much 👍 This takes a lot off my plate. Now we will have something presentable. I will just make a few changes before merging.

@omaralbeik Maybe you can add the banner when you get some time?

SD10:
I'll take care of any changes from the `v0.4.0` branch, thank's for getting us this far!

MacMeDan:
Ya, that would be awesome. That is kinda what I was thinking by pull requesting it. I had done everything that I had access to and was open to anyone fixing what they wanted and pull requesting that. So I appreciate it.

PR #60, opened 2017-08-12
[WIP] [Example] Feature/attributed text bubble

SD10:
So I'm having trouble getting `NSAttributedString`'s method `boundingRect(size:options:context)` to return a proper size for the `MessageContainerView` when a `MessageLabel` is using attributed text.

The method `sizeThatFits` of `UIView` is returning the proper size. However, the `MessagesCollectionViewFlowLayout` needs this size to determine the height of the cell. Therefore, to use this in the layout object we would have to initialize a temporary `UILabel`, set the attributed text on it, and then call `sizeThatFits` to retrieve the size fitting that text. 

This works fine, but I don't like the overhead of having to initialize a UILabel just to get the size and then throw it away 🤔

I've given `MessagesCollectionViewFlowLayout` a type property - `placeHolderMessagesLabel` in hopes of not having to constantly initialize one in the body of the sizing methods. I'm not sure what the implications of this are 😞

Any feedback would be much appreciated - lurkers? @MacMeDan, @fbartho?

EDIT: The label should probably just be a private instance property and it would just float around for the lifetime of the `MessagesCollectionViewFlowLayout` 😂

### DIFF
--- a/Example/Sources/MockMessage.swift
+++ b/Example/Sources/MockMessage.swift
@@ -39,4 +39,11 @@ struct MockMessage: MessageType {
         self.sentDate = Date()
     }
 
+    init(text: NSAttributedString, sender: Sender, messageId: String) {
+        data = .attributedText(text)
+        self.sender = sender
+        self.messageId = messageId
+        self.sentDate = Date()
+    }
+
 }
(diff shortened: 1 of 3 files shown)


fbartho:
@SD10 -- this is an absolutely common/classic problem with CollectionView/TableView Cell Layouts. -- It's very heavyweight but ends up being what people with complex layouts need to do to measure classic layouts ahead of time.

Instead, you could use a heuristic to estimate message cell size based on contents-length, and then let the cell do auto layout as it comes on-screen. -- This has some performance considerations, but we should measure them with this strategy, as auto layout is probably more friendly than fixed sizing.

Issue #64, opened 2017-08-14
Change color of text. 


MacMeDan:
There is no method to change the color of the text for an `incoming Message` vs `outgoing message`

SD10:
@MacMeDan I was hoping they would use:
```Swift
func messageColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor
```
Here you get the message passed in so you can check if the message is from the current sender and return a different UIColor accordingly.

I thought about splitting this into two methods for `incoming` and `outgoing`, but I felt like this method was more flexible because you could have an unlimited number of different colors 🤔

fbartho:
@SD10 I like that API -- because my intended use-case is going to have more types of messages in the stream, and incoming/outgoing is only one of the Axis. I also would tweak the coloring for system messages, location updates, and other events in the message stream.

MacMeDan:
Yes I agree @fbartho and I am sorry guys, I definitely worded this in a way that it seemed like I suggesting just `Incoming` and `Outgoing` that was meant to just be an example use case. I am trying to leave it open ended just a method for providing an override point to accomplish this.

SD10:
This is closed by #66

PR #65, opened 2017-08-14
Message Text Color Customization. 

MacMeDan:
Added methods to the `MessagesDisplayDataSource` protocol to allow for text color customization and removed unnecessary parameters.


Sources/MessagesDisplayDataSource.swift
```diff
@@ -26,7 +26,9 @@ import Foundation
 
 public protocol MessagesDisplayDataSource: class, MessagesDataSource {
 
-    func messageColorFor(_ message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor
+    func textColorFor(_ message: MessageType) -> UIColor
```
SD10, line 29: Adding a method for the `textColor` is not a bad idea. I was originally thinking they could override the color of the `messageLabel` in the cell to make this change. This is definitely more friendly. My only concern was adding to the surface area of the API. You could make dozens of these customization methods.

Sources/MessagesDisplayDataSource.swift
```diff
@@ -26,7 +26,9 @@ import Foundation
...
+    
+    func backgroundColorFor(_ message: MessageType) -> UIColor
```
SD10: As for the excluded parameters, those are actually pretty important IMO. We should try and pass in as much information as possible if we have it. You can use `indexPath` to set colors based on even or odd numbered sections for example. You can use `messagesCollectionView` to get the total number of sections.

Sources/MessagesDisplayDataSource.swift
```diff
@@ -41,8 +43,12 @@ public protocol MessagesDisplayDataSource: class, MessagesDataSource {
 }
 
 public extension MessagesDisplayDataSource {
-
-    func messageColorFor(_ message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor {
```
SD10: Unfortunately, I didn't rebase the `v0.5.0` branch so this API is out of date.

SD10 (commented):
Good work but see my comments in #64. I believe this introduces some limitations.

Sources/MessagesDisplayDataSource.swift
MacMeDan: I see this as the only customisation available. I would not add a `textColor` method for the top / bottom / header / footer labels and since they are `attributedText` it doesn't really make sense either.

Sources/MessagesDisplayDataSource.swift
MacMeDan: but what is the reasoning for `in messagesCollectionView: MessagesCollectionView`

MacMeDan: I feel like there could be a method for this information that way you have the option to get it or not. I was just finding the information redundant and that's why I chose that route.  But if you can see a good case for it I could add it back in.

MacMeDan:
Closed #64

SD10:
@MacMeDan The API you were making changes to is outdated. 
`v0.5.0` now represents the current API but there are conflicts.

We should definitely be passing `IndexPath` and `MessagesCollectionView` into all the delegate methods. If we add a delegate method to change the text, it should match the current API:

```Swift
func textColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor
```

MacMeDan:
@SD10 I think `messageColor` is confusing just as a name itself in determining what the method is doing.

SD10:
@MacMeDan I totally agree. We should rename it to:

```Swift
func backgroundColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) ->UIColor
```

I'm confused why there are all these commits from `v0.4.0` in this pull request to `v0.5.0`. The `v0.5.0` branch already has these commits...

MacMeDan:
Yes I am also confused. Something went wrong with the rebase. or you rebased the branch and then I did.

fbartho:
This is why I recommend against rebasing (`git pull --rebase`being the only exception)

SD10:
We'll probably just have to open a new PR with the changes. I don't want to corrupt the git history

MacMeDan:
Im just going to close this and make a new request. lets not make branches preemptively.

PR #66, opened 2017-08-14
Message text color2

### LINKED ISSUE #64: Change color of text.

MacMeDan:
Closes #64 
Adds Tests for the `MessagesDisplayDataSource` protocol
Makes `MessagesDisplayDataSource` have a default implementation.
Adds a `textColor` method for setting the color of the text for an individual message.

### DIFF
--- a/Sources/MessagesCollectionView.swift
+++ b/Sources/MessagesCollectionView.swift
@@ -33,8 +33,6 @@ open class MessagesCollectionView: UICollectionView {
     open weak var messagesLayoutDelegate: MessagesLayoutDelegate?
 
     open weak var messageCellDelegate: MessageCellDelegate?
-    
-    open weak var messagesDisplayDataSource: MessagesDisplayDataSource?
 
     private var indexPathForLastItem: IndexPath? {
 
--- a/Sources/MessagesCollectionView.swift
+++ b/Sources/MessagesCollectionView.swift
@@ -33,6 +33,8 @@ open class MessagesCollectionView: UICollectionView {
     open weak var messagesLayoutDelegate: MessagesLayoutDelegate?
 
     open weak var messageCellDelegate: MessageCellDelegate?
+    
+    open weak var messagesDisplayDataSource: MessagesDisplayDataSource?
 
     private var indexPathForLastItem: IndexPath? {
 
--- a/Sources/MessagesDisplayDataSource.swift
+++ b/Sources/MessagesDisplayDataSource.swift
@@ -25,41 +25,52 @@
 import Foundation
 
 public protocol MessagesDisplayDataSource: class, MessagesDataSource {
-
-    func messageColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor
-
-    func avatar(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> Avatar
-
+    
+    func textColor(for message: MessageType, at indexPath: IndexPath) -> UIColor
+    
+    func backgroundColor(for message: MessageType, at  indexPath: IndexPath) -> UIColor
+    
+    func avatar(for message: MessageType) -> Avatar
+    
     func messageHeaderView(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> MessageHeaderView?
-
+    
     func messageFooterView(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> MessageFooterView?
-
+    
     func cellTopLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
-
+    
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
-
+    
 }
 
-public extension MessagesDisplayDataSource {
-
-    func messageColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor {
+// "where Self: MessagesViewController" make these methods a `default` implimentation for any view that subclasses of MessagesViewController
+public extension MessagesDisplayDataSource where Self: MessagesViewController {
+    
+    func textColor(for message: MessageType, at indexPath: IndexPath) -> UIColor {
+        return isFromCurrentSender(message: message) ? .white : .black
+    }
+    
+    func backgroundColor(for message: MessageType, at  indexPath: IndexPath) -> UIColor {
         return isFromCurrentSender(message: message) ? .outgoingGreen : .incomingGray
     }
-
+    
+    func avatar(for message: MessageType) -> Avatar {
+        return Avatar()
+    }
+    
     func messageHeaderView(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> MessageHeaderView? {
         return nil
     }
-
+    
     func messageFooterView(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> MessageFooterView? {
         return nil
     }
-
+    
     func cellTopLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
         return nil
     }
-
+    
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
         return nil
     }
-
+    
 }
--- a/Sources/MessagesDisplayDataSource.swift
+++ b/Sources/MessagesDisplayDataSource.swift
@@ -26,11 +26,11 @@ import Foundation
 
 public protocol MessagesDisplayDataSource: class, MessagesDataSource {
     
-    func textColor(for message: MessageType, at indexPath: IndexPath) -> UIColor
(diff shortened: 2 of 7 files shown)


Sources/MessagesDisplayDataSource.swift
```diff
@@ -25,41 +25,52 @@
 import Foundation
 
 public protocol MessagesDisplayDataSource: class, MessagesDataSource {
-
-    func messageColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor
-
-    func avatar(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> Avatar
-
+    
+    func textColor(for message: MessageType, at indexPath: IndexPath) -> UIColor
```
SD10: We need the `messagesCollectionView` still 😅

Sources/MessagesDisplayDataSource.swift
```diff
@@ -25,41 +25,52 @@
...
+    
+    func backgroundColor(for message: MessageType, at  indexPath: IndexPath) -> UIColor
```
SD10: We need the `messagesCollectionView` still 😅

Sources/MessagesDisplayDataSource.swift
```diff
@@ -25,41 +25,52 @@
...
+    
+    func avatar(for message: MessageType) -> Avatar
```
SD10: The old API was what I was looking for 😅

MacMeDan: I don't understand this comment.

Sources/MessagesCollectionView.swift
```diff
@@ -33,6 +33,8 @@ open class MessagesCollectionView: UICollectionView {
     open weak var messagesLayoutDelegate: MessagesLayoutDelegate?
 
     open weak var messageCellDelegate: MessageCellDelegate?
+    
+    open weak var messagesDisplayDataSource: MessagesDisplayDataSource?
```
SD10: We don't need this property because `MessagesDisplayDataSource` conforms to `MessagesDataSource`

Sources/MessagesDisplayDataSource.swift
SD10: Sorry, I meant we should continue to use:
```Swift
func avatar(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> Avatar
```
The goal here is to be able to access any MessageKit component from these methods, just like how UIKit does with UICollectionView & UITableView.

We can access the specific row, the specific section, the `MessagesCollectionView` itself, thus we can also access the `MessagesCollectionViewFlowLayout` object, you can even access a specific `MessagesCollectionViewCell`. I'm not looking for brevity but rather flexibility and customization. @fbartho already mentioned he would need access to more components than just the `MessageType` to make his customizations

MacMeDan: I agree in general but I think in respect to the avatar I think it's unnecessary and will not be used thus is just bloat.

Sources/MessagesCollectionView.swift
MacMeDan: fixed.

SD10:
@MacMeDan Let's try to use descriptive commit messages even if we disagree 😐

The reason why we're passing them in here is because these methods are used by:
```Swift
func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell
```
You should have everything available to you that UIKit gives you for when these messages are configured.

I just don't think we should sacrifice customization for brevity. Not a single delegate method in `UICollectionView` or `UITableView` is short, and that's by design.

SD10:
@MacMeDan Let's say a user wanted to group recent messages by a user together into a single bubble. Only returning an avatar for the top message in the group. To do this you would need the current `IndexPath` and the `MessagesCollectionView` so that you can evaluate the surrounding messages and determine whether to return an Avatar. It's likely the return type changes to `Avatar?` in the future to support this -- but I haven't worked out the specifics.

I'm just trying to provide you with a use case so you can understand I'm not trying to hyper critical of you or acting on a whim. If I'm ever putting my foot down on something I'll always try to provide sound reasoning.

Sources/MessagesDisplayDataSource.swift
```diff
@@ -25,41 +25,52 @@
...
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
-
+    
 }
 
-public extension MessagesDisplayDataSource {
-
-    func messageColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor {
+// "where Self: MessagesViewController" make these methods a `default` implimentation for any view that subclasses of MessagesViewController
+public extension MessagesDisplayDataSource where Self: MessagesViewController {
```
SD10: I think this was better without the `Self: MessagesViewController` requirement, because now you only get default behavior for `MessagesViewController`. 

If the `MessagesDisplayDataSource` is another object, you get no defaults. The original implementation gives you defaults for any object because the `MessagesDisplayDataSource` is not dependent on `MessagesCollectionView` -- only the `MessagesDataSource`. 

It's very possible to use this library without the `MessagesViewController` and just the `MessagesCollectionView` in your own controller. It's also very possible to use a view model helper object as the `MessagesDisplayDataSource`. In fact, because they're decoupled from the original `delegate`/`dataSource` methods -- they're not even limited by `NSObjectProtocol` 😃

Tests/TestMessagesViewControllerModel.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  TestMessagesViewControllerModel.swift
+//  MessageKit
+//
+//  Created by Dan Leonard on 8/14/17.
+//  Copyright © 2017 MessageKit. All rights reserved.
+//
+
+import Foundation
+import Foundation
```
SD10: Double import here

Tests/TestMessagesViewControllerModel.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  TestMessagesViewControllerModel.swift
+//  MessageKit
+//
+//  Created by Dan Leonard on 8/14/17.
+//  Copyright © 2017 MessageKit. All rights reserved.
+//
```
SD10: It would be nice if Xcode did something to autogenerate the MIT license at the top of the file 🤔

Sherlouk: This is possible as of Xcode 9, nice article explaining it [here](https://oleb.net/blog/2017/07/xcode-9-text-macros/)!

SD10: @Sherlouk Thanks! Much appreciated 👍

Sources/MessagesDisplayDataSource.swift
MacMeDan: Ahh I did not even give that a thought. Thank you for pointing it out. Also this change was based on the branch before it was rebased on master. I did it to make testing work I could definitely see it being unnecessary. I will change it back and make sure it all works. 💯 Thanks for catching it.

Tests/TestMessagesViewControllerModel.swift
MacMeDan: My Bad.

Tests/TestMessagesViewControllerModel.swift
MacMeDan: Do you want me to manually just put the license or what would you like done in this case?

MacMeDan:
@SD10 
> @MacMeDan Let's say a user wanted to group recent messages by a user together into a single bubble. Only returning an avatar for the top message in the group. To do this you would need the current IndexPath and the MessagesCollectionView so that you can evaluate the surrounding messages and determine whether to return an Avatar. It's likely the return type changes to Avatar? in the future to support this -- but I haven't worked out the specifics.

Thank you for the reasoning this means allot. I have worked with a few people that could not provide this and appreciate the time you took to do so. I will do better at my descriptions.

Tests/TestMessagesViewControllerModel.swift
lzyraining: Check this post https://robots.thoughtbot.com/creating-custom-xcode-templates, it's possible to create a custom template add MIT license.

SD10: @MacMeDan I think we can just paste it manually for now. I'm still using Xcode 8 but we will automate this in Xcode 9 😃 @lzyraining Thank you for that 👍

legalimpurity: Hi Guys, I would like to contribute as well. Your slack invitation is also expired. Can anyone please send me a link at legalimpurity@gmail.com

MacMeDan: Hey @legalimpurity I update the readme link to not expire. https://join.slack.com/t/messagekit/shared_invite/MjI4NzIzNzMyMzU0LTE1MDMwODIzMDUtYzllYzIyNTU4MA

MacMeDan: Looks like you are already in 🤘

MacMeDan:
Alright I think all of the review suggestions have been addressed and this is ready again.

SD10 (approved):
Thanks @MacMeDan 🎉

Issue #61, opened 2017-08-13
Crash issue when no message


superdev89:
The sample app crashes when you sent empty message list.

I debugged into the code and found the issue in MessagesCollectionView.swift.

    var indexPathForLastItem: IndexPath? {

          let lastSection = numberOfSections > 0 ? numberOfSections - 1 : 0
          guard numberOfItems(inSection: lastSection) > 0 else { return nil }
          return IndexPath(item: numberOfItems(inSection: lastSection) - 1, section: lastSection)
    }

if numberOfSections is 0, the lastSection should not exist and return nil, but here retrieving section 0.

SD10:
@superdev89 Thank you for reporting this! 🎉 
You can submit a PR to the `v0.5.0` branch if you'd like to fix this yourself.

MacMeDan:
@superdev89 Let us know if you would like to fix it or we will open it up to anyone. Great job finding it. 💪😎👊 I am going to assign it to you for now.

AndrewLauder:
I fixed this issue locally, should I open a PR?

MacMeDan:
That would be great @AndrewLauder.

PR #73, opened 2017-08-16
Fix for Crash issue when no message #61

AndrewLauder:


SD10:
@AndrewLauder Thank you for this 🎉

PR #75, opened 2017-08-18
Update the Slack Link

MacMeDan:
Update the Slack Link to a link that never expires. And add it to the table of contents.


SD10 (approved):
Thank you @MacMeDan 🎉

PR #79, opened 2017-08-25
Update README with Carthage installation instructions

cwalo:
Carthage simply requires the Xcode scheme to be shared, which was already the case. This PR updates the README with formal installation instructions.


SD10:
Thanks @cwalo 🎉 🍻

PR #84, opened 2017-08-25
Redone MessageInputBar and its related views

nathantannar4:
Created a more customizable InputAccessoryView

Also updated the Example to show usage


SD10:
@nathantannar4 Thanks for taking the initiative to work on this 🎉 🍻

Sources/InputBarItem.swift
```diff
@@ -0,0 +1,260 @@
...
+
+open class InputBarButtonItem: UIButton {
+    
+    public enum Spacing {
+        case fixed(CGFloat)
+        case flexible
+        case none
+    }
+    
+    public typealias InputBarButtonItemAction = ((InputBarButtonItem) -> Void)?
```
SD10: We shouldn't type this explicitly as an optional type. It makes the call site unclear as to whether or not the type is optional.

Sources/InputTextView.swift
```diff
@@ -25,95 +25,121 @@
...
+    
+    private var placeholderLabelConstraintSet: NSLayoutConstraintSet?
+    
+    open var placeholder: String? {
+        get {
+            return placeholderLabel.text
+        }
+        set {
+            placeholderLabel.text = newValue
+        }
```
SD10: Can we just use `didSet` here?

Sources/InputTextView.swift
```diff
@@ -25,95 +25,121 @@
...
+    }
+    
+    open var placeholderTextColor: UIColor {
+        get {
+            return placeholderLabel.textColor
+        }
+        set {
+            placeholderLabel.textColor = newValue
+        }
```
SD10: Can we just use `didSet` here?

Sources/InputTextView.swift
```diff
@@ -25,95 +25,121 @@
...
+    
     // MARK: - Properties
-
-    open var placeholder: NSString? {
+    
+    open let placeholderLabel: UILabel = {
+        let label = UILabel()
+        label.numberOfLines = 0
+        label.textColor = .lightGray
+        label.text = "Aa"
```
SD10: Let's have the default text for the placeholder be "New Message"

Sources/InputTextView.swift
```diff
@@ -25,95 +25,121 @@
...
+        label.backgroundColor = .clear
```
SD10: Can we set `label.isOpaque = false` if we're going to make it clear?

Sources/MessageInputBarDelegate.swift
```diff
@@ -24,8 +24,14 @@
...
-
+    
+    @objc optional func inputBar(_ inputBar: MessageInputBar, didSelectSendButtonWith text: String)
+    
+    @objc optional func inputBar(_ inputBar: MessageInputBar, didChangeIntrinsicContentTo size: CGSize)
+    
+    @objc optional func inputBar(_ inputBar: MessageInputBar, textViewTextDidChangeTo text: String)
+    
+    @objc optional func inputBar(_ inputBar: MessageInputBar, didSwipeTextViewWith gesture: UISwipeGestureRecognizer)
 }
```
SD10: Instead of marking these with the `@objc` attribute to make them optional. Can we just provide a default implementation via a protocol extension that is a no-op?

Sources/MessageInputBar.swift
```diff
@@ -24,133 +24,434 @@
...
+    }
+    
+    /// The InputBarItems held in the leftStackView
+    open var leftStackViewItems: [InputBarButtonItem] = []
+    
+    /// The InputBarItems held in the rightStackView
+    open var rightStackViewItems: [InputBarButtonItem] = []
+    
+    /// The InputBarItems held in the bottomStackView
+    open var bottomStackViewItems: [InputBarButtonItem] = []
```
SD10: Maybe we want the storage for these buttons to be `private(set)` so they're not able to be modified directly

Sources/MessageInputBar.swift
```diff
@@ -24,133 +24,434 @@
...
-        sendButton.isEnabled = false
-
-        return sendButton
+    
+    /// When set to true, the blurView in the background is shown and the backgroundColor is set to .clear. Default is FALSE
+    open var isTranslucent: Bool = false {
+        didSet {
+            blurView.isHidden = !isTranslucent
+            backgroundColor = isTranslucent ? .clear : .white
+        }
```
SD10, line 52: Same as before, can we toggle the `isOpaque` property when we switch between colors that are opaque or not?

Sources/MessageInputBar.swift
```diff
@@ -24,133 +24,434 @@
...
+        if previousIntrinsicContentSize != size {
+            delegate?.inputBar?(self, didChangeIntrinsicContentTo: size)
+        }
+        previousIntrinsicContentSize = size
+        return size
+    }
+    
+    /// The maximum intrinsicContentSize height. When reached the delegate 'didChangeIntrinsicContentTo' will be called.
+    open var maxHeight: CGFloat = 288 {
+        didSet {
```
SD10: Not too worried about this now but it would be nicer if the default value took into account the size of the device's screen. I'm not sure how this would affect the iPad.

Sources/InputBarItem.swift
nathantannar4: Yes you are write. The methods that are used to set the actions also don't need to have the optional type so I fixed that

Sources/MessageInputBarDelegate.swift
nathantannar4: Didn't know that was possible! Changed it to that

Sources/MessageInputBar.swift
nathantannar4: Could you explain what you mean by this? I thought isOpaque is only used when drawing content using `draw(_:)` method. https://developer.apple.com/documentation/uikit/uiview/1622622-isopaque

Sources/MessageInputBar.swift
nathantannar4: How about `open var maxHeight: CGFloat = UIScreen.main.bounds.height / {some constant}`

Sources/MessageInputBar.swift
SD10: Good to know 👍 Looks like we can leave these as is.

Sources/MessageInputBarDelegate.swift
```diff
@@ -24,8 +24,14 @@
...
 import Foundation
 
+@objc
 public protocol MessageInputBarDelegate: class {
-
-    func sendButtonPressed(sender: UIButton, textView: UITextView)
-
+    
+    @objc optional func inputBar(_ inputBar: MessageInputBar, didSelectSendButtonWith text: String)
+
```
SD10: Nit but do you mind renaming these to `messageInputBar`, I'd like to be more specific.
Do you think `didSelectSendButtonWith` -> `didPressSendButtonWith` is better?
Select makes me think some sort of state was set.

Also, the rest of the MessageKit delegates broke away from the Apple pattern of using the object as the method name. Instead, the method name represents exactly what it does.

IDK which approach is the best to be honest, but bike shedding aside, I think the API should feel similar across the library.

SD10 (dismissed):
@nathantannar4 Great work on this. Could you iterate on some of my feedback and also remove the builds for iOS 8 from our travis.yml file (since we are dropping iOS 8)?

nathantannar4:
Suggestions Made and Committed

SD10:
@nathantannar4 There's one last thing we need to check before merging regarding the bottom content inset of the collectionview. Now that this message input bar can be 2 levels high, I think we're prevented from scrolling to the very bottom and the message gets cut off. I can look tonight if you can't before me.

etoledom:
If there are problems with bottom content inset, check https://github.com/MessageKit/MessageKit/pull/87 , there might be a solution

nathantannar4:
Yea I can check but I agree with @etoledom I think this requires changes to the MessageViewController. There can't be a fixed inset because when the message input bar grows so does that inset

SD10:
@etoledom I've very briefly briefly looked at #87 but I'm skeptical about using a flag and doing this in `layoutSubviews`

nathantannar4:
I added the delegate method I had made earlier for the input bar for MessagesViewController.

`// MARK: - MessageInputBarDelegate

extension MessagesViewController: MessageInputBarDelegate {
    
    public func messageInputBar(_ inputBar: MessageInputBar, didChangeIntrinsicContentTo size: CGSize) {
        view.layoutIfNeeded()
        UIView.performWithoutAnimation {
            self.messagesCollectionViewBottomAnchor?.constant = -size.height - 38
            self.view.layoutIfNeeded()
        }
    }
}`

Corrects the bug. Maybe there is a more elegant way of doing this? But it can be a temp fix for now

etoledom:
@SD10 Sure, I don't like much either. I will simplify the pull request to just management of the insets. 
My mistake to try to do more stuff in there.
A fast question: Does it work the scroll to bottom on `viewWillAppear` for you?

SD10:
@nathantannar4 @etoledom Yeah that's why originally we were using `UIKeyboardFrameWillChange` notification to change the inset as the keyboard grows. It got reverted in an earlier PR. I think this will be a common place on conflict. I have a solution in mind.

Example/Sources/Playgrounds/Avatar.playground/Contents.swift
```diff
@@ -13,12 +13,12 @@ var avatarView = AvatarView()
 //: By default its a circlular avatar with a gray background and initals of "?"
 
 //: Create an avatar object and set it for the view.
-//var avatarObject = Avatar(image: testImage)
-//avatarView.set(avatar: avatarObject)
+var avatarObject = Avatar(image: testImage)
+avatarView.set(avatar: avatarObject)
```
SD10, line 17: Let's try to keep PR's to a specific scope so they're easier to review

Sources/InputBarItem.swift
```diff
@@ -0,0 +1,260 @@
...
+            invalidateIntrinsicContentSize()
+        }
+    }
+    
+    open override var intrinsicContentSize: CGSize {
+        var contentSize = size ?? super.intrinsicContentSize
+        switch spacing {
+        case .fixed(let width):
+            contentSize.width += width
+        default:
```
SD10, line 67: I don't like using `default` for switching on enum cases. If we add an extra case at a later date we won't get a compile error for it not being handled. No need to change just FYI

SD10:
@nathantannar4 I'm going to merge this into `v0.6.0` to make my changes. Thanks for all your hard work 🎉 🍺

SD10 (approved):

Sources/InputBarItem.swift
nathantannar4: For sure

nathantannar4:
@etoledom No scroll to bottom does not work for me

Issue #86, opened 2017-08-26
messagesCollectionView constraints and contentInsets


etoledom:
I have noticed a small issue with the `messagesCollectionView` top constraint, which is not with its superview’s top but on `topLayoutGuide`. I believe it was made this way to avoid problems between `autoInset` of the scroll view and the inset manually changed when showing / hidding the keyboard.

I think the messagesCollectionView’s top layout constraint should be with the top of its superview anyways to get advantage of the navigation bar’s transparency (as iMessage app does)

etoledom:
Already merged to 0.6.0 👌

PR #87, opened 2017-08-27
To solve issue #86 

etoledom:
**Changes**:
- messagesCollectionView now is constrained to top and bottom of superView
- Better management of inset to count for navigations bar, tab bars and the inputAccessoryView
- messagesCollectionView scrolls when showing keyboard to show the last message
- Correct bottom inset for any kind of keyboard (hardware keyboard not tested)

### DIFF
--- a/Sources/MessagesViewController.swift
+++ b/Sources/MessagesViewController.swift
@@ -254,16 +254,9 @@ extension MessagesViewController {
     func handleKeyboardWillShow(_ notification: Notification) {
         guard let keyboardFrame = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }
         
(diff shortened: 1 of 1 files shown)


etoledom:
Updated to avoid using `isFirstLayout` flag.

Observations:
- Following iMessage, I made it to scroll to bottom when someone open the keyboard. But thinking about it, it might not always be a desired behavior so I portably better remove it.
But it might be good (for now or later) to be able to implement this behavior easily from the library's user perspective

If you agree I'll remove that.

etoledom:
@SD10 so by the described in https://github.com/MessageKit/MessageKit/pull/84 ,  should I not touch the UIKeyboardFrameWillChange handler?, With the change on constraints, most of it became unnecessary, at least in the way it was implemented on 0.6.0 (actually it became simpler)

I'd just remove the automatic scrollToBottom when showing the keyboard

PR #90, opened 2017-08-29
messagesCollectionView top constraint to superView

etoledom:
Added top inset to topLayoutGuide.length
Bottom inset / constraints handling is untouched


SD10 (approved):
@etoledom thank you, great work 🎉

PR #92, opened 2017-08-31
Fix side effect of inputAccessoryView rendering bug fix

etoledom:


SD10:
inb4travisfails 😭

SD10:
We both agree that it's a little bit of a hack but we'll live with it for now. Thank you for taking care of this @etoledom 🎉 💯

PR #97, opened 2017-09-03
add updated logo

timglorioso:
Willing to make edits if desired!


timglorioso:
I edited the readme on my fork so you can preview it 👍

SD10:
@omaralbeik @nathantannar4 @MacMeDan What do you guys think? I like it but it could change again based on further README changes. 

Really appreciate you making these changes @timglorioso 💯 🎉

nathantannar4 (dismissed):
I like this!

README.md
```diff
@@ -1,5 +1,5 @@
 <p>
-  <img src="https://cdn.rawgit.com/MessageKit/MessageKit/master/Assets/mklogo.svg" title="MessageKit logo">
+  <img src="https://raw.githubusercontent.com/timglorioso/MessageKit/master/Assets/messagekit-logo.png" title="MessageKit logo">
```
SD10: @timglorioso Can we change this back and I'll merge the PR?

SD10 (changes requested):
See comments

README.md
timglorioso: @SD10 Assuming you want the URL updated to reflect the new filename?

SD10: You could also just rename the file to match the old one 😃

timglorioso:
I went ahead and kept using `raw.githubusercontent.com` because the RawGit CDN [permanently caches based on filename](https://github.com/rgrove/rawgit/blob/master/FAQ.md#how-long-does-the-cdn-cache-files-how-can-i-make-it-refresh-my-file), so we'd have to keep changing it (and the img src) for any future iterations 😉

SD10:
@timglorioso Any reason we are using a `.png` over the previous `.svg`?

timglorioso:
@SD10 The SVG format doesn't preserve text data as paths, so fonts don't render correctly if they can't be found client-side. I could force the text into a path if you'd like, shouldn't be hard 😁

SD10 (approved):
@timglorioso No need. Thanks for explaining and all your iterations 😃

PR #99, opened 2017-09-03
A workaround solution for the render issue with MessageInputBar

nathantannar4:
It seems as though when using a UINavigationController there is an issue with the rendering of an inputAccessoryView. This workaround creates a copy during the animation and then deletes it.

### DIFF
--- a/Sources/InputTextView.swift
+++ b/Sources/InputTextView.swift
@@ -37,6 +37,12 @@ open class InputTextView: UITextView {
         label.translatesAutoresizingMaskIntoConstraints = false
         return label
     }()
+    
+    open override var text: String! {
+        didSet {
+            placeholderLabel.isHidden = !text.isEmpty
+        }
+    }
 
     private var placeholderLabelConstraintSet: NSLayoutConstraintSet?
 
@@ -84,7 +90,8 @@ open class InputTextView: UITextView {
     }
 
     required public init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+        super.init(coder: aDecoder)
+        setup()
     }
 
     open func setup() {
--- a/Sources/MessagesViewController.swift
+++ b/Sources/MessagesViewController.swift
@@ -31,6 +31,7 @@ open class MessagesViewController: UIViewController {
 	open var messagesCollectionView = MessagesCollectionView(frame: .zero, collectionViewLayout: MessagesCollectionViewFlowLayout())
 
     open var messageInputBar = MessageInputBar()
+
     private var messageInputBarCopy: MessageInputBar?
     
 	override open var canBecomeFirstResponder: Bool {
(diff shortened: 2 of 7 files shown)


Sources/InputTextView.swift
```diff
@@ -84,7 +90,8 @@ open class InputTextView: UITextView {
     }
 
     required public init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+        super.init(coder: aDecoder)
+        setup()
```
SD10, line 94: @nathantannar4 ~Is this part of the workaround? There was a problem with unarchiving?~

Sources/MessagesViewController.swift
```diff
@@ -30,14 +30,15 @@ open class MessagesViewController: UIViewController {
 
 	open var messagesCollectionView = MessagesCollectionView(frame: .zero, collectionViewLayout: MessagesCollectionViewFlowLayout())
 
-	open var messageInputBar = MessageInputBar()
-
+    open var messageInputBar = MessageInputBar()
+    private var messageInputBarCopy: MessageInputBar?
```
SD10: Can we add a small comment above this explaning why we have it?

Sources/MessagesViewController.swift
```diff
@@ -127,6 +134,12 @@ open class MessagesViewController: UIViewController {
                                               toItem: bottomLayoutGuide, attribute: .top, multiplier: 1, constant: 0))
 	}
 
+    private func setupMessageInputBarCopy() {
+        messageInputBarCopy = messageInputBar.createCopy()
+        guard let copy = messageInputBarCopy else { return }
+        view.addSubview(copy)
+        copy.addConstraints(nil, left: view.leftAnchor, bottom: view.bottomAnchor, right: view.rightAnchor, topConstant: 0, leftConstant: 0, bottomConstant: 0, rightConstant: 0, widthConstant: 0, heightConstant: 0)
```
SD10, line 141: ~Mind splitting this up onto multiple lines?~

Sources/MessagesViewController.swift
```diff
@@ -75,6 +75,13 @@ open class MessagesViewController: UIViewController {
     open override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         addKeyboardObservers()
+        messageInputBarCopy?.removeFromSuperview()
+        messageInputBarCopy = nil
```
SD10: Can you put these two things in its own private method `removeMessageInputBarCopy()`?

SD10 (changes requested):
@nathantannar4 If you could just take care of a few nitty things. Mainly I want to add that comment for why we are doing it. One line should suffice

SD10:
@nathantannar4 ~I'm taking care of these now 👍~
Great work. Feel free to merge this in whenever 😃

SD10 (approved):

nathantannar4:
Thanks @SD10, sorry I didn't have a chance to get to this yesterday

PR #104, opened 2017-09-06
Possible fix for cut initials in avatar view.

iBolom:


MacMeDan (changes requested):
Lets go ahead and make some improvements here.

Sources/AvatarView.swift
```diff
@@ -48,7 +48,7 @@ open class AvatarView: UIView {
         let context = UIGraphicsGetCurrentContext()!
         
         //// Text Drawing
-        let textRect = CGRect(x: 5, y: 6, width: 20, height: 20)
+        let textRect = CGRect(x: 0, y: 0, width: 30, height: 30)
```
MacMeDan, line 51: I like this idea but let's go ahead an make some improvements here. Since we are really just using the views frame instead of hardcoding this size. So remove the `textRect` and replace any reference to it with the  `view.frame`.

MacMeDan: This way if a different size Avatar is used it will adjust accordingly. There is still no safety in preventing a large `font size` but at least it is an improvement.

MacMeDan (changes requested):
Take a look at the comments. there is more that we could do here but we can start with this I think it is an improvement.

SD10:
More discussion in #104

SD10:
@BojanAnox Thank you for looking into this 👍 It looks like we're moving forward with #111 in favor of this

PR #105, opened 2017-09-09
NSLayoutConstraint Usage Tweaks

fbartho:
- Move some constraint construction calls to their equivalent but simpler layoutAnchor forms
- Move from directly manipulating `isActive` to using `NSLayoutConstraint.activate()/deactivate()` -- as per [AppleDocs](https://developer.apple.com/documentation/uikit/nslayoutconstraint/1526955-activate)
>Typically, using this method is more efficient than activating each constraint individually.


nathantannar4 (approved):
Looks fine if Apple Docs says its faster

SD10 (approved):

Issue #110, opened 2017-09-10
[Prerelease] v0.7.0


SD10:
Hey everyone 😅,
I've just finished up MessageKit `v0.7.0` and think that this release was a big step in the right direction. Hopefully, with a few more minor versions we will have something of production quality.

### Improvements / Changes
- New `MessageStyles` supporting iMessage & Google Hangouts style bubbles
- Default `DateHeaderView` is shown on new messages received after a 1hr time interval
- `incomingAvatarSize` & `outgoingAvatarSize` deprecated in favor of a more dynamic method
- Support for `avatarAlwaysLeading` and `avatarAlwaysTrailing` layouts
- Removes `bottomLabelExtendsPastAvatar` and `topLabelExtendsPastAvatar` flags
- Replaced with ability to adjust to cellTopLabel/cellBottomLabel to 5 positions: `.cellLeading`, `.cellTrailing`, `.cellCenter`, `.messageLeading`, `.messageTrailing`
- Renames `AvatarPosition` -> `AvatarAlignment`
- Renames `MessagesDisplayDataSource` -> `MessagesDisplayDelegate`
- Reorganizes `MessagesLayoutDelegate`, `MessagesDataSource` and `MessagesDisplayDelegate`
to better represent the methods they contain
- The `MessageCellDelegate` now conforms to `MessageLabelDelegate`
- `MessagesDisplayDelegate` (prev `MessagesDisplayDataSource`) no longer conforms to `MessagesDataSource`
- Added some availability attributes for anyone who has potentially started using the project
- Fixes the text rendering delay for the `MessageInputBar`

### Thanks to
@oyalhi For working on the message styles 
@antonijap For providing the images for the message bubbles
@nathantannar4 For some bugfixes related to the `MessageInputBar`

## Contributing
Looking to get involved in the project? Come join us on at the [MessageKit Slack Channel](https://messagekit.slack.com/join/shared_invite/MjI4NzIzNzMyMzU0LTE1MDMwODIzMDUtYzllYzIyNTU4MA)

hyouuu:
Great work!!

sergpol:
Thank Your for You great work!!!
I try to use this library, in POD file:
pod 'MessageKit'
On ver 0.5.0 all works fine but labels did not align at right side (always on left side).
Ver 0.7.0 when i run chat controller get an exception: 
fatal error: unexpectedly found nil while unwrapping an Optional value
[...]

PR #112, opened 2017-09-11
Update assets to be included within the Podspec

Sherlouk:
Fixes issue raised here: https://github.com/MessageKit/MessageKit/issues/110#issuecomment-328442664

Tested in the example project (powered by a Podfile rather than referencing it directly - I would recommend doing this, happy to make a follow up PR)

Moved assets out of a .bundle as CocoaPods will create this for us (and don't want bundles inside of bundles!)

Also first contribution hype!


Sources/Bundle+Extensions.swift
```diff
@@ -27,9 +27,17 @@ import Foundation
 extension Bundle {
 
     static func messageKitAssetBundle() -> Bundle {
-        let bundle = Bundle(for: MessagesViewController.self)
-        let resourceUrl = bundle.resourceURL!.appendingPathComponent("MessageKitAssets.bundle")
-        return Bundle(url: resourceUrl)!
+        let podBundle = Bundle(for: MessagesViewController.self)
+        
+        guard let resourceBundleUrl = podBundle.url(forResource: "MessageKitAssets", withExtension: "bundle") else {
+            fatalError("MessageKit: Could not create path to the assets bundle")
```
SD10, line 33: Great, much more descriptive

SD10 (approved):

SD10:
@Sherlouk I appreciate you taking care of this so quickly! 🎉 I'd definitely accept another PR reconfiguring our project structure to catch issues like this 😃

PR #114, opened 2017-09-11
Update path for images to fix issue where bubble images don't load

Sherlouk:
Follow up to #112 

My bad for missing it, typical on my first contribution 🙄


SD10 (approved):
@Sherlouk No worries, I'm actually relieved this is the problem. Equally my fault because I missed this in review 😅 Thanks for the follow up!

PR #119, opened 2017-09-11
Add Danger to CI

SD10:


SD10:
This has gone stale and we switched to buddybuild. Will do this after `v0.9.0`

PR #127, opened 2017-09-16
Added Slack and iMessage setup code to Example project

nathantannar4:


SD10:
@nathantannar4 I'll review this a bit later. 
If you want what you can do is add more rows to the initial tableView in the example app:
ex) `iMessage`, `Slack` then design those screens
We're also probably capable of doing a light version of Google Hangouts

Sources/MessageInputBar.swift
```diff
@@ -396,7 +396,7 @@ open class MessageInputBar: UIView {
     /// - Parameters:
     ///   - newValue: New widthAnchor constant
     ///   - animated: If the layout should be animated
-    open func setLeftStackViewWidthContant(to newValue: CGFloat, animated: Bool) {
+    open func setLeftStackViewWidthConstant(to newValue: CGFloat, animated: Bool) {
```
SD10, line 399: 👍

SD10 (approved):
Great work as usual @nathantannar4 🎉

PR #125, opened 2017-09-14
Refactoring keyboard handling

SD10:
### Changes:
- No longer manipulates the top `contentInset` of the collection view. This was blocking #117 
- Reduce number of notifications observed
- Only add a messageInputBarCopy if isFirstLayout

PR #128, opened 2017-09-16
Bugfix/message input bar rendering

nathantannar4:


SD10:
@nathantannar4 Can you take a look at #125 and see if we can merge this into that branch?
I'm thinking that we may be able to combine this flag with the `isFirstLayout` flag I introduced there.
If it's not too much trouble you could also fix up the one pull request review I made there as well.

Issue #129, opened 2017-09-18
InputTextView sizes incorrectly for long messages


SD10:
![screen shot 2017-09-16 at 8 00 24 pm](https://user-images.githubusercontent.com/7445580/30526251-2eaadf80-9bdd-11e7-90d8-5cbccf620c41.png)

### How to reproduce:
1. Paste or type a large amount of text inside of the `UITextView`. 

Something that will be greater than the the `maxWidth` allowed when calculating the `intrinsicContentSize`

**Note**: Reaching a `maxWidth` with new lines don't trigger this bug

Ganzolo:
I just battled a lot with this issue. I think I found a fix with minimum impact on current code. I think the problem relies on UIKit. What I did is getting rid of intrinsicContentSize to change the height of view and only relies on :

- Content Compression resistance
- layout height constraint

nathantannar4:
Hey,
I looked over what you had and you left out some things. I don't think changing the entire intrinsicContentSize is needed. I have been doing some texting and I think adding a heightconstaint just to the InputTextView when height > maxHeight is all we need

SD10:
This is fixed in #173

PR #135, opened 2017-09-18
Issue #129

nathantannar4:


Sources/MessageInputBar.swift
```diff
@@ -424,14 +424,14 @@ open class MessageInputBar: UIView {
     }
     
     open func textViewDidChange() {
-        let trimmedText = inputTextView.text.trimmingCharacters(in: .whitespacesAndNewlines)
```
SD10, line 427: Was it necessary to remove this? I like how the send button is enabled/disabled by taking into account if a message only contains white space

SD10:
@nathantannar4 I'm going to close this because it doesn't fix the issue and it changes the functionality around enabling/disabling the send button

Issue #134, opened 2017-09-18
Check if carthage installation is working


SD10:
Carthage support was probably broken by the new project structure regarding MessageKit assets 🤔

cwalo:
I can confirm that I'm able to get this working using Carthage with `0.7.3`, but the images are not found.

cwalo:
After `0.7.4`, looks like we're in good shape.

SD10:
Great 😃 Sorry I signed up for ZenHubIO and it reopened a ton of issues...

PR #139, opened 2017-09-20
Add missing resources for Carthage

SD10:
Hopefully a fix for #134 😅


SD10:
@marlontojal This is my attempt to solve your problem. I'm not a Carthage user so any input you have is appreciated 😃

Sherlouk:
Fingers crossed 🙏🏼 Wonder if there’s a way to auto validate these installation methods for both pods and Carthage 🤔

PR #150, opened 2017-09-22
LocationMessageDisplayDelegate to add custom annotations

etoledom:
Added LocationMessageDisplayDelegate to add custom annotations to locationMessages
Modified example proyect implementing LocationMessageDisplayDelegate with a custom annotation


Sources/MessagesCollectionView.swift
```diff
@@ -36,6 +36,8 @@ open class MessagesCollectionView: UICollectionView {
 
     open weak var messageCellDelegate: MessageCellDelegate?
 
+    open weak var locationMessagesDisplayDelegate: LocationMessageDisplayDelegate?
```
SD10: We don't need this property because we can assign it to `messagesDisplayDelegate` since `LocationMessageDisplayDelegate` conforms to `MessagesDisplayDelegate`

Sources/LocationMessageCell.swift
```diff
@@ -69,11 +70,32 @@ open class LocationMessageCell: MessageCollectionViewCell<UIImageView> {
 
         let snapShotter = MKMapSnapshotter(options: snapshotOptions)
         snapShotter.start { (snapshot, error) in
-            self.activityIndicator.stopAnimating()
-            if error == nil {
-                self.messageContentView.image = snapshot?.image
+
+            guard let snapshot = snapshot, error == nil else {
+                //show an error image?
+                return
```
SD10: Need to call `stopAnimating()` here

etoledom: We could use defer {} to never forget again 😬

Sources/MessagesCollectionView.swift
etoledom: Got it! Thanks!

Sources/LocationMessageCell.swift
```diff
@@ -69,11 +70,33 @@ open class LocationMessageCell: MessageCollectionViewCell<UIImageView> {
...
+            snapshot.image.draw(at: .zero)
+
+            var point = snapshot.point(for: location.coordinate)
+            //Move point to reflect annotation anchor
+            point.x -= annotationView.bounds.size.width / 2
+            point.y -= annotationView.bounds.size.height / 2
+            point.x += annotationView.centerOffset.x
+            point.y += annotationView.centerOffset.y
+
+            annotationView.image?.draw(at: point)
```
SD10, line 97: My only question is what if the user is not using an image for the `MKAnnotationView`? It seems like we're not using `MKAnnotationView` at all really. It's just wrapping up our `UIImage`. Am I making sense?

etoledom: `MKAnnotationView` is not only wrapping the image but also the anchor information necessary for drawing `annotationView.centerOffset`, so we shouldn't just ask for an image in the delegate.

But is true that this API how it is now, allows the mistake of not adding an image, so maybe we could use a simpler wrapping Struct with `image` and `centerOffset` making the image required.

I also thought some users might not want so show an annotation at all, so we could make the delegate return optional adding the possibility of a map without annotation.

SD10: @etoledom Isn't there some way to add the `MKAnnotationView` as a whole?

SD10: What I'm trying to say is what if I want to style my `MKAnnotationView` using the properties of `UIView` and not just a `UIImage`

etoledom: @SD10 I see what you mean.

We can actually draw the annotationView's layer with:
`annotationView.layer.render(in: context)`
I tested it and seems to work fine.

For some reason, every tutorial and example I have found uses the image drawing thing, but the Apple documentation on `MKAnnotationView` says that is allowed also to draw the pin overriding `draw(_:)`, so in that case probably will be the best, as you said, to draw the whole view instead of just the image for maximum flexibility.

I will test it a bit more, and PR with this little drawing change.

etoledom: Works with `MKAnnotationView`, `MKMarkerAnnotationView`, custom subclasses of `MKAnnotationView`, but it doesn't want to work with `MKPinAnnotationView` (our default) 😅

Sources/LocationMessageCell.swift
```diff
@@ -69,11 +70,35 @@ open class LocationMessageCell: MessageCollectionViewCell<UIImageView> {
...
         let snapShotter = MKMapSnapshotter(options: snapshotOptions)
         snapShotter.start { (snapshot, error) in
-            self.activityIndicator.stopAnimating()
-            if error == nil {
-                self.messageContentView.image = snapshot?.image
+            defer {
+                self.activityIndicator.stopAnimating()
             }
+            guard let snapshot = snapshot, error == nil else {
+                print("\(#function) Error creating map snapshot: \(String(describing: error))") //show an error image?
```
SD10: We don't want to `print` anything in production. Let's just return here

etoledom: Sure!, I'll fix it now

SD10 (approved):

SD10:
@etoledom Thanks for your great work on this 💯 👍 I've invited you to the MessageKit organization, it would be great to have your involvement in our project 😃 No worries if you're not feeling up to it!

PR #155, opened 2017-09-24
[WIP] Making copy feature using partial Collection view mechanism. Could no…

Ganzolo:
…t rely entirely on Collection view MenuController has it does not support customisation of target frame.

### DIFF
--- a/Example/ChatExample.xcodeproj/project.pbxproj
+++ b/Example/ChatExample.xcodeproj/project.pbxproj
@@ -346,7 +346,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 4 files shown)


Ganzolo:
After doing some researches, to implement this feature there seems to be 2 different ways :

1. Implement it with a gesture on UICollectionViewCell's level
2. Implement it from the controller using CollectionView's Menu controller (same as @hamzaozturk suggested)

First option give us more control in the way we display menu controller. Especially controlling its location. However it has a big drawback as the copy action would happen within cell subclass. Meaning that we need to have a reference to Collection View controller to grasp the data from Data source. I consider as inappropriate to just copy content of cell.

Second option occurs in UICollectionViewDelegate, which is a perfect place to have access to raw data. However it has a huge drawback, it seems that there is no support to adjust the position of the menu controller. Therefor it is just displayed above the whole cell. In our case we have bubbles and we would like to have menu controller centred above our bubbles.

After doing researches and experiment I found out a trick using Menu Controller's notifications to override the position. I don't see any other way of doing it. But maybe someone has a better proposal.

SD10:
@Ganzolo Sorry for the delay. I hope to review this really soon! Thank you for taking the time to work on this 👍

SD10:
Closing in favor of #248

PR #169, opened 2017-09-26
Issue #129 : Removing intrinsic content size and using NSLayoutContra…

Ganzolo:
…int without animation remove copy/paste glitch


SD10:
@nathantannar4 Can this be closed in favor of #173?

nathantannar4:
Yes it can be

PR #173, opened 2017-09-27
Height Fix for MessageInputBar during paste

nathantannar4:


SD10 (approved):
I'm a little concerned about making the InputTextView it's own delegate but let's move forward with this for now. 

Let's add a CHANGELOG entry for **Fixed** and lets also add a **Breaking Change** under **Changed** letting people know the `InputTextView` is now its own delegate

PR #175, opened 2017-09-27
Make header and footer view initializers public

cwalo:
In order to subclass MessageHeaderView and MessageFooterView, the initializers need to be marked public. Tangential, but I vote we stick with this order moving forward:
`<protection level> <designation>`

### DIFF
--- a/Sources/MessageDateHeaderView.swift
+++ b/Sources/MessageDateHeaderView.swift
@@ -32,7 +32,7 @@ open class MessageDateHeaderView: MessageHeaderView {
 
     // MARK: - Initializers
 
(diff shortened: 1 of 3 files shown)


fbartho:
I like this style, personally, I usually add @objc on the preceding line if it becomes necessary

SD10:
SwiftLint has an open PR for this rule I believe. It's driving me crazy though. After the initial release I had hoped to make a starter issue for someone to lint it for me 😅

SD10 (approved):
Thank you @cwalo 🎉 I've invited you to join the MessageKit organization. No pressure if you're not up to it 😃

Issue #164, opened 2017-09-25
Setting an Avatar asynchronously


SD10:
There's currently no clear way of doing this and the priority of this feature is extremely high. The question is how to expose an API for this without:

- A) Adding any networking code to the library
- B) Making any assumption on how the user manages async images, `SDWebImage`, `Kingfisher`, etc.

It would be best if we just had some sort of hook to retrieve an Avatar and at the very most all we know is this method is going to be performed asynchronously.

mperovic:
@SD10 I implemented it using Nuke + Preheat. It works great. By my opinion option B is a quite acceptable solution.

SD10:
@mperovic Thanks for the feedback. I know solving this is sort of user-specific, but I still think there should be some form of very basic first class support in the library. Glad to hear you were able to work things out.

That being said, the worst thing we can do is force users to use a specific method or add a specific 3rd party dependency.

eliburke:
I agree strongly. Sub-dependencies are a huge pain, and everyone's backend requirements are different. 

I built something like the following for a project I am working on:
* An NSCache-derived AvatarCache and <CachedAvatar> container object
* CachedAvatar stores identifier, initials, optional avatar, optional placeholder, updateInProgress, lastUpdateAttempt
* avatar lookup takes the ID and flags indicating if placeholder / initial images are acceptable
* if initials are enabled, placeholder image is disabled, and vice-versa
* avatar fetch is async

In practice: the view controller (or cell or whoever) requests an avatar, and indicates if initials / placeholder is OK. Cache looks up the ID and returns it if available, or kicks off an async query and then generates an AvatarInitials image and returns it. Next call to the cache will get AvatarInitials image immediately. Eventually when the async query completes, the avatar is written to the cache, so when the view controller next requests the avatar it gets the real deal. 

In my particular case, the async query is a method but I think it could instead be an optional closure for users to override. Also in my case, avatars are written to a Realm database so it is easy to watch for update notifications. To accommodate other uses, a weak reference of the AvatarCache could be passed to the fetch closure, so the resulting image can be written into the cache, and then the cache can notify MessageKit. There would probably also need to be a closure for "saving" the avatar to database / filesystem. [Edit] one more wrinkle is cancellation. I use my implementation globally so I don't mind outstanding requests, but a way to cancel could be useful, or it could be left to subclassing. 

Some other thoughts: 
[...]

PR #176, opened 2017-09-27
Issue #164: Setting avatars asynchronously

cwalo:
I've added an additional `MessagesDatasourceDelegate` method that returns an optional closure. If the closure is returned, we'll call the closure when configuring the cell, otherwise we use `avatarView.set(avatar: avatar)` with the avatar returned from the non-optional delegate method - `avatar(for: message...)`.


SD10:
Going to close this because we've been discussing another method, but thanks for paving the way @cwalo!

PR #177, opened 2017-09-27
Modify protection level of UICollectionViewDataSource methods

cwalo:
Making these methods `open` allows for more flexibility, but _does_ have the potential for more user error when not used as intended.


SD10 (approved):
Great 👌

PR #178, opened 2017-09-28
A proper fix to the rendering bug from #128

nathantannar4:


Sources/MessagesViewController.swift
```diff
@@ -63,17 +61,10 @@ open class MessagesViewController: UIViewController {
...
         super.viewWillAppear(animated)
-        if isFirstLayout {
-            setupMessageInputBarCopy()
-        }
-    }
-
-    open override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-        removeMessageInputBarCopy()
+        messageInputBar.inputTextView.setNeedsDisplay()
```
SD10: Was `setNeedsDisplay` left intentionally?

nathantannar4: oh no that was from testing, shouldn't be there

Issue #180, opened 2017-09-28
non-Transculent navigation bar break Messages collection view


Ganzolo:
If your project have a non translucent navigation bar it is not possible to scroll properly. CollectionView cannot scroll to the bottom of the screen.

![simulator screen shot - iphone 8 - 2017-09-28 at 14 14 49](https://user-images.githubusercontent.com/13142882/30966197-6580bdee-a458-11e7-8392-40ca6f9f41ca.png)

PR #181, opened 2017-09-28
fix proposal for #180

Ganzolo:


SD10 (approved):

SD10:
@Ganzolo Thanks for fixing this 👍 💯 I've invited you to join the MessageKit organization. No pressure if you're not up to it 😎

PR #194, opened 2017-09-29
[WIP] Added a top UIStackView to MessageInputBar

nathantannar4:
This is to support features like @mention and a typing indicator. Updated the Example to show changes.

There are some **breaking changes** but only with the backgroundView of the MessageInputBar to support difference background colors behind the top stack view and main content


SD10:
I don't mind adding another `UIStackView` but something doesn't feel right about having the typing indicator inside of the `MessageInputBar`. I really don't want these two components coupled 😕

The interaction between the keyboard/message input bar and the collectionview is already really painful. I'm not sure I want to add more state management to the part of the library.

Let's leave this open and see what others think.

Sources/MessageInputBar.swift
```diff
@@ -27,27 +27,39 @@ import UIKit
...
     
     /// When set to true, the blurView in the background is shown and the backgroundColor is set to .clear. Default is FALSE
     open var isTranslucent: Bool = false {
         didSet {
-            blurView.isHidden = !isTranslucent
-            backgroundColor = isTranslucent ? .clear : .white
+            if isTranslucent && blurView.superview == nil {
+                backgroundView.addSubview(blurView)
+                blurView.fillSuperview()
+            } else if !isTranslucent {
```
SD10: It's probably best just to add this once if they set`isTranslucent` to `true`

Sources/MessageInputBar.swift
```diff
@@ -89,6 +110,8 @@ open class MessageInputBar: UIView {
         return view
     }()
     
+    open let typingIndicatorLabel = UILabel()
```
SD10: We need to remove this as we discussed

SD10 (changes requested):
Let's just leave the top stack view and I think this is good :P

nathantannar4:
Sorry I accidentally committed a merge to this PR. Its for autocomplete

SD10:
No worries 😃 I know the `MessageInputBar` is your jam but I really need to limit what goes into `v0.9.0`. This is also ~500 lines of code now so it can become difficult for me to review. Let's try and push enhancements of the `MessageInputBar` until after `v0.9.0`. Bugfixes are welcome any time

ohazar:
Hello, what is the status of this change? Is it going to be added or canceled? Ticket looks "closed" but its label is "in progress".

Issue #193, opened 2017-09-29
Lint project for access control & override order


SD10:
SwiftLint doesn't have a rule for this yet and the inconsistency is driving me crazy:
```Swift
// Good
public override func foo() {}

// Bad
override public func foo() {}
```

tapanprakasht:
@SD10 I have written some custom swiftlint rule to handle mentioned the issue. This can be enhanced further. This is a just initial step. May it won't recognize all the possible scenarios as of now. After applying the rule project shows 24 new warnings. 
Can I fix that warning?

tapanprakasht:
@SD10  I hope that requirements mentioned here are fulfilled by the PR #196. I think this issue can be closed now.

SD10:
Great, closing this 👍

PR #195, opened 2017-09-29
New swiftlint rule to manage proper placing of override keyword

tapanprakasht:
This pull request is to handle new swiftlint rule mentioned in #193 issue.
Currently, this rule will take care of override open or override public written in the code and show a warning message.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -8,6 +8,6 @@ disabled_rules:
 custom_rules:
   override_func: # rule identifier
     name: "override in func" # rule name. optional.
-    regex: "(override open) | (override public)" # matching pattern
+    regex: "override (open|public|private|internal|fileprivate)" # matching pattern
     message: "Use like open override or public override instead" # violation message. optional.
     severity: warning # violation severity. optional.
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -5,3 +5,9 @@ disabled_rules:
 - line_length
 - type_body_length
 - file_length
+custom_rules:
+  override_func: # rule identifier
+    name: "override in func" # rule name. optional.
+    regex: "(override open) | (override public)" # matching pattern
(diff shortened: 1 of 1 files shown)


.swiftlint.yml
```diff
@@ -5,3 +5,9 @@ disabled_rules:
 - line_length
 - type_body_length
 - file_length
+custom_rules:
+  override_func: # rule identifier
+    name: "override in func" # rule name. optional.
+    regex: "(override open) | (override public)" # matching pattern
```
Sherlouk: Maybe a regex like `override (open|public|private|internal|fileprivate)` would be better to pick up all access control keywords and to prevent duplication of override?

tapanprakasht: Thanks for the suggestion @Sherlouk. I have made required changes again.

SD10:
I had no idea you could write your own rules with regex 🤔

Sherlouk:
@SD10 https://github.com/realm/SwiftLint#defining-custom-rules

Quite powerful to be fair!

May be an idea to add a match_kinds value, I couldn't tell you which is the correct one for what we want - but it'd be nice not to have it pick up the use of "override open" if I put it in a comment for example!

SD10 (changes requested):
@tapanprakasht Just a few things. Can you do this on the `v0.9.0` branch. I changed the PR to target that branch.

Can you run the script and make the corrections?

tapanprakasht:
@SD10  Should I create a separate pull request to do this? 
I can create a separate pull request for making the correction from v0.9.0 as base branch. Would it be fine? 
The current PR doesn't have merge conflict with v0.9.0 branch.

SD10:
@tapanprakasht Whatever works best for you is fine by me 😃

tapanprakasht:
@SD10  I have created a new pull request #196 which handles your suggestion. So there is no need for this pull request hence closing this.

PR #196, opened 2017-09-30
Fix override keyword placement in function declaration

tapanprakasht:
This pull request contains changes suggested in #195. Changes contain new swiflint rule in .swiftlint.yml file and warnings corrections according to the new rule.


SD10 (approved):
Thank you for taking care of this @tapanprakasht 👍 This closes #193.

PR #197, opened 2017-09-30
Allow MessageLabelDelegate to specify enabled data detectors

hyouuu:
In my use case, I don't want to show data detectors, and this change allows delegates to list what they want, or simply return nil to keep the default behavior

### DIFF
--- a/Sources/MessageCollectionViewCell.swift
+++ b/Sources/MessageCollectionViewCell.swift
@@ -39,7 +39,7 @@ open class MessageCollectionViewCell<ContentView: UIView>: UICollectionViewCell 
 
     open var cellTopLabel: MessageLabel = {
         let topLabel = MessageLabel()
-        topLabel.enabledDetectors = []
+        topLabel.defaultEnabledDetectors = []
         return topLabel
     }()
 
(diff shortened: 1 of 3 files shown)


SD10:
@hyouuu I think this is tricky because it's confusing which detectors the cell is going to be using. I do agree that we can improve this. I'm guessing your problem is you want to modify the detectors but it's too difficult for you to subclass?

hyouuu:
I think most MessageKit users would use a subclass to MessagesViewController e.g.:
`class SomeVC: MessagesViewController {}`
But MessagesViewController doesn't provide a way to override the data detectors for each message, whereas its MessagesCollectionView var has MessagesLayoutDelegate etc, which I feel is the best place to override this. 
Let me know if there is a better way of approaching this or I missed something - thanks!

SD10:
@hyouuu I've given this some thought and I don't think it makes sense that this is declared as part of the `MessageLabelDelegate`. I'm going to close this for now but thank you for making me aware of this limitation. In an upcoming change I will address this and ping you for your input.

hyouuu:
Looking forward to it!

PR #204, opened 2017-10-01
[Fixed] Set extendedLayoutIncludesOpaqueBars to true by default

### LINKED ISSUE #202: MessagesCollectionView content inset sizing wrong for opaque navigation bars
A lot of people are experiencing an issue where the content inset of `MessagesCollectionView` is sized wrong for controllers with an opaque navigation bar. 

This can be corrected by setting:
```Swift
extendedLayoutIncludesOpaqueBars = true
```
Enough users have complained about this that I think it should be default MessageKit behavior.

SD10:
### TODO:
- [ ] CHANGELOG entry

This resolves #202


cwalo:
👍

Issue #207, opened 2017-10-01
The toolbar is placed under safe area (iOS 11 iphoneX)


samsonwyt:
I have tried using iphone X to test the example, and the input bar is placed under the iphone X safe area!

[https://imgur.com/a/O2s5d](url)

alexbinary:
@samsonwyt your link is broken, looks like a Markdown problem.
Here is a working link : https://imgur.com/a/O2s5d

nathantannar4:
Fixed in v0.10.0

PR #209, opened 2017-10-01
[WIP] [Fixed] iPhone X inputAccessoryView layout issue

SD10:
### TODO:
- [ ] CHANGELOG entry

Would solve #207

Not sure if this is expected behavior for inputAccessoryView. Looks like there is a radar here:
http://www.openradar.me/34411433

The view background color is set to the color of the `MessageInputBar`

### DIFF
--- a/Sources/MessageInputBar.swift
+++ b/Sources/MessageInputBar.swift
@@ -25,6 +25,17 @@
 import UIKit
 
 open class MessageInputBar: UIView {
+
+    // MARK: - iPhone X Layout Fix
+    // Could be fixed by Apple soon
+    // https://stackoverflow.com/questions/46282987/iphone-x-how-to-handle-view-controller-inputaccessoryview
+    open override func didMoveToWindow() {
+        if #available(iOS 11.0, *) {
+            if let window = window {
+                bottomAnchor.constraintLessThanOrEqualToSystemSpacingBelow(window.safeAreaLayoutGuide.bottomAnchor, multiplier: 1.0).isActive = true
+            }
+        }
+    }
     
     public enum UIStackViewPosition {
         case left, right, bottom
--- a/Sources/MessagesViewController.swift
+++ b/Sources/MessagesViewController.swift
@@ -116,11 +116,16 @@ open class MessagesViewController: UIViewController {
 
     private func setupConstraints() {
         messagesCollectionView.translatesAutoresizingMaskIntoConstraints = false
-
         let top = messagesCollectionView.topAnchor.constraint(equalTo: view.topAnchor, constant: topLayoutGuide.length)
         let leading = messagesCollectionView.leadingAnchor.constraint(equalTo: view.leadingAnchor)
         let trailing = messagesCollectionView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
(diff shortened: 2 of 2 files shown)


nathantannar4:
I can't believe apple didn't add native support for this so we don't have to

nathantannar4 (changes requested):
With the changes i made in my latest PR the color you will need is: messageInput.backgroundView.backgroundColor

Sources/MessagesViewController.swift
```diff
@@ -116,11 +116,16 @@ open class MessagesViewController: UIViewController {
...
         let top = messagesCollectionView.topAnchor.constraint(equalTo: view.topAnchor, constant: topLayoutGuide.length)
         let leading = messagesCollectionView.leadingAnchor.constraint(equalTo: view.leadingAnchor)
         let trailing = messagesCollectionView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
-        let bottom = messagesCollectionView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        var bottom = messagesCollectionView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+
+        // iPhone X workaround
+        if #available(iOS 11.0, *) {
+            view.backgroundColor = messageInputBar.backgroundColor
+            bottom = messagesCollectionView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)
```
alexbinary, line 127: I believe `view.layoutMarginsGuide.bottomAnchor` would work as well.

`layoutMarginsGuide` is available since iOS 9.

Sources/MessageInputBar.swift
```diff
@@ -25,6 +25,17 @@
 import UIKit
 
 open class MessageInputBar: UIView {
+
+    // MARK: - iPhone X Layout Fix
+    // Could be fixed by Apple soon
```
alexbinary, line 30: I don't think this is a bug. See my answer on the SO post for a cleaner fix.

nathantannar4:
I notice that when adding those constraints it breaks stuff. Another option is to increase the bottom padding and then reduce it when the keyboard becomes the first responder

SD10:
I don't like any of this to be honest lol. Supporting the iPhone Notch is low priority to me this weekend :/

etoledom:
@nathantannar4 @SD10 Try this:

If you change the line 280 on `MessageInputBar.swift` to:
`bottom: bottomStackView.bottomAnchor.constraint(equalTo: layoutMarginsGuide.bottomAnchor, constant: -padding.bottom),`
The inputView seems to be in the correct position for the iPhone X.
Is not the definitive answer since the messagesCollectionView bottom inset is still wrong, but I think this could be a better start.

The change would be to pin the lower subview of MessageInputBar to `layoutMarginsGuide.bottomAnchor` instead to the `.bottomAnchor`

etoledom:
simulator screen shot - iphone2017-c - 2017-10-01 at 16 46 26

etoledom:
Anything new about this? I could check it more in detail

SD10:
@etoledom Be my guest 👍 To be honest, I haven't read through everything here in great detail. Other more pressing issues taking precedence

PR #210, opened 2017-10-01
Added custom animation support for displaying map on location messages.

etoledom:


Sources/LocationMessageDisplayDelegate.swift
```diff
@@ -50,6 +50,19 @@ public protocol LocationMessageDisplayDelegate: MessagesDisplayDelegate {
...
+    /// The animation block is called with the view to be animated. You can animate it with CoreAnimation, UIView.animate or any library you prefer.
+    ///
+    /// default: nil
+    ///
+    /// - Parameters:
+    ///   - message: The location message with the map to animate
+    ///   - indexPath: Message's index path
+    ///   - messagesCollectionView: The collection view requesting the information
+    /// - Returns: Your customized animation block.
+    func animationBlockForLocation(message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> ((UIView) -> Void)?
```
SD10: I think that the `LocationMessageCell`s `ContentView` is a `UIImageView` and not a `UIView`? We may want to represent this in the block

etoledom: Sure! I will change that 🙂

SD10:
@etoledom We also need a CHANGELOG entry following the same guidelines we are using in `v0.9.0`.
 
This would go under the `Added` section and doesn't need to be marked as a breaking change. The `Added` section word should act like your first word in the CHANGELOG message.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - `messageLabelInsets(for:indexPath:messagesCollectionView` method to `MessagesLayoutDelegate`. 
 [#162](https://github.com/MessageKit/MessageKit/pull/162) by [@SD10](https://github.com/SD10).
 
+-`animationBlockForLocation(message:indexPath:messagesCollectionView)` method to `LocationMessageDisplayDelegate` to customize the display animation of the location message's map
```
SD10: 1. Can you add a space so we have a proper bullet point? 😉 
2. Can you end every sentence with a period? 😂 
3. Can you mark the PR number with the `#` character?

Very nitty, I can change them after merge if you'd like. I will add these things to the CHANGELOG guidelines as well.

etoledom: I will change them, don't worry. Is good to not forget later on 😁

Example/Sources/ConversationViewController.swift
```diff
@@ -333,6 +333,17 @@ extension ConversationViewController: LocationMessageDisplayDelegate {
...
 
+    func animationBlockForLocation(message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> ((UIImageView) -> Void)? {
+        return { view in
+            view.layer.transform = CATransform3DMakeScale(0, 0, 0)
+            view.alpha = 0.0
+            UIView.animate(withDuration: 0.6, delay: 0, usingSpringWithDamping: 0.9, initialSpringVelocity: 0, options: [], animations: {
+                view.layer.transform = CATransform3DIdentity
+                view.alpha = 1.0
+            }, completion: nil)
+        }
```
SD10, line 344: We may change this default animation but merging for now!

SD10 (approved):
I appreciate you working on this @etoledom 💯

Issue #183, opened 2017-09-28
Add scrollsToBottomOnFirstLayout flag


SD10:
It would be nice to add a property to `MessagesViewController` that allows the user to scroll to the bottom of the collection view on the first load.

This probably needs to be set inside of the `isFirstLayout` block inside of `viewDidLayoutSubviews()`:

```Swift
if isFirstLayout {
/// ...
if scrollsToBottomOnFirstLayout { messagesCollectionView.scrollToBottom() }
}
```

Jeyhey:
That's a very good idea. For the moment you would have to implement it in `viewWillAppear` of the subclass like so for instance:
```
override func viewWillAppear(_ animated: Bool) {
        super.viewWillAppear(animated)
        messagesCollectionView.layoutIfNeeded()
        maybeGoToBottom()
}

func maybeGoToBottom(_ animated: Bool = false) {
        guard goToBottom else {
            return
        }
        goToBottom = false
        let contentSize: CGSize = messagesCollectionView.contentSize
        if contentSize.height > messagesCollectionView.bounds.size.height {
            let targetContentOffset = CGPoint(x: 0, y: contentSize.height - messagesCollectionView.bounds.size.height + navigationController!.navigationBar.bounds.height)
            messagesCollectionView.contentOffset = targetContentOffset
            messagesCollectionView.setContentOffset(targetContentOffset, animated: animated)
        }
}
```

PR #213, opened 2017-10-02
- added open var scrollsToBottomOnFirstLayout

FraDeliro:
Added open var scrollsToBottomOnFirstLayout, if true messagesCollectionView scrollToBottom.

Thank you for your replay, it's one my first contributions, I have made the changes you have requested.
Let me know if everything is ok.
Thanks ;)
#183


Sources/MessagesViewController.swift
```diff
@@ -82,6 +84,13 @@ open class MessagesViewController: UIViewController {
             addKeyboardObservers()
             messagesCollectionView.contentInset.bottom = messageInputBar.frame.height
             isFirstLayout = false
+            
+            //scroll to bottom at first load
+            guard scrollsToBottomOnFirstLayout == true else {
```
SD10: Can we just make this read as:
```Swift
if scrollsToBottomOnFirstLayout {
    messagesCollectionView.scrollsToBottom(animated: false)
}
```

I'd also like explicitly set `isFirstLayout` at the very end of this scope

Sources/MessagesViewController.swift
```diff
@@ -82,6 +84,13 @@ open class MessagesViewController: UIViewController {
...
+                return
+            }
+            scrollsToBottomOnFirstLayout = false
```
SD10: let's not modify the bool after we scroll to bottom, all we're really concerned with is the `isFirstLayout` flag

SD10 (changes requested):
@FraDeliro Thanks for taking care of this 💯 Just a few nit revisions. Can you also add a CHANGELOG entry matching the same style we are using in the `v0.9.0` branch CHANGELOG? This would go under **added** and is not a breaking change

SD10 (approved):
@FraDeliro Thank you for taking care of this for us 👍

FraDeliro:
@SD10 Thanks for your library ;)

fishcharlie:
What is the status of this? This code doesn't look to be in the library currently.

Issue #165, opened 2017-09-25
messagesCollectionView is hidden by keyboard


Jeyhey:
When the soft keyboard is shown messagesCollectionView does not scroll up with the keyboard.

I guess the code that is supposed to handle is the line
 `messagesCollectionView.contentInset.bottom = bottomInset` 
in `handleKeyboardDidChangeState()`.

I don't know why, but I can only convince the collection view to move up when I insert a call to `messagesCollectionView.setContentOffset(someOffset)` right below this line.

SD10:
@Jeyhey You can still view all the content, correct? You're just talking about how the user has to manually scroll to the bottom of the `MessagesCollectionView`?

You can scroll to the bottom of `MessagesCollectionView` using the `scrollToBottom(animated:)` method. We probably want to add a flag to determine if this is done automatically when the keyboard is shown.

alijaza:
+1 The bottom contentinset is giving me all sorts of trouble. I can't get the messageInputBar to stop covering the CollectionView no matter what I do. Seems even if I set it view the view loads, handleKeyboardDidChangeState eventually changes it back to the wrong value. For some reason, when using the slack keyboard from the example project the messageInputBar height has the wrong value when this method fires.

Jeyhey:
@SD10 no the cells get hidden by the keyboard. I was talking about the behaviour of the collection view when the keyboard gets shown. I would expect some standard behaviour implemented in the base class of MessageViewController.

Following the lines of [https://developer.apple.com/library/content/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/KeyboardManagement/KeyboardManagement.html](url) I get the intuitive behaviour by changing the method to:
```
func handleKeyboardDidChangeState(_ notification: Notification) {

        guard let keyboardEndFrame = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }

        if (keyboardEndFrame.origin.y + keyboardEndFrame.size.height) > view.frame.size.height {
            // Hardware keyboard is found
            messagesCollectionView.contentInset.bottom = view.frame.size.height - keyboardEndFrame.origin.y
        } else {
            //Software keyboard is found
            let bottomInset = keyboardEndFrame.height > messageInputBar.frame.height ? keyboardEndFrame.height : messageInputBar.frame.height
            messagesCollectionView.contentInset.bottom = bottomInset
        }
        var aRect = self.view.frame
[...]

PR #217, opened 2017-10-03
[Added] scrollsToBottomOnKeyboardDidBeginEditing property

### LINKED ISSUE #203: Flags to perform automatic scrolling
### Flags to perform automatic scrolling
It would be nice to have two flags for users to control scrolling:
```Swift
var scrollToBottomOnKeyboardBeginsEditing = false
var scrollToBottomOnFirstLoad = true
```

@cwalo I think if we put the `messagesCollectionView.scrollToBottom()` call inside of the `isFirstLayout` flag in `viewDidLayoutSubviews` we don't have to do anything with calculating the `contentOffset`?

SD10:
This closes #203 && #165

### DIFF
--- a/Sources/MessagesViewController.swift
+++ b/Sources/MessagesViewController.swift
@@ -34,6 +34,8 @@ open class MessagesViewController: UIViewController {
     
     open var scrollsToBottomOnFirstLayout: Bool = false
 
+    open var scrollsToBottomOnKeybordBeginsEditing: Bool = false
+
     private var isFirstLayout: Bool = true
 
     open override var canBecomeFirstResponder: Bool {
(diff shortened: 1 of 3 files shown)


Sources/MessagesViewController.swift
```diff
@@ -240,15 +243,23 @@ extension MessagesViewController: UICollectionViewDataSource {
 
 extension MessagesViewController {
 
+
     fileprivate func addKeyboardObservers() {
         NotificationCenter.default.addObserver(self, selector: #selector(handleKeyboardDidChangeState), name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(handleTextViewDidBeginEditing), name: .UITextViewTextDidBeginEditing, object: messageInputBar.inputTextView)
```
Sherlouk, line 258: Out of interest, why are you choosing to add an observer here over assigning the delegate of the textView?

Sources/MessagesViewController.swift
```diff
@@ -240,15 +243,23 @@ extension MessagesViewController: UICollectionViewDataSource {
...
     }
 
     fileprivate func removeKeyboardObservers() {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillChangeFrame, object: nil)
```
Sherlouk, line 262: Do we need to remove the notification `.UITextViewTextDidBeginEditing` here?

SD10: Only for some versions of iOS. Good catch 😉

Sources/MessagesViewController.swift
SD10: I thought it would be better to not reserve making the delegate our controller

Issue #182, opened 2017-09-28
Add emoji message type


SD10:
I'm wondering if we can create a new case under `MessageData`:
```Swift
case emoji(String)
// OR
case emoji([Character])
```
This could do something like double the size of the emojis like is done on Slack.

We could parse all text messages to do this without adding a new case... but I don't want to perform that operation on every message 😉

SD10:
This is closed by #222

PR #222, opened 2017-10-03
Add support for emoji-only messages

SirArkimedes:
Adds a new message type that handles text that doubles the font size and removes the message background to model how iMessage displays emoji-only messages.

Issue: #182 

Previews of how they look:


SirArkimedes:
I do have a question, though. Should I force the "text" MessageData to use this "emoji" MessageData or should I leave it up to the user?

I know the issue talks about parsing, but was unsure if I should parse the message at all.

SD10:
@SirArkimedes This is a good start!! 👍 I was hoping we didn't have to create another subclass of `MessagesCollectionViewCell` and could just use `TextMessageCell` (I'm trying to decrease the number of different cell types we have). 

I'll be able to review this a little more in-depth later. 

About parsing, nothing to do there. It should be the user's responsibility to determine if a message contains only emojis or not.

cc @cwalo for some extra eyes

fbartho:
Re: parsing, you could add a utility method that trivially detects emoji (minus spaces), and add an example in documentation of how to use wire this together to return the right message type.

eliburke:
If you don't want to embed emoji detection code, I'd suggest a comment pointing users at:
https://stackoverflow.com/a/39425959/1758224

Using these two extensions, the appropriate logic is:
````
if messageText.containsOnlyEmoji, messageText.glyphCount < 5 {
    // do the thing
}
````

**Comment by SD10** [comment]
@fbartho @eliburke Both great points, thank you!!

**Comment by cwalo** [comment]
Seconded @eliburke. Here's a [gist](https://gist.github.com/cwalo/dc54be3ae74020e91ca4b8bbf037b47e) with our extension (mostly lifted from the referenced SO answer). I'll check out what's been done so far.

**Comment by SirArkimedes** [comment]
I can understand that adding a comment to make it clear how to use the emoji type, but where would I put that comment?

**Comment by SD10** [comment]
@SirArkimedes This would be reflected in external docs. Don't worry about merging `v0.9.0` too much. It's changing quickly. We'll rebase when I review this

File: Sources/MessagesCollectionViewFlowLayout.swift [hunk 1]
```diff
@@ -30,6 +30,8 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
     // MARK: - Properties
 
     open var messageLabelFont: UIFont
+    open var emojiLabelFont: UIFont
```
SD10: I think this should be private if we're going to override it by doubling messageLabelFont

Sources/MessagesCollectionViewFlowLayout.swift
```diff
@@ -30,6 +30,8 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
     // MARK: - Properties
 
     open var messageLabelFont: UIFont
```
SD10: This probably needs a `didSet` to reflect the changes in `emojiLabelFont`

Sources/MessagesViewController.swift
```diff
@@ -171,8 +171,8 @@ extension MessagesViewController: UICollectionViewDataSource {
         let message = messagesDataSource.messageForItem(at: indexPath, in: messagesCollectionView)
 
         switch message.data {
-        case .text, .attributedText:
-            let cell = messagesCollectionView.dequeueReusableCell(TextMessageCell.self, for: indexPath)
+        case .text, .attributedText, .emoji:
+            let cell = collectionView.dequeueReusableCell(TextMessageCell.self, for: indexPath)
```
SD10: this should be `messagesCollectionView` but maybe its because this branch is not current with `v0.9.0`

Sources/TextMessageCell.swift
```diff
@@ -70,6 +70,11 @@ open class TextMessageCell: MessageCollectionViewCell<MessageLabel> {
             messageContentView.text = text
         case .attributedText(let text):
             messageContentView.attributedText = text
+        case .emoji(let text):
+            messageContainerView.backgroundColor = .clear
```
SD10: Yeah, I was thinking how to handle the style. Maybe we shouldn't set the color to `.clear` here and expect the user to apply a style of `.none` to the `.emoji` case.

SirArkimedes: I was originally thinking the same thing, by just setting the style to .none here. But the ```MessageCollectionViewCell``` sets the background color from the delegate, which creates just a colored rectangle behind the message.

SD10: Tough one. Can we just do this in the delegate then? You'll have to do an `if case let` or switch on the `MessageType`. If it is emoji return a `.clear` background color. Until we find a better way to handle it.

SirArkimedes: Moved to the delegate.

SD10:
@SirArkimedes This looks good to me! We need to add a CHANGELOG entry under `Added`. This would be a breaking change notifying the users that we added a new `.emoji(String)` case to the `MessageData` type.

Please take care to follow the existing CHANGELOG in v0.9.0. We also have a CHANGELOG_GUIDELINES file in that PR

SirArkimedes:
Done!

SD10 (approved):
@SirArkimedes Thank you for your contribution to MessageKit 💯 🎉 I've invited you to join the MessageKit organization. No pressure if you aren't feeling up to it 😎

PR #229, opened 2017-10-04
Update header/footer dequeueing for new reuseIdentifiers

SD10:
- Deprecates the old header/footer dequeuing methods
- Changes the delegate methods to return non-optionals
- Moves extensions for `UICollectionView` -> `MessagesCollectionView`

### TODO:
- [x] Requires CHANGELOG entry


fbartho (approved):
This looks great.

Foolish on my part that I didn't catch that the final change I made of changing the identifier strings would break things here!

SD10:
No worries 😃 This was needing a refactor. It may clear up a rare bug regarding Swift + Objc interop that @cwalo was facing.

I'm wondering if these methods should be defined as part of `MessagesCollectionView` directly. Then we could possibly ditch the protocol by adding the `reuseIdentifier()` method to `MessageCollectionViewCell`. However, this gets the app running for now.

PR #233, opened 2017-10-04
Fix #221 scroll to bottom is not working in sample app

### LINKED ISSUE #221: Scroll to latest message is not happening after typing the message.
I am trying out sample Chat application inside MessageKit project and found the following behavior. After typing the message, we need to scroll the view to see the new message. I think this is not good for a chat application. The keyboard is hiding the new message. See the below workflow I have tried.
bug

Is it a bug or Is there any configurable property to make it work?
**scrollsToBottomOnKeybordBeginsEditing** is already set to true in the sample application.

tapanprakasht:
The pull request contains a fix for the issue mentioned in #221. The sample app did not call **scrollToBottom**() method after calling sending action. So by default, it is not going to latest message.


SD10 (approved):
Thank you @tapanprakasht 👍

PR #236, opened 2017-10-05
[Example] Fix warning from thread access on reload

SirArkimedes:
Xcode 9 likes to spam console when the UI elements aren’t accessed on the main thread. This fixes one of those warnings.


SD10:
It's not spam :P I was wondering when someone was going to mention this. This may already get fixed in an upcoming PR but if that doesn't get merged I'll merge this one. So we'll leave it open.

SD10:
I decided to go with a local change. Using a completion handler is more convenient I think

PR #240, opened 2017-10-06
Fixing input bar on iPhone X

etoledom:
Making `bottomStackView.bottomAnchor.constraint` equal to `layoutMarginsGuide.bottomAnchor` on  `MessageInputBar.swift`makes the input bar to increase its height on iPhone X to cover the extra bottom space. This breaks the current messagesCollectionView's bottom inset management, so some extra calculations on the bottom inset were added to take into account this extra input bar height.

Also, I noticed that `messagesCollectionView.contentInset.top = additionalTopContentInset + topLayoutGuide.length` was creating an extra top inset on iOS 11, while It was necessary on iOS 10, so a OS version check was added to take into account this difference. I'm not sure If it is the best way to fix this issue.

I tested on iPhone 5S with iOS 10, iPhone 7 with iOS 11 and iPhone X from simulator.
All top and bottom insets where working properly with and without keyboard.

I had no chance to test with a hardware keyboard so it might no work properly with the iPhone X

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -80,10 +80,16 @@ open class MessagesViewController: UIViewController {
             defer { isFirstLayout = false }
 
             addKeyboardObservers()
-            messagesCollectionView.contentInset.top = additionalTopContentInset + topLayoutGuide.length
-            messagesCollectionView.contentInset.bottom = messageInputBar.frame.height
-            messagesCollectionView.scrollIndicatorInsets.bottom = messageInputBar.frame.height
             
+            if #available(iOS 11.0, *) {
+                messagesCollectionView.contentInset.top = additionalTopContentInset
+            } else {
+                messagesCollectionView.contentInset.top = additionalTopContentInset + topLayoutGuide.length
+            }
+            let bottomInset = messageInputBar.frame.height - bottomLayoutGuide.length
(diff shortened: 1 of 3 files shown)


nathantannar4:
Thank you for your work on this. @SD10 and I looked at the different options and determined the best solution shown in #266

etoledom:
Nice! 
I saw you used the same strategy that I used, but testing it on v0.10.0, you didn't solve the extra bottom inset that this gives in the iPhone X. Any plan to solve it?

nathantannar4:
Extra bottom inset? We didn’t see that can you share more on what you mean?

etoledom:
Sorry for the late replay, there's an screenshot :)
I just pull 0.10.0

screen shot 2017-10-22 at 7 26 48 pm

nathantannar4:
Ah interesting. Well that’s more MessagesCollectionView layout issues so the PR should be kept separate. Is there the same extra inset when it’s not an iPhone X?

etoledom:
No, just on iPhone X. Anyways I got to a possible solution to this, I can send it again in a new PR for you to check it out :). But I understand that how this is just an iPhone X problem, it might not be a priority 😄

SD10:
If this is happening on the current `v0.10.0` then it's likely an iPhone X specific problem

nathantannar4:
It might have to do with where the MessagesCollectionView is being anchored

PR #241, opened 2017-10-06
[Fixed] Emoji 2x sizing bug

SD10:
This fixes the mismatching fonts between `configure` and `apply(attributes:)`


SirArkimedes:
I guess this is what you were trying to say originally, huh? Haha.

SirArkimedes (approved):

SD10:
@SirArkimedes Yes, but your idea was still great 😃 I just didn't consider that the methods could not be in sync

Issue #143, opened 2017-09-21
Add copy menu for long press gesture on message bubbles


SD10:
Since we're not using `UITextView` for the message label, the text is not selectable. 
We should add support for copying the text of the `MessageLabel`s.

etoledom:
I'm interested in this one. 
Should I take it from the 0.8.0 branch?
Is there any particular reason we are not using UITextView?
What kind of UI do you have in mind? It would be cool maybe something fancy like Telegram and WhatsApp, but Messages shows a simple action sheet as the copy control 😂

SD10:
Hey @etoledom, I'm working on trying to finish up `v0.8.0` now so this is probably better for `v0.9.0`+
I know @Ganzolo expressed interest in this and may have started working on it.

We're not using `UITextView` because I was advised not to -- which is good enough for me :)

As for the UI, I'm thinking something like the old iMessage/JSQMVC style. Although, the greater flexibility we can provide the better 😃

etoledom:
Cool! Let’s hope @Ganzolo can tell us if he is in this 😅. 
#145 is interesting for me too, I have been working a lot with maps lately.

hamzaozturk:
In JSQMVC, uitextview was being used. The main problem was that, when you use the UIMenuController, there was a conflict between the collection view's UIMenuController and textview's. When you want to allow data detectors, you need to set uitextview selectable. Once you set it as selectable, then it would allow data detectors, then there will be a conflict with the collection view's UIMenuController. Therefore, we couldn't have used data detectors in JSQMVC. 

I think we can use the collection view's UIMenuController to copy the texts. These way we can easily copy both texts and media messages.

SD10:
Awesome @hamzaozturk thanks for this clear explanation and the valuable insight ❤️

Ganzolo:
Yes, I started working on it, I am going to PR something very soon, within next couple of days.

Ganzolo:
Just sharing some progress here. I am using following set of requirements for implementing copy :

- Copy Menu should be centered above message container (a.k.a bubbles)
- Message container should be the only clickable zone to display menu. Meaning clicking on white background of a message cell should not trigger this menu
- Color of message should be slightly highlighted to reflect selection
- Whole MessageCollectionView's scrollview should stop allowing scrolling when having menu displayed. In my tests, this prevents Copy message to be misaligned when scrolling (and also this is default behavior of most other tasks, Messages and Viber for instance).

[...]

PR #248, opened 2017-10-09
Adding copy/paste support from MessagesViewController

Ganzolo:
(Thanks for sending a pull request! Please make sure you click the link above to view the contribution guidelines, then fill out the blanks below.)

Copy implementation

#143 

Any other comments?
-------------------
Better implementation

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -72,7 +72,37 @@ open class MessagesViewController: UIViewController {
         registerReusableViews()
         setupDelegates()
 
+		self.addMenuControllerObservers()
     }
+	
+	fileprivate func addMenuControllerObservers() {
+		NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerWillShow(aNotification:)), name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+		NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerDidHide(aNotification:)), name: NSNotification.Name.UIMenuControllerDidHideMenu, object: nil)
+	}
+	
+	fileprivate func removeMenuControllerObservers() {
+		NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+		NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerDidHideMenu, object: nil)
+	}
+	
+	fileprivate var targetedCell:UICollectionViewCell?
+	fileprivate var targetedFrame = CGRect.zero
+	@objc func menuControllerWillShow(aNotification:Notification) {
+		if let currentMenuController = aNotification.object as? UIMenuController,
+			let currentTargetedCell = targetedCell {
+			NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+			currentMenuController.setMenuVisible(false, animated: false)
+			currentMenuController.setTargetRect(self.targetedFrame, in: currentTargetedCell)
+			currentMenuController.setMenuVisible(true, animated: true)
+			self.messagesCollectionView.isUserInteractionEnabled = false
+		}
+	}
+	
+	@objc func menuControllerDidHide(aNotification:Notification) {
+		NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerWillShow(aNotification:)), name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+		self.messagesCollectionView.isUserInteractionEnabled = true
+		self.targetedCell = nil
+	}
 
     open override func viewDidLayoutSubviews() {
         // Hack to prevent animation of the contentInset after viewDidAppear
@@ -95,6 +125,7 @@ open class MessagesViewController: UIViewController {
 
     deinit {
         removeKeyboardObservers()
+		removeMenuControllerObservers()
     }
 
     // MARK: - Methods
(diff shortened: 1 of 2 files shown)


Sources/Controllers/MessagesViewController.swift
```diff
@@ -72,7 +72,37 @@ open class MessagesViewController: UIViewController {
         registerReusableViews()
         setupDelegates()
 
+		self.addMenuControllerObservers()
```
SD10: Can we remove `self` here?

Sources/Views/MessageLabel.swift
```diff
@@ -406,10 +406,6 @@ open class MessageLabel: UILabel, UIGestureRecognizerDelegate {
         addGestureRecognizer(tapGesture)
         tapGesture.delegate = self
 
-        let longPressGesture = UILongPressGestureRecognizer(target: self, action: #selector(handleGesture(_:)))
```
SD10, line 409: I still have to look into this

Sources/Controllers/MessagesViewController.swift
```diff
@@ -132,6 +163,68 @@ open class MessagesViewController: UIViewController {
     }
 }
 
+// MARK: - UICollectionViewDelegate
+
+extension MessagesViewController: UICollectionViewDelegate {
+	
+	public func collectionView(_ collectionView: UICollectionView, shouldShowMenuForItemAt indexPath: IndexPath) -> Bool {
```
SD10: This needs to be marked as `open`

Sources/Controllers/MessagesViewController.swift
```diff
@@ -132,6 +163,68 @@ open class MessagesViewController: UIViewController {
...
+				return true
+			}
+		default:
+			break
+		}
+		
+		return false
+	}
+	
+	public func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath, withSender sender: Any?) -> Bool {
```
SD10: This needs to be marked as `open`

Sources/Controllers/MessagesViewController.swift
```diff
@@ -132,6 +163,68 @@ open class MessagesViewController: UIViewController {
...
+		return (action == NSSelectorFromString("copy:"))
+	}
+	
+	public func collectionView(_ collectionView: UICollectionView, performAction action: Selector, forItemAt indexPath: IndexPath, withSender sender: Any?) {
```
SD10: This needs to be marked as `open`

Sources/Controllers/MessagesViewController.swift
```diff
@@ -72,7 +72,37 @@ open class MessagesViewController: UIViewController {
...
+		NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerDidHide(aNotification:)), name: NSNotification.Name.UIMenuControllerDidHideMenu, object: nil)
+	}
+	
+	fileprivate func removeMenuControllerObservers() {
+		NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+		NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerDidHideMenu, object: nil)
+	}
+	
+	fileprivate var targetedCell:UICollectionViewCell?
+	fileprivate var targetedFrame = CGRect.zero
```
SD10: Can we keep properties at the top of the file like we're doing with the rest of them?
Can we also put all of the logic related to the `MenuController` in an extension under Keyboard Handling?

SD10 (changes requested):
Thanks @Ganzolo 👍 I just have to review the `Long Press Gesture` and any other gesture conflicts. I left you some comments regarding how I think we should organize the code

Ganzolo:
@SD10 Thanks for feedbacks. I did all suggested fixes. I have nothing to say about what u suggested, except for putting all properties at the top. I usually like to put them close to methods that are using it so refactoring is easier, scope is more concise. 

However, then we cannot use extensions which is also a good way to break the code. So both solution is fine with me. The best would be to be able to add properties in an extension :)

SD10:
@Ganzolo Can you add a CHANGELOG entry for this?

nathantannar4:
I like this!

SD10:
Me too, I just have to make sure the gestures are behaving properly. If either of you can pull it and run it that would be great. I will try myself again soon.

I think there was some inconsistency with the gestures last time I tested it. Sometimes the copy menu would be displayed for a long press. Sometimes for a long press and lift up. Things like that.

nathantannar4 (approved):
Looks like a good feature we need

SD10:
@nathantannar4 Were you able to test the gestures?Yes we definitely need it but it needs to be working properly.

nathantannar4:
@SD10 @Ganzolo I notice that the "Copy" prompt appears even when a long press is recognized on the AvatarView or anywhere else outside of the MessageLabel. Is this behavior we want?

Only other thing I see is if you long press a link such as an address, the "Copy" prompt appears before the notifier is sent that the user has tapped an address. I wonder if this can be prevented?

SD10:
@nathantannar4 Thanks for the extra set of eyes. I definitely missed some of those. These are all valid points. Looks like we're going to have to look into this a little bit more.

It's a pretty important feature to have but easy enough for clients to implement on their own while we figure this out.

Ganzolo:
@SD10 @nathantannar4 
I looked today into this issue and I am afraid that we would need to use custom touch on a Cell level to prevent this behavior from happening.

I tried to make contentView transparent background or to assign a long press gesture to background of cell. But it does not work.

The other idea would be to respond false when this method is called only when the touch is not in the bubble area. However I didn't find a way to get the position of the touch from this method.

`func collectionView(_ collectionView: UICollectionView, shouldShowMenuForItemAt indexPath: IndexPath) -> Bool`

cwalo:
Sorry, just now getting to this. For your consideration - add a long press gesture recognizer to `MessageLabel` and either pass the gesture type (maybe via a custom enum) along in the `MessageLabelDelegate` methods and/or have an additional `didLongPressMessageLabel` delegate method with a default implementation that displays the MenuController. @Ganzolo @SD10 @nathantannar4

SD10:
@cwalo Good suggestions. We may have to move this to the cell level as @Ganzolo mentioned. I'm going to be rewriting `MessageLabel` and then we can re-evaluate from there

relaxsus:
Hello! I can't call the menu with long press (copy, custom action...). when will this function be in MessageKit? Or help me, please, to write the right code. It was not the problem to call this menu on JSQ.

SD10:
Hey @relaxsus,

You can remove the long press gesture from your MessageKit copy for now. Right now we require people to add copy/paste support themselves.

I agree this feature is something we should provide, but due to its ease of implementation outside of MessageKit, I’ve spent my time trying to resolve more serious issues. It’s on the roadmap though


nathantannar4:
#422

PR #250, opened 2017-10-10
Fixes #243 by deprecating addiontalTopContentInset

### LINKED ISSUE #243: Keyboard will cover 4th and 5th messages in the newly initiated chat
## General Information

* MessageKit Version: **0.9.0**

* iOS Version(s): **Any**

* Swift Version: **Any**

* Devices/Simulators: **Any**

* Reproducible in ChatExample? (Yes/No): **Yes**

##  What happened?
When you initiate the new chat and start sending messages, depending on the screen size, 4th and 5th messages will be covered by the keyboard until you send the new message and then, the chat view will 'jump up' so every new message will be displayed properly.
You can refer to my screenshot to see what's the issue in this case.

issue

To have every single message visible without keyboard covering it.

SD10:
### TODO: 
- [x] CHANGELOG entry

This should fix #243 where the `contentInset.top` is set incorrectly for the `MessagesCollectionView`.
It looks like just setting `extendedTopLayoutIncludesOpaqueBars` is all that is necessary. By merging the two PRs in the wrong order I missed this.

cc @cwalo, hope I'm not going crazy 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Removed
+
+- **Breaking Change** Removed `additionalTopContentInset` property of `MessagesViewController` because this is no longer necessary
+when `extendedLayoutIncludesOpaqueBars` is `true`.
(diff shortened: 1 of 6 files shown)


cwalo:
Hmm. Might this explain why I had to set `extendedLayoutIncludesOpaqueBars = false` for my opaque navigation bar?

Issue #96, opened 2017-09-02
Unit Testing


SD10:
### Increase Unit Test Coverage
Increasing test coverage is always an easy way to get involved in the project. I plan on focusing on this in upcoming releases but always open to help 🤗

LeLuckyVint:
Taking this one, @SD10

SD10:
That would be awesome @LeLuckyVint ❤️

PR #253, opened 2017-10-10
Increase Unit Test coverage

pryaniq:
Increase Unit Test coverage.

WIP on #96


SD10:
@pryaniq This is awesome ❤️! I'll be able to review it in more depth later

SD10:
@pryaniq Just wanted to follow up. I was able to get part way through reviewing this tonight. It's looking good 👍 Hope to get this merged in this week

SD10:
@nathantannar4 We're also going to have to try and get this merged in before it becomes stale. Some of these tests involve the `MessageInputBar`.

nathantannar4:
I'm no expert in XCTest but we def need this. I checked the tests for MessageInputBar and after my updates there will be some tests that fail. Should we really be testing things like default properties?

SD10:
@nathantannar4 Yeah I definitely think it's reasonable to test defaults. We can catch things like "Aa" vs "New Message" 🤣 Among other things~

SD10:
@pryaniq & @LeLuckyVint Thanks for taking the time to get the unit tests started 💯 I've invited you both to the organization 👍 No pressure if you're not feeling up to it 😎

SD10 (approved):

pryaniq:
@SD10 Thank you for invitation, cheers! 🍺 
In the future I will continue working on unit tests :)

Issue #257, opened 2017-10-14
ChatExample's messageList didSet causes code to run on non-ui thread


candiun:
## General Information

* MessageKit Version: 0.9.0

* iOS Version(s): All

* Swift Version: All

* Devices/Simulators: All

* Reproducible in ChatExample? (Yes/No): Yes

##  What happened?

The main thread sanitizer detected that code did not run on the main UI thread.

## What did you expect to happen?

The code should have ran on the main UI thread

SD10:
Hey @evilmint thanks for reporting this 👍 it has already been fixed in the upcoming version

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #258, opened 2017-10-14
Fixes ui call on non-main event

candiun:
This PR fixes code call on non-main UI thread #257


SD10 (approved):
Thank you @evilmint 💯

Issue #260, opened 2017-10-16
Placeholder Alignment is always set to Left.


hemangshah:
## General Information

* MessageKit Version: 0.8.1 **and** 0.9.0

* iOS Version(s): 11.0

* Swift Version: 3.0

* Devices/Simulators: iPhone 7+

* Reproducible in ChatExample? (Yes/No): Yes

##  What happened?

In the ChatExample, setting this line after setting the delegate of `messageInputBar` doesn't change the placeholder alignment to `right`. It's always showing the placeholder alignment from `left`.

`messageInputBar.inputTextView.placeholderLabel.textAlignment = .right`
`messageInputBar.inputTextView.textAlignment = .right`

**Both don't work.**

## What did you expect to happen?

I am expecting the placeholder to come from right side.

nathantannar4:
Fixed with PR #262

PR #262, opened 2017-10-16
Bug Fix for issue #260

nathantannar4:
It appears that inside a UITextVIew anchoring a label to the left and right anchors don't extend the labels width all the way, thus adjusting the textAllignment doesn't have visible changes. (Thanks @hemangshah)

This fix changes the constraints of the label to use a width anchor instead of a right anchor. Also removed the UITextViewDelegate as I saw it was causing other MessageInputBar height issues and was previously only included as a temp solution to another bug that was corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


SD10:
@nathantannar4 Can you add a CHANGELOG entry for this?

SD10 (approved):

PR #263, opened 2017-10-17
Improve MessagesCollectionViewLayout performance

SD10:
This pull request should cut down reload data / layout calculation time by ~75%

**Introduces**:
- `MessageIntermediateLayoutAttributes` to eliminate redundant calculations
- A cache using the `messageId` property of `MessageType`
- A `func clearCachedAttributes(for message: MessageType)` method to invalidate the cache

**There are some problems with this**:

MessageKit is a bit too dynamic and that makes caching layout information difficult. This approach assumes that once a `MessageType` is calculated, those calculations will remain constant. However, the MessageKit API allows the user to write code that does not follow this expectation, thus resulting in undefined behavior.

For example, consider the following:
```Swift
func avatarSize(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGSize {
    if indexPath.section % 2 == 0 {
        return CGSize(width: 30, height: 30)
    } else {
       return .zero
   }
}
```
This allows MessageKit users to make the size of an avatar dynamic over an `IndexPath`. Even index paths will have a size of `30 x 30` and odd index paths will be `0 x 0`. 

If the user removes a cell from the `MessagesCollectionView` this will result in the cached attributes becoming invalid, yet they will continue to be used. This is undefined behavior.

**How to solve this?**
I'm not sure really, ideally I'd like to assume that everything is static and make the user responsible for invalidating the cache otherwise.

The `clearCachedAttributes` method I provided doesn't really help in this scenario because there's no good time for the user to call this. Ideally, we need an opportunity to invalidate the cache before any `sizeForItem(at indexPath: IndexPath)` call in `MessagesCollectionViewFlowLayout`. Another delegate method? Possibly. However, this could lack the context where the user is making the layout dynamic in the first place.

### TODO:
- [x] CHANGELOG
~- [ ] Improving cell contents origin caching~
- [x] Cache invalidation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Added
+
+- Added `removedCachedAttributes(for:MessageType)`, `removeAllCachedAttributes()`, and `attributesCacheMaxSize` to
+`MessagesCollectionViewFlowLayout` to manage the caching of layout information for messages.
+[#263](https://github.com/MessageKit/MessageKit/pull/263) by [@SD10](https://github.com/sd10).
+
+### Changed
+
+- Layout information is now being cached by `MessagesCollectionViewFlowLayout` for each `MessageType` using the
+`messageId` property. (This means if your layout is dynamic over the `IndexPath` you need to handle cache invalidation).
+[#263](https://github.com/MessageKit/MessageKit/pull/263) by [@SD10](https://github.com/sd10).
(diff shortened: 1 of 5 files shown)


nathantannar4 (approved):
Once the caching is done I can submit a PR for my bouncy layout animations 😊

SD10:
@nathantannar4 Probably can't have support for springy bubbles any-time soon. It was an experimental feature in JSQMessagesViewController. More of a MessageKit `2.0` feature in my eyes 🤔

PR #266, opened 2017-10-18
iPhone X Support

nathantannar4:
Changes the `MessageInputBar` bottom `UIStackView`'s bottomAnchor to `layoutMarginsGuide.bottomAnchor`. This will prevent the home indicator on the iPhone X from overlapping on the `InputTextView`


SD10 (approved):
Cool 🎉 Feel free to merge if it passes CI

PR #267, opened 2017-10-18
Changing MessagesViewController reference to MessageInputBar

nathantannar4:
In the `MessagesViewController` we set the initial contentInset.bottom of the MessagesCollectionView to a direct reference to `messageInputBar`. This can pose problems if someone ever overrides the `inputBarAccessoryView` with a custom input bar that is not a `MessageInputBar`. 

This PR changes a direct reference to `messageInputBar` to just `inputAccessoryView`


SD10 (approved):
LGTM 👍

PR #273, opened 2017-10-21
New Classes: SeparatorLine & InputStackView

nathantannar4:
Easier reuse by making these subclasses and then setting up their initial properties to what we need them to be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Added
+
+-  Created `SeparatorLine` and `InputStackView` as their own subclass of `UIView` and `UIStackView` respectively. This just improves reusability
+[#273](https://github.com/MessageKit/MessageKit/pull/273) by [@nathantannar4](https://github.com/nathantannar4).
+
+### Changed
+
+-  **Breaking Change**  The properties `leftStackView`, `rightStackView` and `bottomStackView` in `MessageInputBar` are now of type `InputStackView`. The property `separatorLine` is also now of type `SeparatorLine` in `MessageInputBar`
+[#273](https://github.com/MessageKit/MessageKit/pull/273) by [@nathantannar4](https://github.com/nathantannar4).
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Changed
+
+-  Created `SeparatorLine` and `InputStackView` as their own subclass of `UIView` and `UIStackView` respectively. This just improves reusability.
```
SD10: This should go under `Added`

Sources/Views/MessageInputBar.swift
```diff
@@ -52,43 +52,33 @@ open class MessageInputBar: UIView {
...
-        let view = UIStackView()
-        view.axis = .horizontal
-        view.translatesAutoresizingMaskIntoConstraints = false
-        view.distribution = .fill
-        view.alignment = .fill
-        view.spacing = 15
-        return view
-    }()
-    
+    open let separatorLine = SeparatorLine()
```
SD10, line 55: This is a breaking change because `seperatorLine` is no longer typed as `UIView`

Sources/Views/MessageInputBar.swift
```diff
@@ -52,43 +52,33 @@ open class MessageInputBar: UIView {
...
+    
+    /**
+     The InputStackView at the InputStackView.left position
+     
+     ## Important Notes ##
+     1. It's axis is initially set to .horizontal
+     */
+    open let leftStackView = InputStackView(axis: .horizontal, spacing: 0)
```
SD10, line 63: Same here, now all the stackViews are typed as `InputStackView`

SD10 (changes requested):
Ok @nathantannar4 this looks good to me, just 2 things:
1. We need to move that entry to `Added` stating that we added 2 new classes.
2. We need to add an entry under changed saying that we changed the types of the `separatorLine` property and the 3 stackview properties

SD10 (approved):
Looks good to me if it passes CI 👍

PR #274, opened 2017-10-21
MessageInputBar Items to a protocol InputItem 

nathantannar4:
New protocols: `InputItem` 

`MessageInputBar` can now be given items that conform to `InputItem` rather than just `InputBarButtonItem`. This means you can create your own items that don't have to be a UIButton.


nathantannar4:
I am going to remove `InputManager` for now and keep this PR focused on the switch to the new protocol `InputItem`

PR #275, opened 2017-10-21
Documentation Updates

nathantannar4:
Docs for InputBarButtonItem, MessageInputBarDelegate and InputTextView

### DIFF
--- a/Sources/Views/InputBarItem.swift
+++ b/Sources/Views/InputBarItem.swift
@@ -24,8 +24,19 @@
 
 import UIKit
 
+/**
+ A InputItem that inherits from UIButton
+ 
+ ## Important Notes ##
+ 1. Intended to be used in an `InputStackView`
+ */
 open class InputBarButtonItem: UIButton {
     
+    /// The spacing properties of the InputBarButtonItem
+    ///
+    /// - fixed: The spacing is fixed
+    /// - flexible: The spacing is flexible
+    /// - none: There is no spacing
     public enum Spacing {
         case fixed(CGFloat)
         case flexible
@@ -36,24 +47,25 @@ open class InputBarButtonItem: UIButton {
     
     // MARK: - Properties
     
-    /// Set automatically when using the InputBarAccessoryView method 'setStackViewItems'
+    /// A weak reference to the MessageInputBar that the InputBarButtonItem used in
     open weak var messageInputBar: MessageInputBar?
     
-    public var spacing: Spacing = .none {
+    /// The spacing property of the InputBarButtonItem
+    open var spacing: Spacing = .none {
         didSet {
             switch spacing {
             case .flexible:
-              setContentHuggingPriority(UILayoutPriority(rawValue: 1), for: .horizontal)
+                setContentHuggingPriority(UILayoutPriority(rawValue: 1), for: .horizontal)
             case .fixed:
-              setContentHuggingPriority(UILayoutPriority(rawValue: 1000), for: .horizontal)
+                setContentHuggingPriority(UILayoutPriority(rawValue: 1000), for: .horizontal)
             case .none:
-              setContentHuggingPriority(UILayoutPriority(rawValue: 500), for: .horizontal)
+                setContentHuggingPriority(UILayoutPriority(rawValue: 500), for: .horizontal)
             }
         }
     }
     
     /// When not nil this size overrides the intrinsicContentSize
-    private(set) var size: CGSize? = CGSize(width: 20, height: 20) {
+    private var size: CGSize? = CGSize(width: 20, height: 20) {
         didSet {
             invalidateIntrinsicContentSize()
         }
@@ -70,7 +82,8 @@ open class InputBarButtonItem: UIButton {
         return contentSize
     }
     
-    internal var parentStackViewPosition: MessageInputBar.UIStackViewPosition?
+    /// A reference to the stack view position that the InputBarButtonItem is held in
+    open var parentStackViewPosition: InputStackView.Position?
     
     /// The title for the UIControlState.normal
     open var title: String? {
@@ -92,6 +105,7 @@ open class InputBarButtonItem: UIButton {
         }
     }
     
+    /// Calls the onSelectedAction or onDeselectedAction when set
     open override var isHighlighted: Bool {
         didSet {
             if isHighlighted {
@@ -102,6 +116,7 @@ open class InputBarButtonItem: UIButton {
         }
     }
     
+    /// Calls the onEnabledAction or onDisabledAction when set
     open override var isEnabled: Bool {
         didSet {
             if isEnabled {
@@ -112,11 +127,12 @@ open class InputBarButtonItem: UIButton {
         }
     }
     
-    // MARK: - Hooks
+    // MARK: - Reactive Hooks
     
     private var onTouchUpInsideAction: InputBarButtonItemAction?
     private var onKeyboardEditingBeginsAction: InputBarButtonItemAction?
     private var onKeyboardEditingEndsAction: InputBarButtonItemAction?
+    private var onKeyboardSwipeGestureAction: ((InputBarButtonItem, UISwipeGestureRecognizer) -> Void)?
     private var onTextViewDidChangeAction: ((InputBarButtonItem, InputTextView) -> Void)?
     private var onSelectedAction: InputBarButtonItemAction?
     private var onDeselectedAction: InputBarButtonItemAction?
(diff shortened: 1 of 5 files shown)


Sources/Views/InputBarItem.swift
```diff
@@ -36,24 +47,25 @@ open class InputBarButtonItem: UIButton {
     
     // MARK: - Properties
     
-    /// Set automatically when using the InputBarAccessoryView method 'setStackViewItems'
+    /// A weak reference to the MessageInputBar that the InputBarButtonItem used in
     open weak var messageInputBar: MessageInputBar?
     
-    public var spacing: Spacing = .none {
+    /// The spacing property of the InputBarButtonItem
+    open var spacing: Spacing = .none {
```
SD10: Think we can make this more descriptive?

Sources/Views/InputBarItem.swift
```diff
@@ -166,90 +191,152 @@ open class InputBarButtonItem: UIButton {
...
     // MARK: - Hook Setup Methods
     
+    /// Used to setup your own initial properties
+    ///
+    /// - Parameter item: A reference to Self
+    /// - Returns: Self
     @discardableResult
-    open func configure(_ setup: InputBarButtonItemAction) -> Self {
-        setup(self)
+    open func configure(_ item: InputBarButtonItemAction) -> Self {
```
SD10, line 199: This is an API breaking change 😅

Sources/Views/InputTextView.swift
```diff
@@ -24,10 +24,32 @@
...
+    
+    open override var text: String! {
+        didSet {
+            textViewTextDidChange()
+        }
+    }
+    
+    open override var attributedText: NSAttributedString! {
+        didSet {
+            textViewTextDidChange()
```
SD10: Very interesting. I'm sorta against manually calling anything that is normally triggered by the notification center, but let's go for it for now. Reason being is because this kind of results in unexpected behavior

Sources/Views/MessageInputBar.swift
```diff
@@ -355,7 +357,7 @@ open class MessageInputBar: UIView {
     ///   - items: New UIStackView arranged views
     ///   - position: The targeted UIStackView
     ///   - animated: If the layout should be animated
-    open func setStackViewItems(_ items: [InputBarButtonItem], forStack position: UIStackViewPosition, animated: Bool) {
+    open func setStackViewItems(_ items: [InputBarButtonItem], forStack position: InputStackView.Position, animated: Bool) {
```
SD10, line 358: We're changing the type here

Sources/Views/MessageInputBar.swift
```diff
@@ -306,7 +306,7 @@ open class MessageInputBar: UIView {
     /// Layout the given UIStackView's
     ///
     /// - Parameter positions: The UIStackView's to layout
-    public func layoutStackViews(_ positions: [UIStackViewPosition] = [.left, .right, .bottom]) {
+    public func layoutStackViews(_ positions: [InputStackView.Position] = [.left, .right, .bottom, .top]) {
```
SD10: I don't think we have a `.top` stackView yet? We should keep new features confined to a single PR. No partial implementations. Also, we change the type here

Sources/Views/InputTextView.swift
```diff
@@ -118,31 +155,46 @@ open class InputTextView: UITextView {
...
+                                               name: Notification.Name.UITextViewTextDidChange,
+                                               object: nil)
+    }
+    
+    // MARK: - Notifications
+    
+    /// Updates the placeholderLabel's isHidden property based on the text being empty or not
+    @objc
+    open func textViewTextDidChange() {
+        placeholderLabel.isHidden = !text.isEmpty
```
SD10: Actually we should probably move this back into a `didSet`? Too easy to override the behavior?

Sources/Views/InputTextView.swift
```diff
@@ -118,31 +155,46 @@ open class InputTextView: UITextView {
...
-        addConstraints()
+        addObservers()
+        addPlaceholderLabel()
     }
-
-    private func addSubviews() {
-
+    
+    /// Adds the placeholderLabel to the view and sets up its initial constraints
+    private func addPlaceholderLabel() {
```
SD10: Probably better named `setupPlaceholderLabel` now because it does more than just add?

Sources/Views/InputTextView.swift
```diff
@@ -38,47 +60,51 @@ open class InputTextView: UITextView {
...
     open override var textAlignment: NSTextAlignment {
         didSet {
             placeholderLabel.textAlignment = textAlignment
         }
     }
-
-    open var placeholderLabelInsets: UIEdgeInsets = UIEdgeInsets(top: 4, left: 7, bottom: 4, right: 7) {
+    
+    /// The textContainerInset of the InputTextView. When set the placeholderLabelInsets is also updated
+    open override var textContainerInset: UIEdgeInsets {
```
SD10: This one kind of confuses me. Why even have two properties if we're going to keep them in sync? `textContainerInset` will always override what you've done in `placeholderLabelInsets`. We should keep these separate?

Sources/Views/InputTextView.swift
```diff
@@ -24,10 +24,32 @@
 
 import UIKit
 
+/**
+ A UITextView that has a UILabel embedded for placeholder text
+ 
+ ## Important Notes ##
+ 1. Changing the font, textAlignment or textContainerInset automatically performs the same modifications to the placeholderLabel
+ 2. Intended to be used in an `MessageInputBar`
+ 3. Default placeholder text is "New Message"
```
SD10, line 33: I think anything that has a default value we should mention it directly in that part of the documentation. This will be easier to maintain and way more obvious

Sources/Views/MessageInputBar.swift
```diff
@@ -319,6 +319,8 @@ open class MessageInputBar: UIView {
             case .bottom:
                 bottomStackView.setNeedsLayout()
                 bottomStackView.layoutIfNeeded()
+            case .top:
```
SD10: More `.top` references

Sources/Views/MessageInputBar.swift
```diff
@@ -386,6 +388,8 @@ open class MessageInputBar: UIView {
                     bottomStackView.addArrangedSubview($0)
                 }
                 bottomStackView.layoutIfNeeded()
+            case .top:
```
SD10: Ditto, the pokemon

Sources/Views/InputTextView.swift
```diff
@@ -38,47 +60,51 @@ open class InputTextView: UITextView {
...
+    
+    /// The textContainerInset of the InputTextView. When set the placeholderLabelInsets is also updated
+    open override var textContainerInset: UIEdgeInsets {
         didSet {
-            updateConstraintsForPlaceholderLabel()
+            placeholderLabelInsets = textContainerInset
         }
     }
     
     open override var scrollIndicatorInsets: UIEdgeInsets {
```
SD10: Just curious if we lose the Apple documentation when we override

Sources/Views/InputBarItem.swift
```diff
@@ -112,11 +127,12 @@ open class InputBarButtonItem: UIButton {
         }
     }
     
-    // MARK: - Hooks
+    // MARK: - Reactive Hooks
     
     private var onTouchUpInsideAction: InputBarButtonItemAction?
     private var onKeyboardEditingBeginsAction: InputBarButtonItemAction?
     private var onKeyboardEditingEndsAction: InputBarButtonItemAction?
+    private var onKeyboardSwipeGestureAction: ((InputBarButtonItem, UISwipeGestureRecognizer) -> Void)?
```
SD10: Let's just make sure this is all good. If I remember correctly it was causing some weird behavior a few versions back. If we keep it, add it to the changelog as an `Added` entry 😃

SD10 (changes requested):
@nathantannar4 Take a look at my feedback and let me know what you think. It will be nice to finish the first area of documentation 😅

Sources/Views/InputTextView.swift
nathantannar4: If someone manually sets the attributedText property to empty, the placeholder label is still hidden without this

Sources/Views/InputBarItem.swift
nathantannar4: Oh no that shouldn't be there. was waiting till everything else is done to add swipe gestures again

Sources/Views/InputTextView.swift
nathantannar4: Looks like apples doc stays on the autocomplete in Xcode 😊

Sources/Views/InputTextView.swift
nathantannar4: Will add it there as well!

Sources/Views/MessageInputBar.swift
nathantannar4: Got rid of all the .top refs for now

Sources/Views/InputBarItem.swift
nathantannar4: Is it? Because of the _ no one should have already enforced the label to be 'setup'

Sources/Views/InputBarItem.swift
nathantannar4: /// The spacing property of the InputBarButtonItem that determines the contentHuggingPriority and any
    /// additional space to the intrinsicContentSize

Sources/Views/InputBarItem.swift
SD10: You're right! I overlooked the parameter name omission

Sources/Views/InputTextView.swift
SD10: Sorry I'm not sure I follow, I still think the `textViewTextDidChange()` body would be better as a `didSet` here 🤔

Sources/Views/InputBarItem.swift
nathantannar4: Cool, i just figured setup was a poor descriptor

Sources/Views/InputTextView.swift
nathantannar4: sure

nathantannar4:
Thanks @SD10 suggestions made

SD10 (approved):
Thanks @nathantannar4 great work as usual 💯

Issue #232, opened 2017-10-04
Stop calculating origins for UI elements with a .zero frame


SD10:
No clue why I'm doing origin calculations on cell contents when they have a frame of .zero

SD10:
This was resolved in #263

PR #278, opened 2017-10-22
fix issue #232 to stop calculating origins for UI elements with a .zero frame

maxxx777:
This PR stops calculating origins for UI elements (avatar, message container, cell bottom label, cell top label) with a .zero content frame by returning .zero origins if content size is .zero.

Yes, issue #232

No

Any other comments?
-------------------
This is my first PR into MessageKit. Hopefully, I got the issue properly.

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 8 Plus SImulator

**iOS Version:** iOS 11.0

**Swift Version:** 4.0

**MessageKit Version:** 0.9.0


SD10 (commented):
@maxxx777 Thanks for taking the time to submit this PR, however, I already resolved #232. It was regarding the frames of the cell's contents. I'm having trouble thinking of a situation where the `contentFrame` would ever be `.zero`, so I don't want to perform the additional check.

PR #279, opened 2017-10-23
Misc/protocol addition

nathantannar4:
Added `InputItem`, a protocol that `InputBarButtonItem` now conforms to. `MessageInputBar` has been switch to allow items that conform to `InputItem` so you no longer are restricted to just `InputBarButtonItem`

The property enum used to define the stack view position, `MessageInputBar.UIStackViewPosition` was changed to the enum `InputStackView.Position`


nathantannar4:
@SD10 We should include this is v0.10.0

SD10:
It has to go in `v0.11.0`, I haven't even had a chance to run the code 😞. I'm just making this release now so that we can change our workflow. We can't have this mix of bugfixes/API breaking changes in a single branch. It makes it too difficult to ship minor patches

nathantannar4:
I think this is outdated now. We can re-evaluate this later

PR #283, opened 2017-10-24
Provide an interface to access the avatarView in MessagesDataSource

azurechen:

```
func avatarView(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView, convertAvatarView: AvatarView) -> AvatarView
```

Provide an interface for users to access the avatarView.
It will be useful when users want to asynchronously load images by `SDWebImage`
Or change some attributes of avatarView like background color.

The following is a sample code.

```
func avatarView(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView, convertAvatarView: AvatarView) -> AvatarView {
    let avatarView = convertAvatarView
    avatarView.backgroundColor = .bgGray
    avatarView.imageView.sd_setImage(with: "http://xxx.xxx.xxx/sample.png")
    
    return avatarView
}
```

And `convertAvatarView` allow user not to new an avatarView again.

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 6, iPhone 7s Simulator

**iOS Version:** 10

**Swift Version:** 3.2

**MessageKit Version:** 0.9.0


nathantannar4:
Hi @azurechen thanks for wanting to contribute! Unfortunately I don’t think your method will work. We reuse the collection view cells to increase performance. If the chat was between one user and another you wouldn’t be able to notice the error. However, once you introduce many users an image will begin to load async with an image view reference, then the cell is reused for a new message. This creates what’s known as a race condition and the image that appears depends on which was loaded first and then keeps changing based on the async calls. We are not sure if we can even support async messages as caching the images loaded should also be involved. As of now we are leaving this to the users to load the images asynchronously

azurechen:
Hi @nathantannar4, thanks for your opinion.

This function now works fine in my project. (A chat with more than 3 members) using `SDWebImage`

I very agree your concern. But I think the asynchronous function is not the point of this Pull Requests.
My opinion is `MessageKit` should provide a flexible usage for developer to implement their idea including asynchronous functions but not only that. User may just want to change the background color.
`SDWebImage` consider this possibility of race condition. But developer also needs to understand using a asynchronous function may cause race condition issues. Even `imageView` in original `UITableViewCell` will meet this problem.

In addition, while reuse a cell you sure set a default value for `avatarView.imageView.image` first and waiting for the async image back.
That is why I put `avatarView = dataSource.avatarView(for:at:in:convertAvatarView)` after `avatarView.set(avatar: avatar)` in `Sources/Views/Cells/MessageCollectionViewCell.swift`

```
  avatarView.set(avatar: avatar)
+ avatarView = dataSource.avatarView(for: message, at: indexPath, in: messagesCollectionView, convertAvatarView: avatarView)
```

`avatarView.set(avatar: avatar)` will provide a default value for reused cell. And each time call `avatarView.imageView.sd_setImage(with: )` will reset the imageURL of the reused imageView. URL information was store in this imageView. When the imageView receive the image back. It can check the URL to make sure to accept the right image.

SD10:
Hey @azurechen, after reviewing this pull request I don't think this is the right approach for MessageKit. Support for async loadings is on the roadmap and the team has a different set of ideas in mind. We still really appreciate the time you took to put this together. Feel free to get in contact on the Slack channel 👍

Issue #186, opened 2017-09-28
Maintain scroll position after loading new messages


SD10:
This has been requested a few times already and I think people will only continue to request this capability.

Great info here:
 [#1838 maintaining scroll position when loading older messages](https://github.com/jessesquires/JSQMessagesViewController/issues/1838)
[#1870](https://github.com/jessesquires/JSQMessagesViewController/issues/1870)

SD10:
This was completed in #284 by adding a `reloadDataAndKeepPosition()` method to `MessagesCollectionView`. It will be available in `v0.12.0`.

PR #284, opened 2017-10-25
Enhancement/load more messages and maintain scroll position

azurechen:
Pull down to load more messages and keep the position.
The example was updated. Reviewers can try the effect at `ConversationViewController`

#186 Maintain scroll position after loading new messages

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 6, iPhone 7s Simulators

**iOS Version:** 10

**Swift Version:** 3.2

**MessageKit Version:** 0.9.0


Sources/Views/MessagesCollectionView.swift
```diff
@@ -67,5 +70,31 @@ open class MessagesCollectionView: UICollectionView {
...
+            
+            // calculate the offset and reloadData
+            let beforeContentSize = self.contentSize
+            super.reloadData()
+            self.layoutIfNeeded()
+            let afterContentSize = self.contentSize
+            
+            // reset the contentOffset after data is updated
+            self.setContentOffset(CGPoint(
+                x: self.contentOffset.x + (afterContentSize.width - beforeContentSize.width),
```
cwalo: For readability, consider creating a let variable for the new offset and then use that in `setContentOffset`.

Sources/Views/MessagesCollectionView.swift
```diff
@@ -67,5 +70,31 @@ open class MessagesCollectionView: UICollectionView {
         guard let indexPath = indexPathForLastItem else { return }
         scrollToItem(at: indexPath, at: .bottom, animated: animated)
     }
+    
+    override open func reloadData() {
```
SD10: @azurechen Thanks for submitting this PR! Unfortunately, I don't think overriding `reloadData` is the right approach to achieving this functionality. 

I notice this is the second PR you've submitted recently. You should join the [MessageKit Slack channel](https://join.slack.com/t/messagekit/shared_invite/enQtMjI4NzIzNzMyMzU0LTEyNmYzNzBlNjNiYjAwYjY1ZjhkMmVjNDE4OGNlMDY1OTRmNzVmMzI4MTg3OWQzZTM2NGU4NTU3ZDE5NzkwNzQ) so we can discuss design decisions. 

I think some MessageKit users have already implemented this functionality on their own, it would be good to discuss/compare it to what they've done.

nathantannar4 (changes requested):
Hey @azurechen I like this feature! I agree with @SD10 that we shouldn't override `reloadData` to get this functionality. Perhaps making a separate function that inserts new messages

azurechen:
I have consider using another method to insert old messages. But let me explain why I think override  `reloadData` is a better idea.

First, base on the pattern of original `UICollectionView` and `UITableView`. Developer should use the same `messageList` to handle the data source. If we use another method to insert old data not directly change `messageList` will break this pattern.
In other word, update the `messageList` is necessary, and `reloadData` should be necessary after developer change the value of `messageList`.

Second, in a sense I didn't really override the `reloadData` method because if not trigger the `refreshing`. The behavior is definitely same as the original `reloadData`. I just extend the ability.

```swift
if !messageLoadMoreControl.isRefreshing {
    super.reloadData()
    return
}
``` 

Extend `reloadData` can make load more function simple and elegant. Only needs two lines can enable load more function.

Register the selector
```swift
messageLoadMoreControl.addTarget(self, action: #selector(loadMoreMessages), for: .valueChanged)
```

And insert older data to the `messageList`
```swift
self.messageList.insert(contentsOf: messages, at: 0)
```

I have been thinking about how to use a beautiful way to implement it for a few days. I really want to contribute because I use `MessageKit` in my project even it is still on the beta state.
I may not often discuss on the Slack because my job. But I already join the room and keep following your discussion.

And please try the example, I believe the effect is satisfying.

Sources/Views/MessagesCollectionView.swift
azurechen: Hi I have joined slack few days ago. Thanks so much.
I have explained my idea on the following comment. I don't think I really override the `reloadData` method I just extend it.

cwalo:
@azurechen I thought about this a bit more and I mentioned in slack, but I think a second method - `reloadAndMaintainOffset()` - might be more ideal. It makes the behavior clear and provides a mechanism for users to use whatever reload/refresh control they want. It would be up to the user to decide to call it or the standard `reloadData` method.

azurechen:
@cwalo I think it is a perfect compromise!
I will take your suggestion and update the PR soon. Thank you.

Sources/Views/MessagesCollectionView.swift
```diff
@@ -35,6 +35,8 @@ open class MessagesCollectionView: UICollectionView {
     open weak var messagesLayoutDelegate: MessagesLayoutDelegate?
 
     open weak var messageCellDelegate: MessageCellDelegate?
+    
+    open var messageLoadMoreControl = UIRefreshControl()
```
cwalo: I'm not so sure this should live in the collectionview. Perhaps move to MessagesViewController and add a method for adding/removing it. Similar to JSQ in that regard.

Sources/Views/MessagesCollectionView.swift
```diff
@@ -67,5 +70,31 @@ open class MessagesCollectionView: UICollectionView {
         guard let indexPath = indexPathForLastItem else { return }
         scrollToItem(at: indexPath, at: .bottom, animated: animated)
     }
+    
+    public func reloadAndMaintainOffset() {
+        if !messageLoadMoreControl.isRefreshing {
```
cwalo: I'm waffling between whether or not this method should do any checking of the refresh control or if the user should do it. Otherwise, it should _only_ calculate the new offset and refresh. If this were the case, the user could do any checking they need to, then just call `super`. @SD10 thoughts?

cwalo (commented):
Can't approve until we get some more discussion on this.

Sources/Views/MessagesCollectionView.swift
azurechen: I agree moving it to `MessagesViewController` make sense. but because of `!messageLoadMoreControl.isRefreshing`. it have to be put in `MessageCollectionView`

nathantannar4: I agree. It could also just be called `refreshControl` and lie in `MessagesViewController`

SD10: Yep, agree with you both 👍

Sources/Views/MessagesCollectionView.swift
SD10: I agree, the user should probably do this outside of the call to `reloadAndMaintainOffset()` to determine if the method is called at all. Also, may need to improve method naming on this one 🤔

Sources/Views/MessagesCollectionView.swift
SD10: Possibly, we don't even need to provide this control? The method is the most important thing. Then the user can hook it up to whatever they want? Need to think about this a little bit more

azurechen: There is a possible issue that different control component has different sliding animation duration. if the animation not stop, it is hard to calculate the right offset.
Do you think we can temporarily keep this control, and provide another interface to user in later version to customize their own refreshControl?

Sources/Views/MessagesCollectionView.swift
azurechen: @SD10 could you suggest a name? I think the method name should include reloadData to imply that is a extended method of reloadData

Sources/Views/MessagesCollectionView.swift
azurechen: I though an idea. Create a new class name `MessageMoreControl` but not extends `UIRefreshControl`.
And user have to setup the component by

```
messageMoreControl = MessageMoreControl(component: UIRefreshControl())
messageMoreControl.delegate = self
```

user can choose use `UIRefreshControl` or customized view

```
messageMoreControl = MessageMoreControl(component: CustomRefreshView())
messageMoreControl.delegate = self
```

and provide a `MessageMoreControlDelegate` for user load more data

```
protocol MessageMoreControlDelegate {

    func didPullDown(of messageMoreControl: MessageMoreControl)
}
```

Example/Sources/ConversationViewController.swift
```diff
@@ -28,16 +28,28 @@ import MapKit
...
+                    self.refreshControl.endRefreshing()
+                }
             }
         }
     }
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        
+        automaticallyAdjustsScrollViewInsets = true
```
cwalo: Is this relevant to the PR?

cwalo (approved):
One question about `automaticallyAdjustsScrollViewInsets = true`. Otherwise, looks good. 👍

Example/Sources/ConversationViewController.swift
azurechen: for the UIRefreshControl of this example. If not set automaticallyAdjustsScrollViewInsets, the refreshControl will be covered by navigation bar.

nathantannar4 (approved):

alijaza:
Will this be part of `v0.11.0`?

SD10:
@alijaza I'm reviewing this again now

I'm wondering if this functionality can be moved into the `MessagesCollectionViewFlowLayout`,
then users can toggle a flag to enable/disable this functionality.

See [this comment](https://github.com/jessesquires/JSQMessagesViewController/issues/1838#issuecomment-249215048)

cc @cwalo What do you think?

SD10 (approved):

SD10:

PR #295, opened 2017-10-27
Default frames based on device orrientation

nathantannar4:
Updates the default width/height protocols to return a value based on the devices orientation

Changes This
simulator screen shot - iphone x - 2017-10-27 at 13 11 22

To This
simulator screen shot - iphone x - 2017-10-27 at 13 33 59

And
simulator screen shot - iphone x - 2017-10-27 at 13 33 52


nathantannar4:
@SD10 I used the master branch because it was the latest, idk if that was the wrong one. Also where would the change log entry be put?

SD10:
@nathantannar4 Are we sure this is a bug? It's using all the available space as expected.

Issue #296, opened 2017-10-29
Support undocked/split keyboard on iPad


antonselyanin:
## General Information

* MessageKit Version:
v0.10.0

* iOS Version(s):
11

* Swift Version:
Swift 4

* Devices/Simulators:

* Reproducible in ChatExample? (Yes/No):
Yes

1. Keyboard handling code doesn't support undocked/split keyboard 
https://support.apple.com/en-us/HT207521

2. When the keyboard changes size it doesn't adjust scrolling position (the keyboard covers part of the content)

SD10:
This could be reconsidered for the next version. There's already #297 which aims to solve this

Kaspik:
Should be updated via https://github.com/MessageKit/MessageKit/pull/297

martinpucik:
This should be resolved in #1704

PR #297, opened 2017-10-29
Support undocked keyboard

antonselyanin:
1. Added support for undocked keyboard
In order to support undocked keyboard we need to handle UIKeyboardDidChangeFrame notification.
I tried to mimick the behavior of iMessage app.

2. Fixed problem with keyboard covering the bottom of the content after growing (adjust scrolling position)

3. Tried to fix UI scrolling position after rotation (see `viewWillTransition` in MessagesViewController)
I think there should be a better way to preserver scrolling position.

Yes, #296 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …
Simulators iPhone X, iPad, iPhone 5s

**iOS Version:** …
iOS 9.3, iOS 10.3.1, iOS 11

**Swift Version:** …
Swift 4

**MessageKit Version:** …
v0.10.0


Sources/Controllers/MessagesViewController.swift
```diff
@@ -251,49 +251,77 @@ extension MessagesViewController: UICollectionViewDataSource {
...
 
-        } else {
-            //Software keyboard is found
-            let bottomInset = keyboardEndFrame.height > keyboardOffsetFrame.height ? (keyboardEndFrame.height - iPhoneXBottomInset) : keyboardOffsetFrame.height
-            messagesCollectionView.contentInset.bottom = bottomInset
-            messagesCollectionView.scrollIndicatorInsets.bottom = bottomInset
+        guard shouldUpdateBottomInset else { return }
+
+        UIView.animate(withDuration: 0.5) {
+            self.adjustBottomContentInset(bottomInset ?? 0)
```
SD10, line 311: Does this need to be in an animation block?

SD10:
@antonselyanin Thanks for this PR 👍 I'm still trying to review it, I honestly didn't want to change the keyboard handling logic because it is the cause of so many issues...

nathantannar4:
Hi @antonselyanin can you explain what you mean by, "Added support for undocked keyboard" I don't understand from the code changes what this fixes.

Sources/Controllers/MessagesViewController.swift
```diff
@@ -251,49 +251,77 @@ extension MessagesViewController: UICollectionViewDataSource {
         return messagesLayoutDelegate.footerViewSize(for: message, at: indexPath, in: messagesCollectionView)
     }
 
+    open override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
```
nathantannar4, line 254: iMessage nor Facebook messenger support remembering the scroll position on rotation. So it may be better to leave this behavior to the user to decide if it is what they want

Sources/Controllers/MessagesViewController.swift
```diff
@@ -251,49 +251,77 @@ extension MessagesViewController: UICollectionViewDataSource {
...
 
     func removeKeyboardObservers() {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.removeObserver(self, name: .UIKeyboardDidChangeFrame, object: nil)
         NotificationCenter.default.removeObserver(self, name: .UITextViewTextDidBeginEditing, object: messageInputBar.inputTextView)
     }
 
     @objc
-    func handleTextViewDidBeginEditing(_ notification: Notification) {
+    private func handleTextViewDidBeginEditing(_ notification: Notification) {
```
nathantannar4, line 285: This doesn't need `private`

Sources/Controllers/MessagesViewController.swift
```diff
@@ -251,49 +251,77 @@ extension MessagesViewController: UICollectionViewDataSource {
...
+    /// This leads to wierd animations and an incorrect state of `messagesCollectionView` bottom inset.
+    /// Luckly we always receive the keyboard end frame in `UIKeyboardDidChangeFrame` notification.
+    /// We can use it to animate the UI to the correct state.
+    ///
+    /// So, if the keyboard end frame in `UIKeyboardWillChangeFrame` has zero height,
+    /// we ignore it and calculate layout adjustments in `UIKeyboardDidChangeFrame` handler instead.
     @objc
-    func handleKeyboardDidChangeState(_ notification: Notification) {
-
+    private func handleKeyboardDidChangeState(_ notification: Notification) {
```
nathantannar4, line 302: Or here. Its in a `fileprivate` extension

SD10:
@nathantannar4 Check the link in #296 for more context

Sources/Controllers/MessagesViewController.swift
```diff
@@ -303,11 +331,40 @@ fileprivate extension MessagesViewController {
...
+
+        // check if keyboard covers our view
+        if intersection.height > 0 {
+            // We can't use `intersection.height` as `keyboardHeight` because the keyboard can be in undocked mode
+            keyboardHeight = view.bounds.maxY - keyboardFrame.minY
+        } else {
+            keyboardHeight = 0
+        }
+
+        return keyboardHeight - iPhoneXBottomInset
```
nathantannar4, line 359: This all could be cleaned up into a smaller function.

`return intersection.height > 0 ? (view.bounds.maxY - keyboardFrame.minY - iPhoneXBottomInset ) : -iPhoneXBottomInset`

nathantannar4 (changes requested):

SD10:
I haven't got to run it yet but 2 concerns that jump out at me are:
1. Now we're registered for the `.UIKeyboardDidChangeFrame` notification which may mess up the scrolling behavior & timing
2. I lost track of what was happening for hardware keyboards

Sources/Controllers/MessagesViewController.swift
antonselyanin: `private` modifier makes those methods inaccessible from outside of the extension (if I'm not mistaken). With `fileprivate extension` - other extensions of this class can invoke the methods which could be undesirable.

But, I don't think it's that important to keep those methods as close as possible, `fileprivate extension` also works :)
I can remove it.

nathantannar4:
Had a chance to run your changes. Looks like the correct bottom inset is still being calculated correctly (Y)

Sources/Controllers/MessagesViewController.swift
antonselyanin: Well, at least iMessage remembers that is was pointing to the bottom before the rotation. I think this a polite behavior towards the user. Maybe this - keeping scrolling position at bottom after rotation - should be addressed in another PR. What do you think?

Sources/Controllers/MessagesViewController.swift
antonselyanin: I didn't use elvis operator here because I wanted add this comment: "We can't use `intersection.height` as `keyboardHeight` because the keyboard can be in undocked mode".
Otherwise this code might confuse - why not just use `intersection.height`?

Sources/Controllers/MessagesViewController.swift
antonselyanin: Yes, we have to use animation here, otherwise it'll just snap into place.
I tried to mimick iMessage behavior here.

Should MessageKit mimick iMessage behavior? This is a question :)

antonselyanin:
@SD10 
> I honestly didn't want to change the keyboard handling logic because it is the cause of so many issues...
> I lost track of what was happening for hardware keyboards

Well, the previous keyboard handling code had some issues :)

1. AFAIK it doesn't matter what keyboard the user has. You only care whether it covers your view or not.

2. Keyboard end frame CGRect that come from the notification defined in the screen coordinate system.
It is incorrect to do math on this rect without coverting it to the appropriate coordinates.
This code is potentially buggy 
`let bottomInset = view.frame.size.height - keyboardEndFrame.origin.y - iPhoneXBottomInset`
`view.frame` and `keyboardEndFrame` have different coordinate systems.

You can learn more about this in this year's WWDC session 242:
https://developer.apple.com/videos/play/wwdc2017/242/

Here is "canonical" way of getting the keboard height (copied from the slides)

```
guard let userInfo = notification.userInfo,
let frame = userInfo[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }
let convertedFrame = view.convert(frame, from: UIScreen.main.coordinateSpace)
let intersectedKeyboardHeight = view.frame.intersection(convertedFrame).height
```

OK, this is PR is about undocked/split mode, but maybe it's not that important to support it if it brings problems.

I suggest that we pick one of the two options:

1. MessageKit will not mimick iMessage behavior of undocked keyboard.
Instead, we just do what Apple suggested in 242 session - treat it like keyboard is hidden.

pros: no extra explicit animation blocks in the keyboard handling code
cons: not mimicking iMessage (not a strong cons though :))

2. MessageKit mimicks iMessage behavior of undocked keyboard

pros: it looks and feels like iMessage
cons: more code, extra headache with explicit animation and potential issues :)

If we go with (1), I'll remove `UIKeyboardDidChangeFrame` and rework the keyboard handling code in accordance with Apple recommendations.

What do you think?

cyrilzakka:
Any updates on this? Lovely library btw

nathantannar4:
@SD10 We should revisit this if we want it in 2.1, else push it back

exceptioncatcher91:
Any updates on this? I really need this functionality. Is there any workaround to achieve this without this PR?

hyouuu:
Hi @antonselyanin now I'm kinda taking over the PR reviews of MessageKit, could you kindly pull from develop to resolve conflicts and then I can test and likely merge this? Thanks!

antonselyanin:
@hyouuu I'll give it a try this weekend :)

hyouuu:
Thanks @antonselyanin ! btw currently a few tests are broken and I fixed them in https://github.com/MessageKit/MessageKit/pull/1247

Sources/Controllers/MessagesViewController.swift
Kaspik: I agree this should be addressed in another PR. :)

Sources/Controllers/MessagesViewController.swift
Kaspik: Still not updated.

Sources/Controllers/MessagesViewController.swift
```diff
@@ -251,49 +251,77 @@ extension MessagesViewController: UICollectionViewDataSource {
...
     func addKeyboardObservers() {
-        NotificationCenter.default.addObserver(self, selector: #selector(handleKeyboardDidChangeState), name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(handleKeyboardWillChangeState), name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(handleKeyboardDidChangeState), name: .UIKeyboardDidChangeFrame, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(handleTextViewDidBeginEditing), name: .UITextViewTextDidBeginEditing, object: messageInputBar.inputTextView)
     }
 
     func removeKeyboardObservers() {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.removeObserver(self, name: .UIKeyboardDidChangeFrame, object: nil)
```
Kaspik, line 280: Would probably remove this for this PR as discussed in other comments.

Kaspik (changes requested):
As this is really really old PR - it would be great to first:
- merge `development` branch
- update it for iOS 13
- attach GIF that shows working result (we can also test it of course, but I believe if you are in updating already, why not record it :) )

Thanks!

bguidolim:
+1 for this one

Kaspik:
We are getting rid of develoment branch, thus this PR was autoclosed. Please open it again against master. :)

Kaspik:
@antonselyanin Hey! Are you planning to re-open the PR and update it for current master?

antonselyanin:
Sorry, I tried already. Too many conflicts.

PR #299, opened 2017-10-30
[WIP] Rewriting MessageLabel

SD10:


Sources/Views/MessageLabel.swift
```diff
@@ -23,6 +23,132 @@
...
+    
+    private lazy var layoutManager: NSLayoutManager = {
+        let manager = NSLayoutManager()
+        manager.addTextContainer(self.textContainer)
+        return manager
+    }()
+    
+    private lazy var textContainer: NSTextContainer = {
+        let container = NSTextContainer()
+        container.lineFragmentPadding = 0
```
SD10: Should remember to use the values from the backing properties once they're added

Sources/Views/MessageLabel.swift
```diff
@@ -23,6 +23,298 @@
...
+        }
+    }
+    
+    // MARK: - Parsing
+    
+    func setStorageByParsing(text: NSAttributedString?, completion: @escaping () -> Void) {
+        
+        removeAllDetectedRanges()
+        
+        guard let text = text, text.length > 0, !enabledDetectors.isEmpty else { return }
```
SD10: Maybe do all this in background too... kinda trivial but yeah

Sources/Views/MessageLabel.swift
```diff
@@ -23,6 +23,298 @@
...
+            
+            setNeedsDisplay()
+        }
+    }
+    
+    public var font: UIFont = UIFont.systemFont(ofSize: 10.0) {
+        didSet {
+            textStorage.addAttribute(key: .font, value: font)
+            // Doesn't apply to detected ranges
+            addAttributesToTextStorage(for: enabledDetectors)
```
SD10: probably overkill to apply all the attributes again. instead we should only apply the font attribute that changed

Sources/Views/MessageLabel.swift
```diff
@@ -22,18 +22,552 @@
...
+// allowsDefaultTighteningForTruncation
+// baselineAdjustment
+// minimumScaleFactor
+// highlightedTextColor
+// isHighlighted
+// shadowColor
+// shadowOffset
+
 import UIKit
+extension NSTextStorage {
```
SD10: Needs to go in its own file

Sources/Views/MessageLabel.swift
```diff
@@ -22,18 +22,552 @@
...
+    func detectorAttributes(for detectorType: DetectorType) -> [NSAttributedStringKey: Any] {
+        switch detectorType {
+        case .address: return addressAttributes
+        case .date: return dateAttributes
+        case .phoneNumber: return phoneAttributes
+        case .url: return urlAttributes
+        }
+    }
+    
+    func detectorResults(for detectorType: DetectorType) -> [NSRange: Any?] {
```
SD10: Maybe we can lose the `Any?` typing here

Sources/Views/MessageLabel.swift
```diff
@@ -22,18 +22,552 @@
...
+        
+    }
+    
+    private func setStorageByParsing(text: NSAttributedString?, completion: @escaping () -> Void) {
+        
+        removeAllDetectedResults()
+
+        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+            
+            defer { completion() }
```
SD10: Needs to get called on main thread

Sources/Views/MessageLabel.swift
```diff
@@ -22,18 +22,552 @@
...
+    private var urlResults: [NSRange: URL?] = [:]
+    
+    private var isConfiguring: Bool = false
+    
+    private var paragraphStyle: NSParagraphStyle {
+        
+        guard let text = attributedText else { return NSParagraphStyle() }
+        
+        var range = NSRange(location: 0, length: 0)
+        let existingStyle = text.attribute(.paragraphStyle, at: 0, effectiveRange: &range) as? NSMutableParagraphStyle
```
SD10: Crashing -- docs explicitly mentions a crash. Need to figure this out. May have to apply it in a separate method

SD10:
That broke a lot of things... A task for tomorrow 😅

Sources/Views/MessageLabel.swift
```diff
@@ -462,4 +522,46 @@ open class MessageLabel: UILabel, UIGestureRecognizerDelegate {
...
     }
     
+    // MARK: UIGestureRecognizer Delegate
+    
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        return true
+    }
+    
+    //swiftlint:disable cyclomatic_complexity
+    // Yeah we're disabling this because the whole file is a mess :D
```
nathantannar4, line 550: Lol 😆

Sources/Views/MessageLabel.swift
```diff
@@ -22,415 +22,477 @@
...
+    public weak var delegate: MessageLabelDelegate?
+    
+    public var attributedText: NSAttributedString? {
+        get {
+            return textStorage.attributedString
+        }
+        set {
+            let string = newValue ?? NSAttributedString(string: "")
+            textStorage.setAttributedString(string)
+
```
SD10: We may want to strip some of the attributes from this, `textColor`, `font`, and set the label properties. That will result in a lot of inefficiencies. Better way would be just to get things like `textColor` from `textStorage` maybe

Sources/Views/MessageLabel.swift
```diff
@@ -22,415 +22,477 @@
...
+    
+    public var textColor: UIColor = .darkText {
         didSet {
-            guard textColor != oldValue else { return }
-            guard let attributedText = attributedText else { return }
-            textStorage.setAttributedString(attributedText)
-            setNeedsDisplay()
+            textStorage.addAttribute(key: .foregroundColor, value: textColor)
+            updateDetectorAttributes()
+            if !isConfiguring { setNeedsDisplay() }
```
SD10: This can be done better. We shouldn't do any work if the `isConfiguring` flag is true

Sources/Views/MessageLabel.swift
```diff
@@ -462,4 +524,46 @@ open class MessageLabel: UILabel, UIGestureRecognizerDelegate {
...
+    // MARK: UIGestureRecognizer Delegate
+    
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        return true
+    }
+    
+    //swiftlint:disable cyclomatic_complexity
+    // Yeah we're disabling this because the whole file is a mess :D
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
+
```
SD10, line 552: Gestures are broken because logic is so brittle. Would be good to find a better solution this time around 😓

Sources/Views/MessageLabel.swift
```diff
@@ -22,415 +22,477 @@
...
-
-    private func handleGesture(for detectorType: DetectorType, value: Any?) {
-
-        switch detectorType {
+    
+    private func forwardToDelegateMethod(for detector: DetectorType, value: Any?) {
+        switch detector {
         case .address:
             guard let addressComponents = value as? [String: String] else { return }
             handleAddress(addressComponents)
```
SD10: This is broken because `value: Any?` is a `[NSTextCheckingKey : String]?`. I'm not sure how this cast worked in the old code to be honest. I know `NSTextCheckingKey` has a `RawValue` of `String`. 

Maybe the old code was so bad that I'm A) missing something, B) got away with some sort of lossy conversion. 

Point is I have to either break the API and pass in `[NSTextCheckingKey: String]` or somehow get to `[String: String]` again. I don't mind making a change here as opposed to mapping the collection

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
+        set {
+            let attributedString = NSAttributedString(string: newValue ?? "")
+            let string = addLocalAttributes(to: attributedString)
+            textStorage.setAttributedString(string)
+            
+            setStorageByParsing(text: string) {
+                if !self.isConfiguring { self.setNeedsDisplay() }
+            }
+            
+            if !isConfiguring { setNeedsDisplay() }
```
nathantannar4: Why are we calling setNeedsDisplay twice here?

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
         }
     }
-
-    open override var font: UIFont! {
+    
+    public var font: UIFont = UIFont.systemFont(ofSize: 10.0) {
```
nathantannar4: This is nit picky but I think a default size of 10 is too small. I think standard is like 14 or 15

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
-    open override var textColor: UIColor! {
+    
+    public var textColor: UIColor = .darkText {
         didSet {
-            guard textColor != oldValue else { return }
-            guard let attributedText = attributedText else { return }
-            textStorage.setAttributedString(attributedText)
-            setNeedsDisplay()
+            textStorage.addAttribute(key: .foregroundColor, value: textColor)
+            updateDetectorAttributes()
```
nathantannar4: Do we need to update the detectors when we change the text color?

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
         didSet {
-            guard textInsets != oldValue else { return }
-            setNeedsDisplay()
+            if !isConfiguring { setNeedsDisplay() }
         }
     }
-
-    open var addressAttributes: [NSAttributedStringKey: Any] = [:] {
+    
+    public var enabledDetectors: [DetectorType] = [.address, .phoneNumber, .date, .url] {
```
nathantannar4, line 133: From a performance standpoint wouldn't it be better to have the default detectors be empty? Experienced users can then set the detectors they want

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
+        self.phoneAttributes = defaultAttributes
         self.dateAttributes = defaultAttributes
-        self.phoneNumberAttributes = defaultAttributes
         self.urlAttributes = defaultAttributes
-
+        
         setupGestureRecognizers()
-
+        
+        backgroundColor = .clear
```
nathantannar4: I think by default the background of a UIView is already nil (clear)

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
     }
-
-    // MARK: - Public Methods
-
-    // MARK: UIGestureRecognizer Delegate
-
-    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
-        return true
+    
+    public func configure(configurationBlock: () -> Void) {
```
nathantannar4: I think in the config block you should have a reference to message label and return Self.

```
@discardable_result
public func configure(configurationBlock: (MessageLabel) -> Void) -> Self
```

This could allow for clean inits like.
```
let label = MessageLabel().configure { label in
     // config here
}
```

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
     }
-
-    // MARK: - Parsing Text
-
-    private func parse(text: NSAttributedString, for detectorTypes: [DetectorType]) -> [NSTextCheckingResult]? {
-        guard detectorTypes.isEmpty == false else { return nil }
-        let checkingTypes = detectorTypes.reduce(0) { $0 | $1.textCheckingType.rawValue }
+    
+    private func parse(text: String) -> [NSTextCheckingResult]? {
+        let checkingTypes = enabledDetectors.reduce(0) { $0 | $1.textCheckingType.rawValue }
```
nathantannar4, line 357: I would add `if enabledDetectors.isEmpty { return nil }` here as well

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
-        for (detectorType, ranges) in rangesForDetectors {
-            for (nsRange, value) in ranges {
-                guard let range = Range(nsRange) else { return }
-                if range.contains(index) {
-                    handleGesture(for: detectorType, value: value)
-                }
+        
+        var detectorRangeMap: [DetectorType: [NSRange]] = [:]
+        
+        for detector in enabledDetectors where !enabledDetectors.isEmpty {
```
nathantannar4, line 471: ooo cool didn't know you could add a where in a for loop

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
-    private func handleGesture(for detectorType: DetectorType, value: Any?) {
-
-        switch detectorType {
+    
+    func foo(value: Any?) {
+        let isCheckSucceeded = value is [NSTextCheckingKey: String]
+        print("Value is [NSTextCheckingKey: String]: \(isCheckSucceeded)")
+        let castSucceeded = value as? [String: String] != nil
+        print("The cast to [String: String] worked: \(castSucceeded)")
+    }
```
nathantannar4: I presume this will be taken out for release?

Sources/Views/MessageLabel.swift
nathantannar4: How about rather than using gestures use the `touchesBegan` function thats in every `UIView`

SD10: I can look into it but the good thing about gestures are:
1. long press gestures
2. they are cancellable (which we need)

Sources/Views/MessageLabel.swift
SD10: The one inside the block is `async`. We need some refactoring around this for sure. I will probably add a flag to the configure method: `configure(parseText: Bool)` or something

Sources/Views/MessageLabel.swift
SD10: Good point. We should use whatever the default is for UIKit

Sources/Views/MessageLabel.swift
SD10: Yeah :( Working on this but for example if the user is using a `UIColor.red` for the textColor of detected attribtues, when we change the font here it will override the detected attributes. So we have to reapply them over it. This method has an optional parameter `key` that allows us to update the detectors for just that key. I _may_ be able to use that here but not sure

Sources/Views/MessageLabel.swift
SD10: Yeah, we definitely will default to none. These will be overriden by the `TextMessageDisplayDelegate` that defaults to `[]` anyways but I will clean it up

Sources/Views/MessageLabel.swift
SD10: For some reason it was `UIColor.black`

Sources/Views/MessageLabel.swift
SD10: Possibly but right now it won't be used outside the cell

Sources/Views/MessageLabel.swift
SD10: Yeah, I'm struggling with this. Some of these methods are called within over methods where the check has already been performed. I started removing them because of this

Sources/Views/MessageLabel.swift
nathantannar4: 👍

Sources/Views/MessageLabel.swift
nathantannar4: Ah I see

Sources/Views/MessageLabel.swift
nathantannar4: 👍

Sources/Views/MessageLabel.swift
nathantannar4: 👍

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
+    // MARK: - Properties [Public]
+    
+    public weak var delegate: MessageLabelDelegate?
+    
+    public var attributedText: NSAttributedString? {
+        get {
+            return _textSync
+        }
+        set {
+            _textSync = newValue
```
SD10, line 67: It would be nice to strip these values and apply to the label properties still :/

Sources/Views/MessageLabel.swift
```diff
@@ -22,417 +22,491 @@
...
+    }
+    
+    // MARK: - Manager TextStorage
+    
+    func updateTextStorage(text: NSAttributedString) {
+        
+        textStorage.setAttributedString(text)
+        
+        DispatchQueue.global(qos: .userInitiated).async {
+
```
SD10, line 257: Unless I use something like `NSOperation` to cancel these the entire effort is futile. Crashes as is and will likely lead to `TextStorage` corruption

Issue #268, opened 2017-10-18
General layout issues


jackdem:
## General Information

* MessageKit Version: latest

* iOS Version(s): 11

* Swift Version: 3

* Devices/Simulators: 8 sim

* Reproducible in ChatExample? (Yes/No):

##  What happened?
I have several layout issues with the library, for example the chat window is hidden behind the message typing box, the view doesnt scroll to the bottom on load, opening the full keyboard doesnt life the chat view and obscures it. 

<img width="402" alt="screen shot 2017-10-18 at 11 59 56 am" src="https://user-images.githubusercontent.com/31346145/31715517-25b99344-b3fc-11e7-9bd3-3b02cb3dfd1f.png">

<img width="387" alt="screen shot 2017-10-18 at 12 02 28 pm" src="https://user-images.githubusercontent.com/31346145/31715546-40f497a8-b3fc-11e7-8f3c-bbf49ec126eb.png">

<img width="402" alt="screen shot 2017-10-18 at 12 02 55 pm" src="https://user-images.githubusercontent.com/31346145/31715561-4b7b92b2-b3fc-11e7-8cc6-1c264ecb068f.png">

## What did you expect to happen?
adapts correctly

SD10:
Hey @jackdem,
There is a bug with the top content inset on master. It should be fixed with the next relase.

As for scrolling/keyboard handling. You should look at the `scrollsToBottomOnFirstLoad` and `scrollsToBottomOnKeyboardDidBeginEditing` properties of `MessagesViewController`. If there are any other layout problems please be specific 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jackdem:
do i need to do any additional setup for those calls as either of them in the viewDidLoad doesnt scroll down correctly

```
override func viewDidLoad() {
        super.viewDidLoad()
        self.title = self.conversation.firstname + " " + self.conversation.lastname
        messagesCollectionView.messagesDataSource = self
        messagesCollectionView.messagesLayoutDelegate = self
        messagesCollectionView.messagesDisplayDelegate = self
        messagesCollectionView.messageCellDelegate = self
        messageInputBar.delegate = self
        scrollsToBottomOnFirstLayout = true
        scrollsToBottomOnKeybordBeginsEditing = true
        scheduledTimerWithTimeInterval()
        self.tabBarController?.tabBar.isHidden = true
    }
```

SD10:
[...]

PR #301, opened 2017-10-30
[WIP] Fixing scroll to bottom

antonselyanin:
Hey, I want to try to this issue #268 with scrolling to bottom.

I did some experiments, my current version of scrolling works fairly reliably.
Unfortunately, it uses `DispatchQueue.main.async`, which is kinda hacky.

JSQMessagesViewController had a similar issue: https://github.com/jessesquires/JSQMessagesViewController/issues/256
I think we should study it. Maybe we can just copy and document it properly.

#268 

Where has this been tested?
---------------------------
**Devices/Simulators:** 
iPhone and iPad sims

**iOS Version:** 
iOS 9, 11

**Swift Version:** 
Swift 4

**MessageKit Version:** 
v0.11.0

### DIFF
--- a/Sources/Views/MessagesCollectionView.swift
+++ b/Sources/Views/MessagesCollectionView.swift
@@ -64,8 +64,19 @@ open class MessagesCollectionView: UICollectionView {
     // MARK: - Methods
 
     public func scrollToBottom(animated: Bool = false) {
-        guard let indexPath = indexPathForLastItem else { return }
-        scrollToItem(at: indexPath, at: .bottom, animated: animated)
+//        guard let indexPath = indexPathForLastItem else { return }
+//        scrollToItem(at: indexPath, at: .bottom, animated: animated)
+
+        // Work in progress here :)
+
+        guard indexPathForLastItem != nil else { return }
+
+        // This works fairly reliably, but using `DispatchQueue.main.async` might not let you invoke this method from an animation block :(
+        DispatchQueue.main.async {
+            self.layoutIfNeeded()
+            let scrollRect = CGRect(x: 0, y: self.contentSize.height - 1, width: 1, height: 1)
+            self.scrollRectToVisible(scrollRect, animated: animated)
+        }
     }
 
 }
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -382,8 +382,7 @@ extension ConversationViewController: MessageInputBarDelegate {
     func messageInputBar(_ inputBar: MessageInputBar, didPressSendButtonWith text: String) {
         messageList.append(MockMessage(text: text, sender: currentSender(), messageId: UUID().uuidString, date: Date()))
         inputBar.inputTextView.text = String()
-        messagesCollectionView.reloadData()
-        messagesCollectionView.scrollToBottom()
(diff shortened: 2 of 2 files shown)


Sources/Views/MessagesCollectionView.swift
```diff
@@ -64,8 +64,19 @@ open class MessagesCollectionView: UICollectionView {
     // MARK: - Methods
 
     public func scrollToBottom(animated: Bool = false) {
-        guard let indexPath = indexPathForLastItem else { return }
```
SD10, line 67: The problem here is the indexPath is `nil` by the time this method is called. I need to find a better opportunity to call it.

Sources/Views/MessagesCollectionView.swift
```diff
@@ -64,8 +64,19 @@ open class MessagesCollectionView: UICollectionView {
...
-        scrollToItem(at: indexPath, at: .bottom, animated: animated)
+//        guard let indexPath = indexPathForLastItem else { return }
+//        scrollToItem(at: indexPath, at: .bottom, animated: animated)
+
+        // Work in progress here :)
+
+        guard indexPathForLastItem != nil else { return }
+
+        // This works fairly reliably, but using `DispatchQueue.main.async` might not let you invoke this method from an animation block :(
```
SD10, line 74: @antonselyanin Just touching on somethings
- What were your concerns here, you mention "fairly reliably"?
- Why have we wrapped this in `DispatchQueue.main.async`?

antonselyanin: > What were your concerns here, you mention "fairly reliably"?

I don't remember exactly wrote it :(. Maybe there were case where it wouldn't scroll.

> Why have we wrapped this in DispatchQueue.main.async

It looks like we have to skip one run loop pass before we can invoke `layoutIfNeeded()`. UIKit needs this for some reason. At least, this is what I got from my experiments with this issue.

I think this is a dirty hack, which relies on unknown mysteries of UIKit (or, I just missing something).

IMO, this issue worth looking at: https://github.com/jessesquires/JSQMessagesViewController/issues/256

SD10: Thanks for investigating this @antonselyanin. What do you think of [this approach?](https://github.com/MessageKit/MessageKit/issues/268#issuecomment-339680511)

I don't know if one approach has greater benefit over the other, but I'd like to clean up this PR and get this merged in.

antonselyanin: @SD10, I think the code from that comment might be an adaptation of this code: https://github.com/jessesquires/JSQMessagesViewController/blob/develop/JSQMessagesViewController/Controllers/JSQMessagesViewController.m#L386

Method `- (void)scrollToIndexPath:(NSIndexPath *)indexPath animated:(BOOL)animated`, line 386.

If this can be fixed without using `DispatchQueue.main.async` then the approach you linked to is better.

SD10: Your solution seems to be more consistent but it's still not working all of the time 😓

SD10:
Closing this because it's gone stale and doesn't solve the problem. We're investigating #395. Really appreciate the research you've done here @antonselyanin

PR #311, opened 2017-11-01
Fix left、rightStackView layout issues

zhongwuzw:
Fix left、rightStackView layout issues


zhongwuzw:
height_ambiguous

nathantannar4 (approved):
Thank you for noticing this! I believe I missed it because I still observed the correct functionality. Since there is now a topAnchor we also need to adjust the `InputStackView` initial alignment mode from `alignment = .fill` to `alignment = .bottom`. Can you make that change too! 👍

Sources/Views/MessageInputBar.swift
```diff
@@ -252,14 +252,14 @@ open class MessageInputBar: UIView {
         textViewHeightAnchor = inputTextView.heightAnchor.constraint(equalToConstant: maxHeight)
         
         leftStackViewLayoutSet = NSLayoutConstraintSet(
-            top:    inputTextView.topAnchor.constraint(equalTo: topAnchor, constant: padding.top),
+            top:    leftStackView.topAnchor.constraint(equalTo: topAnchor, constant: padding.top),
```
nathantannar4, line 255: Because we are setting a topAnchor we need the default `InputStackView` alignment property to be bottom so the buttons don't remain in the center when the textView grows

nathantannar4 (changes requested):

Sources/Views/MessageInputBar.swift
zhongwuzw: @nathantannar4 I updated the requested changes.:beers::beers::beers:

nathantannar4:
🔥

nathantannar4 (approved):

nathantannar4:
@zhongwuzw Can you just make a change log entry for this in the CHANGELOG.md ? These are non-breaking changes but we need to document them

Issue #312, opened 2017-11-01
Application freezes when (Select All | Select | Paste) is clicked in the Input bar.


azhararmar:
## General Information

* MessageKit Version: 0.1.0

* iOS Version(s): 10.3.2

* Swift Version: 4

* Devices/Simulators: iPhone 6s Device & Simulator

* Reproducible in ChatExample? (Yes/No): Yes

##  What happened?
Write a message in the input bar. Hold the touch till the action tool tip is displayed and then click on Select or Select All or Paste. The application freezes.

## What did you expect to happen?
The application must not freeze when clicking on any of the clipboard action buttons.

SD10:
Thanks for reporting this @azhararmar. I wonder where this came from 🤔. cc @nathantannar4

nathantannar4:
This was a really good bug catch @azhararmar  @cwalo. I too can easily replicate the issue. After going through the stack trace on my own I can see that when a user selects text in the textView the intrinsicContentSize is invalidated A LOT. This may be because we invalidate it on the notification of .UITextViewTextDidChange which much be broadcasted a lot. In earlier versions it was added that in the calculation of the intrinsicContentSize that the InputTextViews content size also be invalidated. This seems to create a race condition! As when many .UITextViewTextDidChange notifications are sent it a circular cycle is created. Fix with PR #313

SD10:
This is closed by #313 and on `master`. Please let us know if the problem persists 👍

PR #313, opened 2017-11-01
Bugfix/select menu button

nathantannar4:
Fix to issue #312 

A race condition is was created during the calculation of the `intrinsicContentSize` which lead to issues when selecting text. I have tested this patch and it is not required to also invalidate the text views content size which seems to create the race condition when the `MessageInputBar`s content size is invalidated many times in a row


cwalo (approved):
LGTM 👍

SD10 (approved):
Thanks @nathantannar4 @cwalo 💯

PR #320, opened 2017-11-03
Add a top InputStackView to MessageInputBar

nathantannar4:
Anchors another `InputStackView` to the top of the `MessageInputBar`. This required adding a new padding set but I have added documentation to show how it works. 

Now developers can add elements to the top such as an image picker or table for autocomplete

Any other comments?
-------------------
- Moves the SeparatorLine into the topStackView initially.
- The top `InputStackView` is initially anchored to the top, left and right with no content insets. To adjust this change the `topStackViewPadding`
- The `padding.top` property separates the left/right stack views and `InputTextView` from the top stack view and no longer the top of the `MessageInputBar`. 

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone X Sim, iPhone 7 Sim, iPhone 6s Plus Device

**iOS Version:** 10.3, 11

**Swift Version:** 4

**MessageKit Version:** 0.10.1

Makes this possible!

simulator screen shot - iphone x - 2017-11-02 at 20 54 04
simulator screen shot - iphone x - 2017-11-02 at 20 54 20

### DIFF
--- a/Sources/Views/InputStackView.swift
+++ b/Sources/Views/InputStackView.swift
@@ -40,7 +40,7 @@ open class InputStackView: UIStackView {
     /// - right: Bottom Stack View
     /// - bottom: Left Stack View
     public enum Position {
-        case left, right, bottom
+        case left, right, bottom, top
     }
     
     // MARK: Initialization
--- a/Sources/Views/InputStackView.swift
+++ b/Sources/Views/InputStackView.swift
@@ -39,6 +39,7 @@ open class InputStackView: UIStackView {
     /// - left: Left Stack View
     /// - right: Bottom Stack View
     /// - bottom: Left Stack View
+    /// - top: Top Stack View
     public enum Position {
         case left, right, bottom, top
     }
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -73,7 +73,7 @@ open class MessageInputBar: UIView {
                 blurView.fillSuperview()
             }
             blurView.isHidden = !isTranslucent
-            backgroundView.backgroundColor = isTranslucent ? (backgroundView.backgroundColor?.withAlphaComponent(0.7) ?? UIColor.white.withAlphaComponent(0.7)) : .white
+            backgroundView.backgroundColor = isTranslucent ? (backgroundView.backgroundColor?.withAlphaComponent(0.75) ?? UIColor.white.withAlphaComponent(0.75)) : .white
         }
     }
     
@@ -85,8 +85,13 @@ open class MessageInputBar: UIView {
      
      ## Important Notes ##
      1. It's axis is initially set to .vertical
+     2. It's alignment is initially set to .fill
      */
-    open let topStackView = InputStackView(axis: .vertical, spacing: 0)
+    open let topStackView: InputStackView = {
+        let stackView = InputStackView(axis: .vertical, spacing: 0)
+        stackView.alignment = .fill
+        return stackView
+    }()
     
     /**
      The InputStackView at the InputStackView.left position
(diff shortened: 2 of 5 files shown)


Sources/Views/InputStackView.swift
```diff
@@ -40,7 +40,7 @@ open class InputStackView: UIStackView {
     /// - right: Bottom Stack View
     /// - bottom: Left Stack View
     public enum Position {
```
SD10: Let's add a comment for the `top` position

Sources/Views/MessageInputBar.swift
```diff
@@ -32,29 +33,67 @@ open class MessageInputBar: UIView {
...
+    /// Determines if the MessageInputBar should have a translucent effect
     open var isTranslucent: Bool = false {
         didSet {
+            if isTranslucent && blurView.superview == nil {
+                backgroundView.addSubview(blurView)
+                blurView.fillSuperview()
+            }
             blurView.isHidden = !isTranslucent
-            backgroundColor = isTranslucent ? .clear : .white
+            backgroundView.backgroundColor = isTranslucent ? (backgroundView.backgroundColor?.withAlphaComponent(0.75) ?? UIColor.white.withAlphaComponent(0.75)) : .white
```
SD10: let's drop the `nil coalescing` here? 
```Swift
let color: UIColor = backgroundView.backgroundColor ?? .white
backgroundView.backgroundColor = isTranslucent ? color.withAlphaComonent(0.75) : color
```

Sources/Views/MessageInputBar.swift
```diff
@@ -156,27 +223,30 @@ open class MessageInputBar: UIView {
...
-    private(set) var leftStackViewItems: [InputBarButtonItem] = []
+    private(set) public var leftStackViewItems: [InputBarButtonItem] = []
     
     /// The InputBarItems held in the rightStackView
-    private(set) var rightStackViewItems: [InputBarButtonItem] = []
+    private(set) public var rightStackViewItems: [InputBarButtonItem] = []
     
     /// The InputBarItems held in the bottomStackView
-    private(set) var bottomStackViewItems: [InputBarButtonItem] = []
+    private(set) public var bottomStackViewItems: [InputBarButtonItem] = []
```
SD10: Lets put public first on all of these

Sources/Views/MessageInputBar.swift
```diff
@@ -78,24 +117,16 @@ open class MessageInputBar: UIView {
...
+    /// The InputTextView a user can input a message in
     open lazy var inputTextView: InputTextView = { [weak self] in
         let textView = InputTextView()
         textView.translatesAutoresizingMaskIntoConstraints = false
         textView.messageInputBar = self
         return textView
     }()
-    
-    /// The padding around the textView that separates it from the stackViews
-    open var textViewPadding: UIEdgeInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 8) {
```
SD10, line 90: What's happening here?

SD10: Ok, I see where it went

SD10 (changes requested):
I think the only API breaking part of this is that `InputStackView.Position` now has a `.top` case. So we should make a changelog entry for this under `Added`

Tests/ViewsTests/MessageInputBarTests.swift
```diff
@@ -43,23 +43,17 @@ class MessageInputBarTests: XCTestCase {
...
-        XCTAssertTrue(sut.blurView.isHidden)
-    }
-
     func testIsTranslucent_isFalseForDefault() {
         XCTAssertFalse(sut.isTranslucent)
     }
 
     func testUISetups_forIsTranslucentIsTrue() {
         sut.isTranslucent = true
-        XCTAssertEqual(sut.backgroundColor, UIColor.clear)
```
SD10, line 56: @nathantannar4 You just deleted the failing tests 😅 Let's move forward on this PR in but in the future, we have to fix these when they break. There is a slight maintenance cost for having tests when adding new functionality. I'm fine with it for now because I will be rewriting a lot using Quick/Nimble

SD10 (approved):

PR #322, opened 2017-11-03
Refactor AvatarAlignment -> AvatarPosition

### LINKED ISSUE #318: We should not conflate authorship with message direction (inbound/outbound)
## General Information

MessageKit assumes a 1:1 conversation between a currentSender and a recipient. This assumption is not necessary and makes it difficult to implement group messaging. For example, if I work at a company which has many representatives sending outbound messages to the same customer, my data source may provide an `author` as well as a direction property (`inbound` or `outbound`) for each message object.

* MessageKit Version: 0.10.0

* iOS Version(s): 11

* Swift Version: 4

* Devices/Simulators: NA

* Reproducible in ChatExample? (Yes/No): Yes, just setup data source as explained.

##  What happened?

I tried to set different authorship to some of the outbound messages.

I expect authorship and direction to be decoupled, but `MessagesDataSource` uses `isFromCurrentSender` to decide whether the message is outbound or inbound.

SD10:
This would resolve #318.

It basically combines `AvatarAlignment` (which previously tracked the `AvatarView`'s vertical position) with our internal `AvatarHorizontalAlignment`, into a new struct called `AvatarPosition`. 

The drawback of this is now if a user wants to override the vertical position, they are responsible for managing the horizontal position as well.

I'm probably going to add another case `natural` to the `AvatarHorizontal` enum. This will infer the position by checking if the message is from the current sender or not. I think this can also guide us towards support for both RTL and LTR languages 👍 

TODO:
- [x] Add `natural` case
- [x] CHANGELOG entry
- [x] Remove `MessagesCollectionViewFlowLayoutFlags`
- [x] Fix docs


Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -145,9 +172,16 @@ public extension MessagesLayoutDelegate {
...
         }
     }
-
-    func avatarAlignment(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> AvatarAlignment {
-        return .cellBottom
+    
+    func avatarPosition(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> AvatarPosition {
+        guard let dataSource = messagesCollectionView.messagesDataSource else {
+            return AvatarPosition(horizontal: .cellLeading, vertical: .messageBottom)
+        }
```
SD10: I want to eliminate all these else return cases in the library. Should be a `fatalError()`

Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -22,6 +22,31 @@
  SOFTWARE.
  */
 
+public struct AvatarPosition {
+    
+    public enum Horizontal {
+        case cellLeading
+        case cellTrailing
```
SD10: Here is where I'm talking about the `.natural` case

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -369,6 +367,13 @@ fileprivate extension MessagesCollectionViewFlowLayout {
         return messagesLayoutDelegate.avatarAlignment(for: attributes.message, at: attributes.indexPath, in: messagesCollectionView)
     }
     
+    func avatarPosition(for attributes: MessageIntermediateLayoutAttributes) -> AvatarPosition {
+        var position = messagesLayoutDelegate.avatarPosition(for: attributes.message, at: attributes.indexPath, in: messagesCollectionView)
+        if avatarAlwaysLeading { position.horizontal = .cellLeading }
+        if avatarAlwaysTrailing { position.horizontal = .cellTrailing }
```
SD10: I'd also like to deprecate these flags as they are no longer all that useful

Sources/Protocols/MessagesLayoutDelegate.swift
nathantannar4: Works for me

Sources/Protocols/MessagesLayoutDelegate.swift
nathantannar4: I think `.natural` would be good and have it be default

nathantannar4 (approved):
I think this adds some good flexibility while leaving things more less the same

PR #324, opened 2017-11-03
Refactoring MessageStyle to use CAShapeLayer

SD10:
Mainly opening this to leave comments on my own code. It's really rough and I'm not loving the API.

What I want to accomplish:
- [ ] Restrict MediaMessageTypes to not allow tails
- [ ] Use CAShapeLayer for TextMessageCells

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - MessageKit (0.10.0)
+  - MessageKit (0.10.1)
 
 DEPENDENCIES:
   - MessageKit (from `../MessageKit.podspec`)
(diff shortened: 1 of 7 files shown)


ndonald2:
@SD10 are there still plans to go this route, i.e. use CAShapeLayer instead of stretched images? If not, I've pinpointed a pretty significant performance improvement related to the way `MessageStyle` loads the bubble outline image. It's loading from disk every single time, which is completely unnecessary - the image data can safely be cached in memory. I'll probably open a PR either way but I just was curious what the future held for this.

From a qualitative standpoint, it improves scrolling performance from semi-choppy to what looks close to 60fps in our app running on an iPhone 6. I need to do some profiling around that but from my perspective it seems like a huge improvement.

SD10:
@ndonald2 Yeah moving away from UIImage to CAShapeLayer is one of the biggest improvements we can make. I was just having some difficulty with the API surrounding this. 

SD10:
@ndonald2 Of course, I'm still open to all PRs meanwhile. Feel free to submit any suggestions you have

SD10:
WIll re-evaluate in the future 😜

PR #325, opened 2017-11-03
Fix MessageInputBar layout issues && code optimize

zhongwuzw:
1. Fix ambiguous layout issues when MessageInputBar run layoutIfNeeded
method before added to superview.
2. Just reload new added message when click send button.
3. Change class-only protocols declaration.
4. Change placeholderLabel’s centerX、centerY priority because we need
satisfy left/right、top/bottom constraint first .

### DIFF
--- a/Sources/Protocols/MessagesLayoutDelegate.swift
+++ b/Sources/Protocols/MessagesLayoutDelegate.swift
@@ -26,7 +26,7 @@ import Foundation
 
 /// A protocol used by the `MessagesCollectionViewFlowLayout` object to determine
 /// the size and layout of a `MessageCollectionViewCell` and its contents.
-public protocol MessagesLayoutDelegate: class {
+public protocol MessagesLayoutDelegate: AnyObject {
 
     /// Specifies the insets for the text rect of the `MessageLabel` in a `TextMessageCell`.
     ///
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -443,8 +443,6 @@ open class MessageInputBar: UIView {
     /// - Parameter positions: The UIStackView's to layout
     public func layoutStackViews(_ positions: [InputStackView.Position] = [.left, .right, .bottom, .top]) {
         
-        guard superview != nil else { return }
-        
         for position in positions {
             switch position {
             case .left:
@@ -509,7 +507,6 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     leftStackView.addArrangedSubview($0)
                 }
-                guard superview != nil else { return }
                 leftStackView.layoutIfNeeded()
             case .right:
                 rightStackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
@@ -519,7 +516,6 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     rightStackView.addArrangedSubview($0)
                 }
-                guard superview != nil else { return }
                 rightStackView.layoutIfNeeded()
             case .bottom:
                 bottomStackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
@@ -529,7 +525,6 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     bottomStackView.addArrangedSubview($0)
                 }
-                guard superview != nil else { return }
                 bottomStackView.layoutIfNeeded()
             case .top:
                 topStackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
@@ -539,7 +534,6 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     topStackView.addArrangedSubview($0)
                 }
-                guard superview != nil else { return }
                 topStackView.layoutIfNeeded()
             }
         }
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -79,7 +79,7 @@ open class MessageInputBar: UIView {
      */
     open let bottomStackView = InputStackView(axis: .horizontal, spacing: 15)
         
-    open lazy var inputTextView: InputTextView = { [weak self] in
+    open lazy var inputTextView: InputTextView = {
         let textView = InputTextView()
         textView.translatesAutoresizingMaskIntoConstraints = false
         textView.messageInputBar = self
@@ -319,6 +319,10 @@ open class MessageInputBar: UIView {
     /// - Parameter positions: The UIStackView's to layout
     public func layoutStackViews(_ positions: [InputStackView.Position] = [.left, .right, .bottom]) {
         
+        guard let _ = superview else {
+            return
+        }
+        
         for position in positions {
             switch position {
             case .left:
@@ -378,6 +382,9 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     leftStackView.addArrangedSubview($0)
                 }
+                guard let _ = superview else {
+                    return
+                }
                 leftStackView.layoutIfNeeded()
             case .right:
                 rightStackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
@@ -387,6 +394,9 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     rightStackView.addArrangedSubview($0)
                 }
+                guard let _ = superview else {
+                    return
+                }
                 rightStackView.layoutIfNeeded()
             case .bottom:
                 bottomStackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
@@ -395,6 +405,9 @@ open class MessageInputBar: UIView {
                     $0.messageInputBar = self
                     $0.parentStackViewPosition = position
                     bottomStackView.addArrangedSubview($0)
+                }
+                guard let _ = superview else {
+                    return
                 }
                 bottomStackView.layoutIfNeeded()
             }
@@ -414,6 +427,9 @@ open class MessageInputBar: UIView {
         performLayout(animated) {
             self.leftStackViewWidthContant = newValue
             self.layoutStackViews([.left])
+            guard let _ = self.superview else {
+                return
+            }
             self.layoutIfNeeded()
         }
     }
@@ -427,6 +443,9 @@ open class MessageInputBar: UIView {
         performLayout(animated) {
             self.rightStackViewWidthContant = newValue
             self.layoutStackViews([.right])
+            guard let _ = self.superview else {
+                return
+            }
             self.layoutIfNeeded()
         }
     }
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -443,6 +443,8 @@ open class MessageInputBar: UIView {
     /// - Parameter positions: The UIStackView's to layout
     public func layoutStackViews(_ positions: [InputStackView.Position] = [.left, .right, .bottom, .top]) {
         
+        guard superview != nil else { return }
+        
         for position in positions {
             switch position {
             case .left:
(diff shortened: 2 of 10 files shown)


Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -26,7 +26,7 @@ import Foundation
 
 /// A protocol used by the `MessagesCollectionViewFlowLayout` object to determine
 /// the size and layout of a `MessageCollectionViewCell` and its contents.
-public protocol MessagesLayoutDelegate: class {
+public protocol MessagesLayoutDelegate: AnyObject {
```
SD10, line 29: @zhongwuzw Why did we change these from `class` -> `AnyObject`? I think a `class` meets the minimum requirement? The reason it's constrained to classes is to have a weak reference

zhongwuzw: @SD10 , yeap, class-only protocol keyword has changed from `class` to `AnyObject` in `Swift4`. [Class-Only Protocols](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Protocols.html#//apple_ref/doc/uid/TP40014097-CH25-ID281)

Sources/Views/MessageInputBar.swift
```diff
@@ -319,6 +319,10 @@ open class MessageInputBar: UIView {
     /// - Parameter positions: The UIStackView's to layout
     public func layoutStackViews(_ positions: [InputStackView.Position] = [.left, .right, .bottom]) {
         
+        guard let _ = superview else {
+            return
```
nathantannar4: Why are we adding this check ?

Moreover why are we checking this multiple times ? That’s redundant

zhongwuzw: @nathantannar4 , because we should not call `layoutIfNeeded` before `MessageInputBar` be added to the `superview`. If we do this , `Auto Layout` System would use `_UITemporaryLayoutWidth/_UITemporaryLayoutWidth` to break our custom constraints. we can reproduce it when `MessageInputBar` become first responder, and then switch the keyboard .

Log ex.

```
[LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x60400029f630 V:|-(6)-[MessageKit.InputTextView:0x7fe2a20a0200]   (active, names: '|':MessageKit.MessageInputBar:0x7fe2a142b590 )>",
    "<NSLayoutConstraint:0x604000480b40 MessageKit.InputTextView:0x7fe2a20a0200.bottom == MessageKit.InputStackView:0x7fe2a1432e10.top   (active)>",
    "<NSLayoutConstraint:0x604000284fb0 MessageKit.InputStackView:0x7fe2a1432e10.bottom == UILayoutGuide:0x6040007b2c20'UIViewSafeAreaLayoutGuide'.bottom - 6   (active)>",
    "<NSLayoutConstraint:0x60000029e690 '_UITemporaryLayoutHeight' MessageKit.MessageInputBar:0x7fe2a142b590.height == 0   (active)>",
    "<NSLayoutConstraint:0x60400029c110 'UIViewSafeAreaLayoutGuide-bottom' V:[UILayoutGuide:0x6040007b2c20'UIViewSafeAreaLayoutGuide']-(0)-|   (active, names: '|':MessageKit.MessageInputBar:0x7fe2a142b590 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x604000284fb0 MessageKit.InputStackView:0x7fe2a1432e10.bottom == UILayoutGuide:0x6040007b2c20'UIViewSafeAreaLayoutGuide'.bottom - 6   (active)>
```

SD10 (approved):
I'm good with these changes if you are @nathantannar4

Sources/Views/MessageInputBar.swift
```diff
@@ -378,6 +382,9 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     leftStackView.addArrangedSubview($0)
                 }
+                guard let _ = superview else {
```
nathantannar4: We don’t need to check got the superview again

Sources/Views/MessageInputBar.swift
nathantannar4: Hmm but this error shouldn’t happen when you don’t animate the layout. Going to check to make sure this doesn’t break anything tomorrow. In any case I would prefer a 
‘guard superview != nil else { return }’

nathantannar4 (changes requested):
Ok layout still works so I am ok with this but please remove the redundant checking of the superview. We only need to check it once at the top.

Please also make it `guard superview != nil else { return }`

Sources/Views/MessageInputBar.swift
zhongwuzw: @nathantannar4 , nope, I think it's needed because It can have no superview when run into `setNewItems` method. you can reproduce it when try to comment the `guard statements` .

zhongwuzw (commented):
@nathantannar4 , nope, I think it's needed because It can have no superview when run into `setNewItems` method. you can reproduce it when try to comment the `guard statements` .

Sources/Views/MessageInputBar.swift
```diff
@@ -510,9 +509,7 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     leftStackView.addArrangedSubview($0)
                 }
-                guard let _ = superview else {
-                    return
-                }
+                guard superview != nil else { return }
```
nathantannar4: This is not needed

Sources/Views/MessageInputBar.swift
```diff
@@ -522,9 +519,7 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     rightStackView.addArrangedSubview($0)
                 }
-                guard let _ = superview else {
-                    return
-                }
+                guard superview != nil else { return }
```
nathantannar4: This is not needed

Sources/Views/MessageInputBar.swift
```diff
@@ -534,9 +529,7 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     bottomStackView.addArrangedSubview($0)
                 }
-                guard let _ = superview else {
-                    return
-                }
+                guard superview != nil else { return }
```
nathantannar4: This is not needed

Sources/Views/MessageInputBar.swift
```diff
@@ -546,6 +539,7 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     topStackView.addArrangedSubview($0)
                 }
+                guard superview != nil else { return }
```
nathantannar4, line 542: This is not needed

Sources/Views/MessageInputBar.swift
```diff
@@ -443,6 +443,8 @@ open class MessageInputBar: UIView {
     /// - Parameter positions: The UIStackView's to layout
     public func layoutStackViews(_ positions: [InputStackView.Position] = [.left, .right, .bottom, .top]) {
         
+        guard superview != nil else { return }
+
```
nathantannar4, line 447: We already check the superview at the top of the function so we don't need to check again

nathantannar4 (changes requested):
Again in `public func layoutStackViews` we don't need to check the superview in each switch statement when we check it at the top 😊

Sources/Views/MessageInputBar.swift
zhongwuzw: @nathantannar4 , sorry , my mistake the comments I make before. I would update the code.

zhongwuzw: @nathantannar4 , Hmm AutoLayout issues still existed when remove redundant `superview` check as you say. 

we can reproduce it when MessageInputBar become first responder, and then switch the keyboard.

ps. I reverted the commits first.

```
Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKit/UIView.h> may also be helpful.
2017-11-06 14:55:39.752989+0800 ChatExample[3284:1031668] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x600000298d30 MessageKit.InputTextView:0x7fc9d1137200.bottom == MessageKit.InputStackView:0x7fc9d056d410.top   (active)>",
    "<NSLayoutConstraint:0x60000028b680 MessageKit.InputStackView:0x7fc9d056d410.bottom == UILayoutGuide:0x6000007aa8e0'UIViewSafeAreaLayoutGuide'.bottom - 6   (active)>",
    "<NSLayoutConstraint:0x600000289c40 V:|-(0)-[MessageKit.InputStackView:0x7fc9d058ab60]   (active, names: '|':MessageKit.MessageInputBar:0x7fc9d0565f70 )>",
    "<NSLayoutConstraint:0x600000288340 MessageKit.InputStackView:0x7fc9d058ab60.bottom == MessageKit.InputTextView:0x7fc9d1137200.top - 6   (active)>",
    "<NSLayoutConstraint:0x600000288480 '_UITemporaryLayoutHeight' MessageKit.MessageInputBar:0x7fc9d0565f70.height == 0   (active)>",
    "<NSLayoutConstraint:0x60000028af50 'UIViewSafeAreaLayoutGuide-bottom' V:[UILayoutGuide:0x6000007aa8e0'UIViewSafeAreaLayoutGuide']-(0)-|   (active, names: '|':MessageKit.MessageInputBar:0x7fc9d0565f70 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x60000028b680 MessageKit.InputStackView:0x7fc9d056d410.bottom == UILayoutGuide:0x6000007aa8e0'UIViewSafeAreaLayoutGuide'.bottom - 6   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKit/UIView.h> may also be helpful.
```

nathantannar4: None of these should happen when using the API correctly, even without the superview check. Are you animating? All of the examples in the Example project do not get these issues so I am curious as to what you are doing in the code? You should not be making the input bar the first responder manually. iOS takes care of that because it’s an input access view

zhongwuzw: @nathantannar4 
Reproduce steps:
messageinputbarbug

SD10:
@nathantannar4 Do you see how the `MessageInputBar` disappears and reappears when its dismissed in @zhongwuzw's GIF? We need to look into that too, I can reproduce in ChatExample app

nathantannar4:
@SD10 thats because in the example a `MessageInputBar` is destroyed and then reinitialized. Then `reloadInputViews()` is called

Sources/Views/MessageInputBar.swift
nathantannar4: @zhongwuzw I don't get the same errors under the latest master branch. Did you adjust anything in the ConversationViewController? I am sorry for being picky

zhongwuzw: @nathantannar4 , nope, I downloaded project from `master` directly without any modification, it can reproduce in Simulator or Device.

zhongwuzw:
@SD10 @nathantannar4 , I make a [demo](https://github.com/zhongwuzw/MessageKit/tree/FixKeyboardReappearBug) to fix keyboard reappear issue. 

The reason is system would restore the state when dismiss.

demo

SD10:
@nathantannar4 I can reproduce these constraint errors. I'd like to get this merged into the next patch release

nathantannar4 (approved):

nathantannar4:
Ok I am good with them

PR #329, opened 2017-11-07
Fix messagesCollection constraints bug on iOS9

zhongwuzw:


zhongwuzw:
`messagesCollectionView` not set up top constraints correctly on `iOS9`.
collectionviewinsetsbug

SD10:
Thanks @zhongwuzw I’ll take a look at this. I have to go through iOS 9 - 11 & iPhone 6 - X and make sure they’re all working correctly. I know we currently have some inconsistencies 

nathantannar4:
@SD10 this is the same issue we knew about with the `automaticallyAdjustContentInset` being treated differently on iOS 11 and < iOS 11

nathantannar4 (commented):
I am not sure if this is what we want as we lose the ability to have the messages appear under a transparent navigation bar

SD10:
No we don't @nathantannar4 we have to find code that works for:
- iOS 9 to 11
- iPhone SE, iPhone 6 - 8, iPhone 6 Plus - iPhone 8 Plus, and iPhone X
- Supports both translucent and opaque nav bars
- `automaticallyAdjustContentInset` should be `false` because we modify the insets ourselves

nathantannar4:
@SD10 I think I have a solution.

nathantannar4:
PR #334

zhongwuzw:
@SD10 , iOS11 added a new API `contentInsetAdjustmentBehavior` can modify the insets . I think we should also set it to `.never`.

nathantannar4:
@zhongwuzw That wouldn't help on devices that are lower than iOS 11 (where the problem is occurring) because the API wouldn't be available

zhongwuzw:
@nathantannar4 , yep, my point is we should set `contentInsetAdjustmentBehavior` to `never` on iOS11, because we will manually adjust contentInset .

@SD10 , I updated the code, it can support the changes of navigationBar, like `hidesBarsOnTap`、`hidesBarsOnSwipe `、`setNavigationBarHidden`. By the way, we should use `setNeedsLayout`、`layoutIfNeeded` when `setNavigationBarHidden` at runtime.

nathantannar4:
Why did you remove the iPhone X inset calculation? That was important for allowing transparency behind the MessageInputBar and having the bottom inset correctly calculated

SD10:
I couldn't look at this yesterday because I got stuck downloading a whole bunch of old iOS simulators 😅 It's going to be painful to get this right across all of our design constraints but I will review all these proposed solutions throughout the week

zhongwuzw:
@nathantannar4 , my fix not work on iPhoneX correctly?

nathantannar4:
We opted to go with #334 as it allows for transparency on the `MessageInputBar` and `UINavigationBar`

zhongwuzw:
@SD10 ,@nathantannar4 , my fix also support transparency I think. But whatever. I think we also needs to set `contentInsetAdjustmentBehavior` to `never` in `iOS11` even it's maybe look good on `iOS11`,the reason based:
1.  It would adjust contentInset , maybe even to `automaticallyAdjustsScrollViewInsets`.
2. Maybe leads to weird behavior? In my own project, I used `UIWebView` 、`UITableView`, I need to adjust contentinsets manually, with not set `contentInsetAdjustmentBehavior` to `never` in `iOS11`, It would break the insets . Also I found some issues found in stackoverflow, like [https://stackoverflow.com/questions/45374136/uicollectionview-remove-whitespace-header-ios11/46107329#46107329](https://stackoverflow.com/questions/45374136/uicollectionview-remove-whitespace-header-ios11/46107329#46107329) 
[https://stackoverflow.com/questions/45314286/ios-11-beta-uicollectionviewcell-resizes-as-it-approaches-top-bar-using-contenti](https://stackoverflow.com/questions/45314286/ios-11-beta-uicollectionviewcell-resizes-as-it-approaches-top-bar-using-contenti)
[https://stackoverflow.com/search?tab=votes&q=contentInsetAdjustmentBehavior](https://stackoverflow.com/search?tab=votes&q=contentInsetAdjustmentBehavior)

PR #334, opened 2017-11-07
Adjust top contentInset on iOS < 11

nathantannar4:
Fixes the top contentInset issues found in iOS versions earlier than 11.

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -66,6 +66,7 @@ open class MessagesViewController: UIViewController {
     open override func viewDidLoad() {
         super.viewDidLoad()
 
+        
         extendedLayoutIncludesOpaqueBars = true
         automaticallyAdjustsScrollViewInsets = false
         view.backgroundColor = .white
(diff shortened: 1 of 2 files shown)


Sources/Controllers/MessagesViewController.swift
```diff
@@ -141,6 +141,20 @@ open class MessagesViewController: UIViewController {
             let trailing = messagesCollectionView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
             NSLayoutConstraint.activate([top, bottom, trailing, leading])
         }
+        adjustScrollViewInset()
+    }
+    
+    @objc
+    private func adjustScrollViewInset() {
+        if #available(iOS 11.0, *) {
+            // No need to add a to the top contentInset
```
BasThomas: Nit: this sentence doesn't make sense. Change it to `no need to add to the top contentInset` maybe?

Sources/Controllers/MessagesViewController.swift
```diff
@@ -141,6 +141,20 @@ open class MessagesViewController: UIViewController {
...
+        } else {
+            let navigationBarInset = navigationController?.navigationBar.frame.height ?? 0
+            let statusBarInset: CGFloat = view.bounds.height > view.bounds.width ? 20 : 0 // No status bar in landscape
```
BasThomas: Would this work on iPad?

nathantannar4: Hmm good point

Sources/Controllers/MessagesViewController.swift
```diff
@@ -141,6 +142,20 @@ open class MessagesViewController: UIViewController {
             let trailing = messagesCollectionView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
             NSLayoutConstraint.activate([top, bottom, trailing, leading])
         }
+        adjustScrollViewInset()
+    }
+    
+    @objc
+    private func adjustScrollViewInset() {
+        if #available(iOS 11.0, *) {
```
SD10, line 150: I guess we can't `!` this statement?

Nonetheless, this PR looks promising. I'd just like to test it against iOS 9-11 and the iPhone X. Also both opaque and translucent navigation bars

nathantannar4: Ya you can't ! and `#available` because I think `#available` is a compiler operation (it swift is similar to C/C++

SD10 (approved):
Thank you @nathantannar4 🥇

PR #335, opened 2017-11-07
Refined Example App Icon

nathantannar4:
This is more of a branding thing to keep consistency


SD10:
We should be using the MessageKit organization logo for this! App Icons shouldn't have names/words in them. I think we can also update the MessageKit banner to use the same color that we use in the organization logo? I think you mentioned something about that last week

nathantannar4:
simulator screen shot - iphone x - 2017-11-07 at 12 29 27

SD10 (approved):
Thanks @nathantannar4 💯

PR #337, opened 2017-11-08
Change from closed to half-open range operator

zhongwuzw:
Oops….


SD10 (approved):
I saw that and was wondering where my off by 1 error was 😅

zhongwuzw:
@SD10 :beers::beers::beers::beers:

Issue #302, opened 2017-10-30
Does not scroll to bottom when a message posted from input bar contains 2 lines or more


azhararmar:
## General Information

* MessageKit Version: 0.1.0

* iOS Version(s): 10.3.2

* Swift Version: 4

* Devices/Simulators: iPhone6 Device & Simulator

* Reproducible in ChatExample? (Yes/No): Yes

##  What happened?
If a message entered from input bar has 2 or more lines, the messages does not scroll to bottom even after `scrollsToBottomOnKeybordBeginsEditing` is set to true

Please note that `func messageInputBar(_ inputBar: MessageInputBar, didPressSendButtonWith text: String) ` already has `messagesCollectionView.reloadData()` and `self.messagesCollectionView.scrollToBottom()` implemented.

## What did you expect to happen?
The message collection must scroll to bottom each time a new message is posted by user.

nathantannar4:
@azhararmar I looked into this and I do not believe it to be a bug. The flag `scrollsToBottomOnKeybordBeginsEditing` is for automatically scrolling the view to the bottom when the `InputTextView` becomes the first responder. That is, the keyboard begins editing as the flag states. It is not a flag that scrolls the view whenever the send button is pressed. That is left for you to implement via `messagesCollectionView.scrollToBottom()`

This is how it is done in the example app. I believe you misunderstood the flag, `scrollsToBottomOnKeybordBeginsEditing`

SD10:
You're both correct here. @azhararmar You have misunderstood the flag, it is as @nathantannar4 described, but you're correct about the `MessagesCollectionView` not scrolling to the bottom for messages with 2 lines or more. 

#301 Looks like the solution to this. Also discussed in #268 && #336

azhararmar:
Sorry for delayed response. I was held up with few stuff.

@nathantannar4 you are right in your explanation. However I did implement `messagesCollectionView.scrollToBottom()` when send button is pressed. @SD10 thanks for clarification and for the fix of-course :-)

SD10:
Being worked on in #395

SD10:
Will be fixed in 0.12.0 by #395

PR #340, opened 2017-11-08
Fix to scroll when the height of the keyboard has changed

### LINKED ISSUE #338: Does not scroll to bottom when the height of the keyboard has changed
## General Information

* MessageKit Version:
0.10.1
* iOS Version(s):
iOS 11.1
* Swift Version:
Swift 4.0
* Devices/Simulators:
iPhone 8 (Simulators)
yes

##  What happened?
messageCollectionView doesn't scroll to bottom when the height of the keyboard has changed.

message-kit

I think that it is better to scroll `messageCollectionView` to bottom when the height of the keyboard has changed like messenger, iMessage and Slack.

hellokenta1024:
## What does this implement/fix? Explain your changes.

Hi. I want to try to resolve #338.
It's a simple improvement.
I just called `messagesCollectionView.scrollToBottom(animated: true)` in `handleKeyboardDidChangeState(_ notification: Notification)` when software keyboard has been found.

Please let me know if you need a flag like `scrollsToBottomOnKeyboardDidChangeState` etc.
Thanks

after-masssage-kit

## Does this close any currently open issues?

resolve #338 

## Where has this been tested?

**Devices/Simulators:** …
iPhone 8 (Simulator)

**iOS Version:** …
iOS 11.1

**Swift Version:** …
Swift 4.0

**MessageKit Version:** …
1.10.1


SD10:
Hey @KEN-chan thanks for taking the time to submit this PR 👍 However, I'm not sure if this is the right approach for this even with a `scrollsToBottomOnKeyboardDidChangeState` flag.

Users may not want to scroll to the bottom on events like dismissing the keyboard or when the keyboard height changes. For example, a user could want to look at a previous message for reference when typing the current message.

I think this might work better by observing the `textViewDidChange` notification. 
What do you think? cc @cwalo @nathantannar4 @zhongwuzw

nathantannar4:
A user could also use the `MessageInputBarDelegte` which can notify when the content size changes. A user could then call scrollToBottom. So I don’t think we need another flag

hellokenta1024:
@SD10 @nathantannar4 
Thanks for your comments.
Indeed, `scrollToBottom` is not good.
How about scrolling the screen as much as the height of the keyboard changes?

I've fixed code. Could you check it??

fixed-message-kit

SD10:
@KEN-chan Do you think we can add a flag for this so that the user can opt in to this behavior?

nathantannar4:
@KEN-chan I notice in the gif that if the content offset is not at the max y (i.e., the user has scrolled up a bit) the rest of the messages still moves. Could you maybe add a check such that it only changes the contentOffset if the current offset is at the very bottom? @SD10 do you agree?

SD10:
@nathantannar4 Hmmm, I’m not sure but I kind of like that it maintains the current contentOffset in the collection view. Any specific reason why you think we should constrain it to the bottom? 

hellokenta1024:
@SD10 
> Do you think we can add a flag for this so that the user can opt in to this behavior?

Sure. 
I've added `scrollsOnKeyboardFrameChanges`.

nathantannar4:
@SD10 I just thought the adjustment when in mid scroll looked odd and that the purpose of scrolling to the bottom would be to keep the bottom most message in the frame

hellokenta1024:
@SD10 @nathantannar4 
I think it is natural because some major message app (Slack, FB Messenger) have this same behavior.
In the case of iMessage, it always scrolls to the bottom when the height of the keyboard changes.
What do  you think??

SD10:
@KEN-chan I really like this functionality but I'm blocking merging it at the moment until I can give it a more in-depth review

hellokenta1024:
@SD10 
I'm sorry to hurry you, could you review this when you have time?? 🙇

SD10:
@KEN-chan This will be merged (I may have to change the name of the flag) but functionality seems good. I'm just trying to fix #301, #302 first and it's giving me some trouble

Sources/Controllers/MessagesViewController.swift
```diff
@@ -45,6 +45,12 @@ open class MessagesViewController: UIViewController {
     ///
     /// The default value of this property is `false`.
     open var scrollsToBottomOnKeybordBeginsEditing: Bool = false
+    
+    /// A Boolean value that determines whether the `MessagesCollectionView` scrolls
+    /// according to the change of the height of the keyboard.
+    ///
+    /// The default value of this property is `false`.
+    open var scrollsOnKeyboardFrameChanges: Bool = false
```
SD10: Let's change `scrollsOnKeyboardFrameChanges` to `maintainPositionOnKeyboardFrameChange`

Sources/Controllers/MessagesViewController.swift
```diff
@@ -281,17 +292,19 @@ fileprivate extension MessagesViewController {
...
-            messagesCollectionView.contentInset.bottom = bottomInset
-            messagesCollectionView.scrollIndicatorInsets.bottom = bottomInset
+            let afterBottomInset = keyboardEndFrame.height > keyboardOffsetFrame.height ? (keyboardEndFrame.height - iPhoneXBottomInset) : keyboardOffsetFrame.height
+            let differenceOfBottomInset = afterBottomInset - messageCollectionViewBottomInset
+            let contentOffset = CGPoint(x: messagesCollectionView.contentOffset.x, y: messagesCollectionView.contentOffset.y + differenceOfBottomInset)
+            
+            messageCollectionViewBottomInset = afterBottomInset
+            
+            if scrollsOnKeyboardFrameChanges {
+                messagesCollectionView.setContentOffset(contentOffset, animated: true)
```
SD10: Let's set `animated` to `false` here

SD10 (changes requested):
@KEN-chan Sorry this took so long to review. We can move forward with this PR. If you could take care of the 2 changes I requested. I'll take care of the CHANGELOG entry for you 👍

Sources/Controllers/MessagesViewController.swift
```diff
@@ -341,8 +336,8 @@ fileprivate extension MessagesViewController {
             
             messageCollectionViewBottomInset = afterBottomInset
             
-            if scrollsOnKeyboardFrameChanges {
-                messagesCollectionView.setContentOffset(contentOffset, animated: true)
+            if maintainPositionOnKeyboardFrameChange {
```
zhongwuzw: @SD10 Change from `maintainPositionOnKeyboardFrameChange ` to `maintainPositionOnKeyboardFrameChanged` would be better?

SD10: Sure, I'll change it when I add the CHANGELOG entry after I merge 👍

SD10:

PR #342, opened 2017-11-09
Fix crash for escaping block

zhongwuzw:
Crash sometimes when switch keyboard.


nathantannar4 (approved):

SD10 (approved):
Thanks again @zhongwuzw 💯

PR #344, opened 2017-11-11
Fix inputTextView calculation bug

zhongwuzw:
messageinputbarconstraintsbug


nathantannar4 (commented):
It looks like you are using an outdated version of `MessageInputBar`. I know this because we calculate the size in `calculateIntrinsicContentSize()` and this is not an issue in the latest release

nathantannar4:
Sorry correction, its fixed in the `development` branch

PR #347, opened 2017-11-12
Bugfix/separator line

nathantannar4:
I don't think we can put the SeparatorLine in the topStackView as we use the safeAreaLayoutGuide for the topStackView. This results in this:
simulator screen shot - iphone x - 2017-11-12 at 12 48 00

Switching back to the way its currently done in the `master` branch


SD10:
Cool, I don't think this needs a CHANGELOG entry because having the `SeparatorLine` in the top stack view has never made it to `master` yet.

PR #348, opened 2017-11-13
Fix MessageInputBar translucent property

zhongwuzw:
1. Change `backgroundView.backgroundColor` from `.white` to `color.withAlphaComponent(1.0)` because user can change `backgroundColor`.
2. Add `MessageInputBar` `backgroundColor` to support `translucent` feature .

### DIFF
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -41,16 +41,9 @@ open class MessageInputBar: UIView {
     open var backgroundView: UIView = {
         let view = UIView()
         view.translatesAutoresizingMaskIntoConstraints = false
-        view.backgroundColor = .white
+        view.backgroundColor = .inputBarGray
         return view
     }()
-    
-    /// Also sets the backgroundView's backgroundColor to the newValue
-    open override var backgroundColor: UIColor? {
-        didSet {
-            backgroundView.backgroundColor = backgroundColor
-        }
-    }
     
     /**
      A UIVisualEffectView that adds a blur effect to make the view appear transparent.
@@ -73,11 +66,8 @@ open class MessageInputBar: UIView {
                 blurView.fillSuperview()
             }
             blurView.isHidden = !isTranslucent
-            let color: UIColor = backgroundView.backgroundColor ?? .white
+            let color: UIColor = backgroundView.backgroundColor ?? .inputBarGray
             backgroundView.backgroundColor = isTranslucent ? color.withAlphaComponent(0.75) : color.withAlphaComponent(1.0)
-            
-            let bgColor: UIColor = backgroundColor ?? .inputBarGray
-            backgroundColor = isTranslucent ? bgColor.withAlphaComponent(0.75) : bgColor.withAlphaComponent(1.0)
         }
     }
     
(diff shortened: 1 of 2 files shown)


Sources/Views/MessageInputBar.swift
```diff
@@ -74,7 +74,10 @@ open class MessageInputBar: UIView {
             }
             blurView.isHidden = !isTranslucent
             let color: UIColor = backgroundView.backgroundColor ?? .white
-            backgroundView.backgroundColor = isTranslucent ? color.withAlphaComponent(0.75) : .white
+            backgroundView.backgroundColor = isTranslucent ? color.withAlphaComponent(0.75) : color.withAlphaComponent(1.0)
```
SD10: 👍

nathantannar4: 👍

nathantannar4 (approved):

Sources/Views/MessageInputBar.swift
```diff
@@ -74,7 +74,10 @@ open class MessageInputBar: UIView {
...
+            
+            let bgColor: UIColor = backgroundColor ?? .inputBarGray
+            backgroundColor = isTranslucent ? bgColor.withAlphaComponent(0.75) : bgColor.withAlphaComponent(1.0)
         }
```
nathantannar4: We can't have this. If you run this you will see it breaks the translucent functionality. We already have:
```swift
/// Also sets the backgroundView's backgroundColor to the newValue
    open override var backgroundColor: UIColor? {
        didSet {
            backgroundView.backgroundColor = backgroundColor
        }
    }
```

nathantannar4 (changes requested):

Sources/Views/MessageInputBar.swift
nathantannar4: Noticed some older inconsistencies, can you actually change this to:
```
let color: UIColor = backgroundView.backgroundColor ?? .inputBarGray
            backgroundColor = .clear
            backgroundView.backgroundColor = isTranslucent ? color.withAlphaComponent(0.75) : color
```
I wanted to make it such that the topStackView can have a different background color which is why we only want to adjust the backgroundView's color

zhongwuzw: @nathantannar4 ,In your opinion , backgroundView's backgroundColor is not equal to MessageInputBar's backgroundColor, we should delete the property observer of backgroundColor?

```
    /// Delete the property observer below ???
    /// Also sets the backgroundView's backgroundColor to the newValue
    open override var backgroundColor: UIColor? {
        didSet {
            backgroundView.backgroundColor = backgroundColor
        }
    }
```

zhongwuzw: @nathantannar4 , as your suggested code, default MessageInputBar's backgroundColor is .inputBarGray, when first set translucent to true, then set to false, MessageInputBar's backgroundColor would always set to .clear, it violate the original setting.

```
let color: UIColor = backgroundView.backgroundColor ?? .inputBarGray
            backgroundColor = .clear
            backgroundView.backgroundColor = isTranslucent ? color.withAlphaComponent(0.75) : color
```

nathantannar4 (approved):

nathantannar4:
Awesome wanna just add a changelog

SD10 (approved):
Thanks for the great work @zhongwuzw @nathantannar4 💯

Issue #349, opened 2017-11-13
Keyboard dismissal infinite loop on the iPhone X 


mlequeux:
## General Information

* MessageKit Version:

v0.10.2

* iOS Version(s):

11.1

* Swift Version:

4

* Devices/Simulators:

iPhone X, iPhone X Simulator

* Reproducible in ChatExample? (Yes/No):

Yes

##  What happened?

I think I found a severe bug that can lead to an infinite loop with memory leaking like crazy, it seems I'm only able to trigger it on the new iPhone X, has anyone got his hands on the phone with the ChatExample app.

To trigger the bug:

- Open MessageKit ChatExample app
- Open Example tab
- Open Keyboard
- Dismiss Keyboard by swiping down on the chat (And keep swiping down, slowly, all the way until the keyboard disappears)

Then your RAM goes up a lot and the app is unresponsive,

The culprit is the inputBar :

- UIKeyboardWillChangeFrame is called repeatedly after the freeze
- Same goes with the intrinsicContentSize of the MessageInputBar

## What did you expect to happen?

Keyboard dismisses properly.

nathantannar4:
Due to this I think we need to go with the more common solution with `didMoveToWindow()`

```swift
open override func didMoveToWindow() {
        super.didMoveToWindow()
        if #available(iOS 11.0, *) {
            guard let window = window else { return }
            bottomStackViewLayoutSet?.bottom?.isActive = false
            bottomStackViewLayoutSet?.bottom = bottomStackView.bottomAnchor.constraintLessThanOrEqualToSystemSpacingBelow(window.safeAreaLayoutGuide.bottomAnchor, multiplier: 1)
            bottomStackViewLayoutSet?.bottom?.isActive = true
        }
    }
```

jslinker:
I've been experimenting with a similar control in my own project. If an `inputAccessoryView` returns anything other than the default height for `intrinsicContentSize`, you get a loop. The stack trace makes it look like it's conflicting with the `intrinsicContentSize` of the `UITextField` or `UITextView` it contains.

nathantannar4:
Will close this as we have a PR that can fix it

PR #350, opened 2017-11-13
MessageInputBar bottomAnchot Fix

nathantannar4:
Fix to issue #349


jslinker:
I just tested this on iPhone X simulator. Seems to fix the infinite loop issue.

SD10 (approved):
Thank you @nathantannar4 🍻

PR #361, opened 2017-11-26
Bottom anchor size

nathantannar4:
Fixes some constraint and sizing issues. Adds size caching so the getter for intrinsicContentSize is faster/

### DIFF
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -190,13 +190,9 @@ open class MessageInputBar: UIView {
         }
     }
     
+    /// Returns the most recent size calculated by `calculateIntrinsicContentSize()`
     open override var intrinsicContentSize: CGSize {
-        let size = calculateIntrinsicContentSize()
-        if previousIntrinsicContentSize != size {
-            delegate?.messageInputBar(self, didChangeIntrinsicContentTo: size)
-            previousIntrinsicContentSize = size
-        }
-        return size
+        return cachedIntrinsicContentSize
     }
     
     /// The intrinsicContentSize can change a lot so the delegate method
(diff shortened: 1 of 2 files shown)


Sources/Views/MessageInputBar.swift
```diff
@@ -190,28 +190,27 @@ open class MessageInputBar: UIView {
...
+    /// The most recent calculation of the intrinsicContentSize
+    private lazy var cachedIntrinsicContentSize: CGSize = calculateIntrinsicContentSize()
+    
     /// A boolean that indicates if the maxTextViewHeight has been met. Keeping track of this
     /// improves the performance
     public private(set) var isOverMaxTextViewHeight = false
     
     /// The maximum height that the InputTextView can reach
-    open var maxHeight: CGFloat = UIScreen.main.bounds.height / 3 {
+    open var maxTextViewHeight: CGFloat = UIScreen.main.bounds.height / 3 {
```
SD10, line 211: This is an API breaking change

Sources/Views/MessageInputBar.swift
```diff
@@ -437,7 +447,10 @@ open class MessageInputBar: UIView {
                 inputTextView.invalidateIntrinsicContentSize()
             }
         }
-        return CGSize(width: bounds.width, height: heightToFit)
+        let totalPadding = padding.top + padding.bottom + topStackViewPadding.top + textViewPadding.top + textViewPadding.bottom
```
SD10, line 450: Do we need to consider `topStackViewPadding.bottom`? 

What about the other stack views? 
Can something like this happen where the left or right stackviews increases the size?
```
[            Top            ]
| L |                       
|   |    [TextView]     [ R ]
[          Bottom           ]
```

nathantannar4: Ya i wanna screenhero about this when you have a moment

nathantannar4: If extra height is needed it will be given because the `MessageInputBar` autoresizingMask = flexibleHeight

PR #364, opened 2017-11-30
Add shouldCacheLayoutAttributes method

### LINKED ISSUE #306: Experimental: Add shouldCacheLayoutAttributes(for message: MessageType) -> Bool
### Reasoning
Some users may have messages that are displayed in cells with dynamic sizes: ie) Link previews.

### Problem Area
MessageKit caches the layout for messages after the first calculation by using the message ID. This means if a message is laid out with size A the first time it appears and then has size B the second time it appears, size A will be used. This requires the users to manually invalidate the cache for this message.

### Approach
I'm thinking if we add a `shouldCacheLayoutAttributes(for message: MessageType) -> Bool` method. The user can specify specific messages that will be used in cells with dynamic sizing, thus not having to worry about invalidating the cache.

cc @cwalo

SD10:
Resolves #306. We no longer do any caching by default. This will resolve a lot of issues people are experiencing and eliminate the need to use the `(messagesCollectionView.collectionViewLayout as! MessagesCollectionViewFlowLayout).attributesCacheMaxSize = 0` workaround.

TODO:
- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 `AvatarView`'s vertical and horizontal position in a `MessageCollectionViewCell`.
 [#322](https://github.com/MessageKit/MessageKit/pull/322) by [@SD10](https://github.com/sd10).
 
+- Added `shouldCacheLayoutAttributes(for:MessageType)-> Bool` method to `MessagesLayoutDelegate`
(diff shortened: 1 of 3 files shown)


SD10:
This will fail due to some tests I broke in the last PR. I'm likely going to rewrite a lot of those with Quick/Nimble. I don't really rely on our current tests. I think some behavior tests would give me more piece of mind. I already see I introduced a bug regarding the bottom label size during the refactor.

PR #367, opened 2017-12-01
Optimize DateFormatter to static variable

zhongwuzw:
`DateFormatter` is expensive to create .

### DIFF
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -220,14 +220,21 @@ extension ConversationViewController: MessagesDataSource {
 
     func cellTopLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
         let name = message.sender.displayName
-      return NSAttributedString(string: name, attributes: [NSAttributedStringKey.font: UIFont.preferredFont(forTextStyle: .caption1)])
+        return NSAttributedString(string: name, attributes: [NSAttributedStringKey.font: UIFont.preferredFont(forTextStyle: .caption1)])
     }
 
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
-        let formatter = DateFormatter()
-        formatter.dateStyle = .medium
+        
+        struct ConversationDateFormatter {
+            static let formatter: DateFormatter = {
+                let formatter = DateFormatter()
+                formatter.dateStyle = .medium
+                return formatter
+            }()
+        }
+        let formatter = ConversationDateFormatter.formatter
         let dateString = formatter.string(from: message.sentDate)
-      return NSAttributedString(string: dateString, attributes: [NSAttributedStringKey.font: UIFont.preferredFont(forTextStyle: .caption2)])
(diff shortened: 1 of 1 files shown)


Example/Sources/ConversationViewController.swift
```diff
@@ -220,14 +220,21 @@ extension ConversationViewController: MessagesDataSource {
...
         let name = message.sender.displayName
-      return NSAttributedString(string: name, attributes: [NSAttributedStringKey.font: UIFont.preferredFont(forTextStyle: .caption1)])
+        return NSAttributedString(string: name, attributes: [NSAttributedStringKey.font: UIFont.preferredFont(forTextStyle: .caption1)])
     }
 
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
-        let formatter = DateFormatter()
-        formatter.dateStyle = .medium
+        
+        struct ConversationDateFormatter {
```
SD10, line 228: I have actually never seen a type nested in a function 😱 This persists beyond calls?

zhongwuzw: @SD10 :blush::blush:,It's ok, I just want to narrow the scope of use to `ConversationDateFormatter`, `function` in `Swift` is `type` like others.

> [Enumerations are often created to support a specific class or structure’s functionality. Similarly, it can be convenient to define utility classes and structures purely for use within the context of a more complex type. To accomplish this, Swift enables you to define nested types, whereby you nest supporting enumerations, classes, and structures within the definition of the type they support. ](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/NestedTypes.html)

> [Every function in Swift has a type, consisting of the function’s parameter types and return type. You can use this type like any other type in Swift, which makes it easy to pass functions as parameters to other functions, and to return functions from functions. Functions can also be written within other functions to encapsulate useful functionality within a nested function scope.](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Functions.html)

BasThomas: But why wrap it in a struct if it is not used beyond the function scope? You can just create a variable.

zhongwuzw: @BasThomas , :flushed:`static` variable cannot declaration in `function`.

SD10: Wow this makes sense now... but I still can't wrap my head around whether this struct exists forever or if we are recreating it everytime in the function body

BasThomas: It's static, so yep. Tested [in a gist](https://gist.github.com/BasThomas/9b95956c991e2c2c23c94715fab027f4).

SD10: Dang, this is cool. Maybe overkill but cool. The fact that it's scoped is misleading

SD10 (approved):

Example/Sources/ConversationViewController.swift
zhongwuzw: @SD10 , I think `struct` is exists forever, runtime has get `struct`'s type info in compiling time, and also , it has `static` property, `static` throughout the entire life cycle.

PR #373, opened 2017-12-04
Remove hack code in tests

zhongwuzw:
#368


SD10:
@zhongwuzw I wonder what I was doing wrong to cause this to fail... I swear I tried that already.
Thank you 🙏

SD10 (approved):

PR #379, opened 2017-12-07
Optimize cell gesture handle logic

zhongwuzw:
Remove cell's each subview's gesture, just handle tap through one gesture.

### DIFF
--- a/Sources/Views/Cells/MessageCollectionViewCell.swift
+++ b/Sources/Views/Cells/MessageCollectionViewCell.swift
@@ -59,8 +59,6 @@ open class MessageCollectionViewCell<ContentView: UIView>: UICollectionViewCell,
 
     open weak var delegate: MessageCellDelegate?
 
-    var messageTapGesture: UITapGestureRecognizer?
-
     // MARK: - Initializer
 
     public override init(frame: CGRect) {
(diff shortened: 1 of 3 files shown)


Sources/Views/Cells/MessageCollectionViewCell.swift
```diff
@@ -137,42 +135,37 @@ open class MessageCollectionViewCell<ContentView: UIView>: UICollectionViewCell,
...
+            delegate?.didTapAvatar(in: self)
+        }
+    
     }
-
-  @objc func didTapBottomLabel() {
-        delegate?.didTapBottomLabel(in: self)
+    
+    /// Handle `ContentView`'s tap gesture, return false when `ContentView` don't needs to handle gesture
+    open func cellContentView(canHandle touchPoint: CGPoint) -> Bool {
```
SD10, line 168: @zhongwuzw Can you explain this one to me? Is this an override point for the user?

Sources/Views/Cells/MessageCollectionViewCell.swift
```diff
@@ -137,42 +135,37 @@ open class MessageCollectionViewCell<ContentView: UIView>: UICollectionViewCell,
...
-        delegate?.didTapTopLabel(in: self)
+    
+    /// Handle tap gesture on contentView and its subviews like messageContainerView, cellTopLabel, cellBottomLabel, avatarView ....
+    @objc func handleTapGesture(_ gesture: UIGestureRecognizer) {
+        guard gesture.state == .ended else {
+            return
+        }
+        
+        let touchLocation = gesture.location(in: self)
+        if messageContainerView.frame.contains(touchLocation) && !cellContentView(canHandle: convert(touchLocation, to: messageContentView)) {
```
SD10: Instead of `if else` here, can we use `switch true` with a `default: break`?

Sources/Views/Cells/TextMessageCell.swift
```diff
@@ -84,4 +78,9 @@ open class TextMessageCell: MessageCollectionViewCell<MessageLabel> {
         }
     }
     
+    /// Handle `ContentView`'s tap gesture, return false when `ContentView` don't needs to handle gesture
+    open override func cellContentView(canHandle touchPoint: CGPoint) -> Bool {
+        return messageContentView.handleGesture(touchPoint)
```
SD10, line 83: Oh I think I see now, sorry I didn't make it this far in the diff before 😅

Sources/Views/MessageLabel.swift
```diff
@@ -192,45 +190,6 @@ open class MessageLabel: UILabel, UIGestureRecognizerDelegate {
...
 
-    // MARK: UIGestureRecognizer Delegate
-
-    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
-        return true
-    }
-
-    //swiftlint:disable cyclomatic_complexity
-    // Yeah we're disabling this because the whole file is a mess :D
-    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
```
SD10, line 203: 👏 👏 👏

SD10 (commented):
@zhongwuzw This is some really really good cleanup! It could also make it possible to move forward with `copy/paste` now 😃

SD10 (approved):
This is great 💯 Thanks for awesome contributions @zhongwuzw 👍

PR #385, opened 2017-12-07
Add MessageLabel tests and stubs

SD10:
### Summary:

This PR adds the `MessageLabelSpec` with a few basic tests.
I've also included some empty test stubs for other things I want to test but haven't written yet.
So if anyone wants to pick them up and add implementations feel free.

I should mention that some of the tests may not be possible to write without changing the `MessageLabel` implementation. This is fine. I've already made some minor improvements myself.

I'll probably leave this PR open as I continue to fill out a few more of the test bodies.

If you want to start working on tests, we have two testing dependencies `Quick/Nimble` which are installed with Carthage. You'll need to run `carthage bootstrap` to build the dependencies.


Tests/ControllersTest/MessageLabelSpec.swift
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// MARK: - Helpers
+
+fileprivate extension MessageLabel {
+
+    var textAttributes: [NSAttributedStringKey: Any]? {
+        let length = attributedText!.length
+        var range = NSRange(location: 0, length: length)
+        return attributedText?.attributes(at: 0, effectiveRange: &range)
```
Sherlouk: You treat `attributedText` here as an Optional, but the length value above is calculated with the value being force-unwrapped. This means either the function will crash if it's nil, or doesn't ever need to return an optional if you're force-unwrapping it!

Tests/ControllersTest/MessageLabelSpec.swift
```diff
@@ -0,0 +1,258 @@
...
+            context("address detection is enabled") {
+                it("applies addressAttributes to text") {
+                    let expectedColor = UIColor.blue
+                    messageLabel.addressAttributes = [.foregroundColor: expectedColor]
+                    messageLabel.text = "One Infinite Loop Cupertino, CA 95014"
+                    messageLabel.enabledDetectors = [.address]
+                    let attributes = messageLabel.textAttributes
+                    let textColor = attributes?[.foregroundColor] as? UIColor
+                    expect(textColor).toNot(beNil())
+                    expect(textColor).to(equal(expectedColor))
```
Sherlouk: Is it necessary to expect the value to be equal to something, but also check it's not nil previously? Can remove the nil check!

Tests/ControllersTest/MessageLabelSpec.swift
SD10: Harsh critic 😂 😂

Sherlouk: ❤️

PR #386, opened 2017-12-08
Add a contentView to make managing constraints easier

nathantannar4:
This doesn't change the functionality or how any of the padding works. This just adds a new `contentView` of type UIView to the `MessageInputBar` to hold the main subviews of the `MessageInputBar`. I did this to clean up the way constraints were assigned as they got more complex with the addition of the top `InputStackView`

I also cleaned up some of the documentation.

In short this reduces lines of code and adds more clarity

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - **Breaking Change** Changed the name of  `MessageInputBar`'s property `maxHeight` to `maxTextViewHeight` as the property is the max height the `InputTextView` can have, not the `MessageInputBar` itself.
 [#380](https://github.com/MessageKit/MessageKit/pull/380) by [@nathantannar4](https://github.com/nathantannar4).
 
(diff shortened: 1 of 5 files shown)


SD10:
Your last commit message is a beauty 😂

nathantannar4:
But very accurate!

zhongwuzw:
@nathantannar4 , it crashed when I run. I make the keyboard appear and then change the keyboard style.
5f64b0e9-03b1-48b4-a8a9-bece8d31c4b4

nathantannar4:
@zhongwuzw hmm that doesn't happen for me. iOS version? Simulator or device?

nathantannar4:
Edit: Talked with @zhongwuzw and crash seemed to be a one off, can't be replicated

nathantannar4:
@zhongwuzw I removed some code that I found after moving things to a `contentView` is unneeded. Including the content size invalidation in InputStackView.layoutIfNeeded()

Let me know if you see any more bugs. I think overall this is more stable. There is just that subtle effect on the iPhone X we could work on fixing

PR #387, opened 2017-12-08
Fix notification observer bug when change the keyboard style

zhongwuzw:


SD10:
@zhongwuzw Can you explain this change to me?

zhongwuzw:
@SD10 ,yep, `addKeyboardObservers` only observe initial messageBar, it would invalidate when we change the keyboard style. Because messageBar becomes a new instance not the older one.

SD10 (approved):

PR #391, opened 2017-12-09
Removing cell generic constraints

SD10:
### Summary:
Removes generic constraints on `MessageCollectionViewCell<ContentView: UIView>`

Everything is subject to change, really just want to be able to see the GitHub diff 😅

### DIFF
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -327,6 +327,10 @@ extension ConversationViewController: MessagesDisplayDelegate {
 
 extension ConversationViewController: MessagesLayoutDelegate {
 
+    func enabledDetectors(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> [DetectorType] {
+        return [.url, .address, .phoneNumber, .date]
+    }
+
     func avatarPosition(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> AvatarPosition {
         return AvatarPosition(horizontal: .natural, vertical: .messageBottom)
     }
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -372,19 +372,19 @@ extension ConversationViewController: MessagesLayoutDelegate {
 
 extension ConversationViewController: MessageCellDelegate {
 
-    func didTapAvatar<T>(in cell: MessageCollectionViewCell<T>) {
+    func didTapAvatar(in cell: MessageCollectionViewCell) {
         print("Avatar tapped")
     }
 
-    func didTapMessage<T>(in cell: MessageCollectionViewCell<T>) {
+    func didTapMessage(in cell: MessageCollectionViewCell) {
         print("Message tapped")
     }
 
-    func didTapTopLabel<T>(in cell: MessageCollectionViewCell<T>) {
+    func didTapTopLabel(in cell: MessageCollectionViewCell) {
         print("Top label tapped")
     }
 
-    func didTapBottomLabel<T>(in cell: MessageCollectionViewCell<T>) {
+    func didTapBottomLabel(in cell: MessageCollectionViewCell) {
         print("Bottom label tapped")
     }
 
--- a/Sources/Views/Cells/LocationMessageCell.swift
+++ b/Sources/Views/Cells/LocationMessageCell.swift
@@ -58,7 +58,6 @@ open class LocationMessageCell: MessageCollectionViewCell {
 
         let snapshotOptions = MKMapSnapshotOptions()
         snapshotOptions.region = MKCoordinateRegion(center: location.coordinate, span: options.span)
-        //snapshotOptions.size = imageView.frame.size
         snapshotOptions.showsBuildings = options.showsBuildings
         snapshotOptions.showsPointsOfInterest = options.showsPointsOfInterest
 
--- a/Sources/Views/Cells/LocationMessageCell.swift
+++ b/Sources/Views/Cells/LocationMessageCell.swift
@@ -25,39 +25,30 @@
 import UIKit
 import MapKit
 
-open class LocationMessageCell: MessageCollectionViewCell<UIImageView> {
+open class LocationMessageCell: MessageCollectionViewCell {
     open override class func reuseIdentifier() -> String { return "messagekit.cell.location" }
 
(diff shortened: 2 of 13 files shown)


Example/Sources/ConversationViewController.swift
```diff
@@ -309,16 +309,16 @@ extension ConversationViewController: MessagesDisplayDelegate {
...
+//    func animationBlockForLocation(message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> ((UIImageView) -> Void)? {
+//        return { view in
+//            view.layer.transform = CATransform3DMakeScale(0, 0, 0)
+//            view.alpha = 0.0
+//            UIView.animate(withDuration: 0.6, delay: 0, usingSpringWithDamping: 0.9, initialSpringVelocity: 0, options: [], animations: {
+//                view.layer.transform = CATransform3DIdentity
+//                view.alpha = 1.0
+//            }, completion: nil)
+//        }
+//    }
```
nathantannar4: Is there a reason we want this commented out?

Sources/Views/Cells/MessageBaseCell.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    open func setupSubviews() {
+        contentView.addSubview(messageContentView)
+        contentView.addSubview(avatarView)
+        contentView.addSubview(cellTopLabel)
+        contentView.addSubview(cellBottomLabel)
+    }
+
+    open override func prepareForReuse() {
+        cellTopLabel.text = nil
```
nathantannar4: Do we need to call super here? idk if we do

Sources/Views/Cells/MessageBaseCell.swift
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /// Handle tap gesture on contentView and its subviews like messageContainerView, cellTopLabel, cellBottomLabel, avatarView ....
+    @objc
+    open func handleTapGesture(_ gesture: UIGestureRecognizer) {
+        guard gesture.state == .ended else { return }
+
+        let touchLocation = gesture.location(in: self)
+
+        switch true {
```
nathantannar4: interesting switch statement 😝

Sources/Views/MediaView.swift
```diff
@@ -25,63 +25,74 @@
...
+        addSubview(playButtonView)
+        addSubview(activityIndicator)
+
+        activityIndicator.translatesAutoresizingMaskIntoConstraints = false
+        activityIndicator.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
+        activityIndicator.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
+
+        playButtonView.translatesAutoresizingMaskIntoConstraints = false
+        playButtonView.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
+       playButtonView.centerYAnchor.constraint(equalTo: centerYAnchor).isActive = true
```
nathantannar4: this is off

Sources/Views/MessageContainerView.swift
```diff
@@ -77,4 +75,58 @@ open class MessageContainerView: UIImageView {
...
+    }
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        addSubview(contentView)
+        contentView.translatesAutoresizingMaskIntoConstraints = false
+        contentView.topAnchor.constraint(equalTo: topAnchor).isActive = true
+        contentView.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
+        contentView.leftAnchor.constraint(equalTo: leftAnchor).isActive = true
+        contentView.rightAnchor.constraint(equalTo: rightAnchor).isActive = true
```
nathantannar4: I would use NSLayoutConstraint.activate() its faster

Sources/Views/MediaView.swift
```diff
@@ -25,63 +25,74 @@
...
+        playButtonView.widthAnchor.constraint(equalToConstant: playButtonView.bounds.width).isActive = true
+        playButtonView.heightAnchor.constraint(equalToConstant: playButtonView.bounds.height).isActive = true
```
nathantannar4: I would use NSLayoutConstraint.activate() its faster

Example/Sources/ConversationViewController.swift
SD10: Not sure 😅 It's in progress. I don't think I'm going to be able to remove the subclasses. It doesn't make sense to support multi text messages to be honest. If anything far down the road we can have a subclass for urls/file previews that shows both an image and text

Sources/Views/MessageContainerView.swift
SD10: agreed, should've left a todo. did this for fast refactor

SD10:
Just looking at the diff for the PR shows it wasn't worth the trouble to make it generic in the first place

Sources/Views/Cells/MessageCollectionViewCell.swift
```diff
@@ -59,114 +53,80 @@ open class MessageCollectionViewCell<ContentView: UIView>: UICollectionViewCell,
 
     open weak var delegate: MessageCellDelegate?
 
-    // MARK: - Initializer
-
-    public override init(frame: CGRect) {
+    override public init(frame: CGRect) {
         super.init(frame: frame)
+        contentView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```
zhongwuzw, line 58: Is this `autoresizingMask` necessary? Cell would manage `contentView`'s frame in my opinion.

Sources/Views/Cells/LocationMessageCell.swift
```diff
@@ -66,7 +58,7 @@ open class LocationMessageCell: MessageCollectionViewCell<UIImageView> {
 
         let snapshotOptions = MKMapSnapshotOptions()
         snapshotOptions.region = MKCoordinateRegion(center: location.coordinate, span: options.span)
-        snapshotOptions.size = messageContainerView.frame.size
+        //snapshotOptions.size = imageView.frame.size
```
zhongwuzw: Just use default size?

Sources/Views/Cells/MessageCollectionViewCell.swift
SD10: I'm going to keep it for now because I can't remember why it is here 😅 I think there may have been a small problem on rotation

PR #392, opened 2017-12-09
Adjust messageInputBar's height when traitCollection changed

zhongwuzw:
1. Change orientation observe  to traitCollection.
2. Adjust messageInputBar's height automatically when traitCollection changed.
3. Don't adjust messageInputBar's height automatically when user set `maxTextViewHeight`, wether or not to adjust up to user.

#386

### DIFF
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -460,7 +460,6 @@ open class MessageInputBar: UIView {
         var inputTextViewHeight = requiredInputTextViewHeight
         if inputTextViewHeight >= maxTextViewHeight {
             if !isOverMaxTextViewHeight {
-                textViewHeightAnchor?.constant = maxTextViewHeight
                 textViewHeightAnchor?.isActive = true
                 inputTextView.isScrollEnabled = true
                 isOverMaxTextViewHeight = true
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -205,18 +205,38 @@ open class MessageInputBar: UIView {
     /// improves the performance
     public private(set) var isOverMaxTextViewHeight = false
     
-    /// The maximum height that the InputTextView can reach
-    open var maxTextViewHeight: CGFloat = (UIScreen.main.bounds.height / 5).rounded(.down) {
+    private var isTraitCollectionDidChanged = false {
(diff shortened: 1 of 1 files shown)


Sources/Views/MessageInputBar.swift
```diff
@@ -205,18 +205,38 @@ open class MessageInputBar: UIView {
     /// improves the performance
     public private(set) var isOverMaxTextViewHeight = false
     
-    /// The maximum height that the InputTextView can reach
-    open var maxTextViewHeight: CGFloat = (UIScreen.main.bounds.height / 5).rounded(.down) {
+    private var isTraitCollectionDidChanged = false {
```
nathantannar4: Do we need to be keeping track of this? I don't think we do

Sources/Views/MessageInputBar.swift
```diff
@@ -636,9 +656,12 @@ open class MessageInputBar: UIView {
...
     /// Invalidates the intrinsicContentSize
-    @objc
-    open func orientationDidChange() {
-        invalidateIntrinsicContentSize()
+    open override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        if traitCollection.verticalSizeClass != previousTraitCollection?.verticalSizeClass || traitCollection.horizontalSizeClass != previousTraitCollection?.horizontalSizeClass {
+            isTraitCollectionDidChanged = true
+            invalidateIntrinsicContentSize()
+        }
```
nathantannar4: Whats the benefit from not using a notification?

nathantannar4 (changes requested):
I think this adds to many flags and properties that would make it harder to keep track of. I implemented this fix already in my other project where I just have a simple flag like, `shouldAutoUpdateMaxTextViewHeight` and when true we update the `maxTextViewHeight` on orientation change. I don't think we need all these internal text view heights and flags. See https://github.com/nathantannar4/InputBarAccessoryView/blob/master/InputBarAccessoryView/Views/InputBarAccessoryView.swift

Sources/Views/MessageInputBar.swift
nathantannar4: Talked over slack, trait collection works better on iPad

nathantannar4 (approved):

PR #395, opened 2017-12-11
Fix scrollToBottom && remove scrollsToBottomOnFirstLayout

zhongwuzw:


Sources/Views/MessagesCollectionView.swift
```diff
@@ -65,7 +65,20 @@ open class MessagesCollectionView: UICollectionView {
 
     public func scrollToBottom(animated: Bool = false) {
         guard let indexPath = indexPathForLastItem else { return }
-        scrollToItem(at: indexPath, at: .bottom, animated: animated)
+        
+        let collectionViewContentHeight = collectionViewLayout.collectionViewContentSize.height
+        let isContentTooSmall = (collectionViewContentHeight < bounds.height * 2)
```
SD10, line 70: Why `bounds.height * 2` here?

SD10:
@zhongwuzw This looks good. I fixed the problem with the `MessageLabel` font. 

The only thing I notice about this change is there is a small flicker from using batch updates 🤔

SD10:
@zhongwuzw I also want to mention you don't have to work from a fork of the project, you're a contributor so you can work from branches on the main MessageKit repo 😅

zhongwuzw:
@SD10 
> The only thing I notice about this change is there is a small flicker from using batch updates
I think we can use `scrollToBottom(animated: true)`, the flicker is because of not animated.

Sources/Views/MessagesCollectionView.swift
zhongwuzw: The reason is `scrollToItem` would sometimes invalid when `collectionViewContentSize`'s `height` slightly larger than `collectionView`'s `height`.

SD10: Yes, I understand that, but why did we decide to use `* 2`? Is this just a random number?

zhongwuzw: Nope, it's based on my tests, I found `scrollToItem` sometimes invalid when last cell's maxY greater than bounds's height but less than 2 * bounds's height.

SD10 (changes requested):
@zhongwuzw Can you add a CHANGELOG entry under **Removed** for the `scrollsToBottomOnFirstLayout` flag? And another under **Fixed** for `scrollToBottom()`?

zhongwuzw:
@SD10 , I added the changelog.

SD10 (approved):
Awesome work @zhongwuzw 💯

PR #401, opened 2017-12-13
Fix messageLabel layout issue

zhongwuzw:
1. Remove AutoLayout layout, use only frame.
2. Fix messageLabel `frame` layout, it leads to wrong layout and not show label text added through `messageInputBar`.


SD10:
@zhongwuzw Wouldn't this be an issue for the other cells as well then?

SD10 (approved):

Issue #403, opened 2017-12-14
Miscalculated layout in customized inputView


antoinelamy:
## General Information

* MessageKit Version: `master`

* iOS Version(s): 11.2

* Swift Version: 4.0

* Devices/Simulators: iPhone 8

* Reproducible in ChatExample? (Yes/No): Yes
Just add a call to `slack()` at the end of `viewDidLoad()` and change the `setStackViewItems` call to be not animated.

##  What happened?
Customizing the inputView in viewDidLoad cause the intrinsicContentSize to be miscalculated and the bottom buttons in the exemple slack keyboard are cut in half.

## What did you expect to happen?
The inputView appear correctly when it appear on screen.

## Notes
There is a weird guard condition in `MessageInputBar.setStackViewItems()` that prevent `layoutIfNeeded()` on the bar's stack views if no superView is present. Removing this guard statement fix the problem.

nathantannar4:
Hi @antoinelamy, I took a look at this. The `guard` statements are not useless, they ensure safety. The bug you experience in `master` is due to how the `bottomAnchor` to the `window` is assigned. The issue is not present on `development`

antoinelamy:
@nathantannar4 what safety? I am unaware that an offscreen layout pass is somehow risky. In fact, the initial view controller's view layout pass is always off screen.

antoinelamy:
Do you have an alternative fix?

nathantannar4:
It's already fixed, checkout `development`! :D 

The safety is the offscreen layout that can be dangerous. @zhongwuzw added the `guard` statements

antoinelamy:
I'm glad the issue is already fixed. If you have any documentation about the dangers of offscreen layout I would be interested, never heard of it.

nathantannar4:
Its mainly to protect against those that pass the `animated` parameter as true. You can't `animate` something not on screen

PR #405, opened 2017-12-14
Remove useless guard and fix issue #403

antoinelamy:
Ensure that bar height is calculated correctly even if the setup is made before view appearance.

#403 

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 8 (Simulator)

**iOS Version:** 11.2

**Swift Version:** 4.0

**MessageKit Version:** 0.11

### DIFF
--- a/Sources/Views/MessageInputBar.swift
+++ b/Sources/Views/MessageInputBar.swift
@@ -520,7 +520,6 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     leftStackView.addArrangedSubview($0)
                 }
-                guard superview != nil else { return }
                 leftStackView.layoutIfNeeded()
             case .right:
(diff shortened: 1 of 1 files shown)


nathantannar4 (changes requested):
These guard statements aren't useless. The bug you experienced is already fixed in `development`

Sources/Views/MessageInputBar.swift
```diff
@@ -520,7 +520,6 @@ open class MessageInputBar: UIView {
                     $0.parentStackViewPosition = position
                     leftStackView.addArrangedSubview($0)
                 }
-                guard superview != nil else { return }
```
zhongwuzw, line 523: Thanks for your `PR`, It's required, because we cannot call `layoutIfNeeded()` when view has not been added to the `superview`.

Issue #404, opened 2017-12-14
ScrollToBottom isn't really to bottom


jansvancer:
I don't know if anyone else has experienced the same issue, but I have footerView with date label and when I call `func scrollToBottom()` is doesn't really scroll to bottom, but to bottom of the bubble. So I don't see the footerView. Is there any fix?

SD10:
@jansvancer Can you give us any more information that we asked you to provide in the issue template. This is being worked on in #395 but I'm not sure if it fixes what you're describing

jansvancer:
Maybe the #395 solves this. I'm not sure. I didn't find it before. Sorry.

Here is more information:

## General Information

* MessageKit Version: 0.11.0

* iOS Version(s): 11

* Swift Version: 4

* Devices/Simulators: both

* Reproducible in ChatExample? (Yes/No): not sure. For me the scrollToBottom in example doesn't work at all.



##  What happened?

After the first layout or whenever the `func scrollToBottom()` is called it scrolls just to bottom of the label. Not to bottom of footerView.
More in the image
<img width="313" alt="snimek obrazovky 2017-12-14 v 20 14 52" src="https://user-images.githubusercontent.com/21357098/34010058-8cba6a5a-e10b-11e7-9e43-86a1b987c8c5.png">



## What did you expect to happen?

Scroll to bottom of footerView.
Like this:
<img width="312" alt="snimek obrazovky 2017-12-14 v 20 15 07" src="https://user-images.githubusercontent.com/21357098/34010121-c79a8a7e-e10b-11e7-84fe-db0f8f740ab1.png">

SD10:
@jansvancer It looks like this issue has been resolved and will be fixed in the upcoming release. Thanks for taking the time to open this issue 👍

jansvancer:
Perfect, thank you :)

zhongwuzw:
@jansvancer , Good catch!:+1:, the `development` branch has been resolved `scroll to bottom bubble` issue, yes, we ignore the last `footerView`, we'll fix it in the upcoming release. Thanks again.:+1:

cc @SD10

PR #408, opened 2017-12-16
Fix scroll to bottom instead of scroll to bottom bubble

zhongwuzw:
#404 

Maybe we can always `scrollRectToVisible` instead of `scrollToItem`.

### DIFF
--- a/Sources/Views/MessagesCollectionView.swift
+++ b/Sources/Views/MessagesCollectionView.swift
@@ -64,20 +64,10 @@ open class MessagesCollectionView: UICollectionView {
     // MARK: - Methods
 
     public func scrollToBottom(animated: Bool = false) {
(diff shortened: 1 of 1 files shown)


SD10 (approved):

nathantannar4:
So why does animated not work here?

zhongwuzw:
@nathantannar4 ,not work? How do you use it, you use `scrollToBottom(animated: true)`?

SD10:
I'll test this again after the holidays but `scrollToBottom()` is difficult to support in a variety of situations. It behaves differently depending on the `collectionView` `contentSize`. What we have now is working better than it not working at all 😕

nathantannar4:
@zhongwuzw Ya I saw there was an animated parameter but in the Slack discussion we have been saying animated was not supported

PR #416, opened 2017-12-18
Add configureImageForAvatarView(::::) protocol to expose avatarView &…

zhongwuzw:
#164 

- [x] Fix tests error.
- [x] Add `avatar(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)` deprecated handler.
- [x] Add CHANGELOG entry.

### DIFF
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -267,10 +267,6 @@ extension ConversationViewController: MessagesDataSource {
         return messageList[indexPath.section]
     }
 
-    func avatar(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> Avatar {
-        return SampleData.shared.getAvatarFor(sender: message.sender)
-    }
-
     func cellTopLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
         let name = message.sender.displayName
         return NSAttributedString(string: name, attributes: [NSAttributedStringKey.font: UIFont.preferredFont(forTextStyle: .caption1)])
@@ -321,6 +317,11 @@ extension ConversationViewController: MessagesDisplayDelegate {
         return .bubbleTail(corner, .curved)
 //        let configurationClosure = { (view: MessageContainerView) in}
 //        return .custom(configurationClosure)
+    }
+    
+    func configureImageForAvatarView(avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
+        let avatar = SampleData.shared.getAvatarFor(sender: message.sender)
+        avatarView.set(avatar: avatar)
     }
 
     // MARK: - Location Messages
--- a/Example/Sources/ConversationViewController.swift
+++ b/Example/Sources/ConversationViewController.swift
@@ -319,7 +319,7 @@ extension ConversationViewController: MessagesDisplayDelegate {
 //        return .custom(configurationClosure)
     }
     
-    func configureImageForAvatarView(avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
+    func configureAvatarView(avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
         let avatar = SampleData.shared.getAvatarFor(sender: message.sender)
         avatarView.set(avatar: avatar)
     }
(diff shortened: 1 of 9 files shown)


SD10:
@zhongwuzw I could only review this on mobile right now but:
- I’m think we change the name to `configureAvatarView`
- We look into whether we need the `Avatar` struct anymore

zhongwuzw:
@SD10 

> I’m think we change the name to configureAvatarView

The reason I choose `configureImageForAvatarView` is we just tell the user who only configure the `image`, don't need to do other configuration like `frame` or something.

> We look into whether we need the Avatar struct anymore

I think we can reserve it,it provide convenient way to set image or initials. But we also need to supply direct setter to set image or initials. 
I want to change `AvatarView` subclass from `UIView` to `UIImageView` to decrease hierarchy num . What do you think?

SD10:
> I want to change AvatarView subclass from UIView to UIImageView to decrease hierarchy num . 

@zhongwuzw I think changing `UIView` to `UIImageView` is a good idea.

> The reason I choose configureImageForAvatarView is we just tell the user who only configure the image, don't need to do other configuration like frame or something.

I still think we should change `configureImageForAvatarView` -> `configureAvatarView`
The user will likely do whatever they want, regardless of the method name.

> I think we can reserve it,it provide convenient way to set image or initials. But we also need to supply direct setter to set image or initials.

I agree 👍

zhongwuzw:
@SD10 :ok_hand:, I'll update after a moment.

Sources/Views/AvatarView.swift
```diff
@@ -24,27 +24,25 @@
...
     // MARK: - Properties
-
-    open var avatar: Avatar = Avatar()
-
-    open var imageView = UIImageView()
-
-    public var image: UIImage? {
-        return imageView.image
+    
+    open var initials: String? {
```
SD10, line 31: Do you think we should just make `setImageFrom(initials:)` public?

SD10: The reason I say that is because the property will store a reference to `initials`... then if we use `set(avatar: Avatar)` a different image may be set... but `initials` property could hold the wrong value.

Sources/Views/AvatarView.swift
```diff
@@ -160,17 +161,14 @@ open class AvatarView: UIView {
         contentMode = .scaleAspectFill
         layer.masksToBounds = true
         clipsToBounds = true
-        imageView.contentMode = .scaleAspectFill
-        imageView.frame = frame
-        addSubview(imageView)
-        imageView.image = avatar.image ?? getImageFrom(initials: avatar.initials)
+        contentMode = .scaleAspectFill
```
SD10: We set this already above ^

Example/Sources/ConversationViewController.swift
```diff
@@ -322,6 +318,11 @@ extension ConversationViewController: MessagesDisplayDelegate {
 //        let configurationClosure = { (view: MessageContainerView) in}
 //        return .custom(configurationClosure)
     }
+    
+    func configureAvatarView(avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
```
SD10: Nit but can we make this `configureAvatarView(_ avatarView: ...`. The `avatarView` is redundant

Sources/Protocols/MessagesDataSource.swift
```diff
@@ -51,16 +51,6 @@ public protocol MessagesDataSource: AnyObject {
...
 
-    /// The `Avatar` information to be used by the `AvatarView`.
-    ///
-    /// - Parameters:
-    ///   - message: The `MessageType` that will be displayed by this cell.
-    ///   - indexPath: The `IndexPath` of the cell.
-    ///   - messagesCollectionView: The `MessagesCollectionView` in which this cell will be displayed.
-    ///
-    /// The default value returned by this method is `Avatar()`.
-    func avatar(for message: MessageType, at  indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> Avatar
```
SD10, line 62: We have to add this to `MessageKit+Availability.swift` to  let people know this change

Sources/Views/AvatarView.swift
zhongwuzw: ok, I'll make `initials` to `private`.

Sources/Protocols/MessagesDataSource.swift
zhongwuzw: ok.

SD10:
@zhongwuzw I'll review this again tomorrow. This project will have thousands of commits regardless, our git history is messy, but it would be great if you try to rebase or amend commits where possible.

allaire:
Great job 👍  this will be useful 🎉

SD10 (approved):
Thanks for working on this @zhongwuzw 🎉

PR #417, opened 2017-12-20
Move tap gesture from cell to collectionView

zhongwuzw:
Reduce tap gesture amount by move tap gesture from cell to collectionView.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Changed
+- **Breaking Change** Moved the `handleTapGesture(_ gesture: UIGestureRecognizer)` method from `MessagesCollectionViewCell` to `MessagesCollectionView`.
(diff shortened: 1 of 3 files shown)


SD10:
@zhongwuzw Can you explain this a little more to me? It feels awkward moving this into `MessagesCollectionView`

nathantannar4:
Would it help performance? Having one tap gesture rather than X. I know you can very easily translate a tap to a point and this get the cell that was tapped 

zhongwuzw:
@SD10 @nathantannar4 , two reasons for changes.

1. As @nathantannar4 says, it's for performance, we don't need to create tap gesture for each cell.
2. At the moment, we only need tap gesture on `messageContainerView` of `cell`, maybe we need tap handler on `headerView`、`FooterView`..... in the future，we must add more and more tap gesture. But by translate the tap gesture to `collectionView`, we can easily do it.

SD10:
@zhongwuzw Can you add a CHANGELOG entry?

### Changed
- **Breaking Change** Moved the `handleTapGesture(_ gesture: UIGestureRecognizer)` method from `MessagesCollectionViewCell` to `MessagesCollectionView`.

PR #422, opened 2017-12-22
Added copy support for message bubble

zhongwuzw:
#248

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - Added `configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)` method in `MessagesDisplayDelegate` `protocol` to configure `avatarView`.
 [#416](https://github.com/MessageKit/MessageKit/pull/416) by [@zhongwuzw](https://github.com/zhongwuzw).
 
+- Added copy support for message bubble.
+[#418](https://github.com/MessageKit/MessageKit/pull/418) by [@zhongwuzw](https://github.com/zhongwuzw).
+
 ### Removed
 
 - **Breaking Change** Removed `avatar(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)` method of `MessagesDataSource`, use `configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)` instead.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - Added `configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)` method in `MessagesDisplayDelegate` `protocol` to configure `avatarView`.
 [#416](https://github.com/MessageKit/MessageKit/pull/416) by [@zhongwuzw](https://github.com/zhongwuzw).
 
-- Added copy support for message bubble.
+- Added copy support for image, text, and emoji messages.
 [#418](https://github.com/MessageKit/MessageKit/pull/418) by [@zhongwuzw](https://github.com/zhongwuzw).
 
 ### Removed
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -59,6 +59,9 @@ open class MessagesViewController: UIViewController {
         return false
     }
     
+    /// Indicated selected indexPath when handle menu action
+    fileprivate var selectedIndexPathForMenu: IndexPath?
+    
     /// A Boolean value used to determine if `viewDidLayoutSubviews()` has been called.
     private var isFirstLayout: Bool = true
     
@@ -84,6 +87,7 @@ open class MessagesViewController: UIViewController {
         setupConstraints()
         registerReusableViews()
         setupDelegates()
+        addMenuControllerObservers()
 
     }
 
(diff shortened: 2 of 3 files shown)


Sources/Controllers/MessagesViewController.swift
```diff
@@ -328,3 +366,35 @@ fileprivate extension MessagesViewController {
...
+        NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+    }
+    
+    /// Show menuController and set target rect to selected bubble
+    @objc func menuControllerWillShow(aNotification:Notification) {
+        if let currentMenuController = aNotification.object as? UIMenuController,
+            let selectedIndexPath = selectedIndexPathForMenu {
+            NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+            defer {
+                NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerWillShow(aNotification:)), name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
```
nathantannar4: Just a question, why do we keep adding ab observer here? It looks like we never remove it

nathantannar4 (commented):
Runs smooth!

Sources/Controllers/MessagesViewController.swift
zhongwuzw: @nathantannar4 , emm, we need remove here first, because we call `currentMenuController.setMenuVisible(true, animated: true)` later, it would leads to endless loop if we not remove in advance.

nathantannar4:
@zhongwuzw i mean why add the observer

zhongwuzw:
@nathantannar4 , we need customize the target rect to bubble, default would target rect to cell.
I remove observer at here. https://github.com/MessageKit/MessageKit/blob/eb1c6f50a631119f62d7f8875fb03221f6e7e789/Sources/Controllers/MessagesViewController.swift#L386

SD10:
Thanks for doing this @zhongwuzw! It will take me some time to be able to review it

Sources/Controllers/MessagesViewController.swift
```diff
@@ -328,3 +366,35 @@ fileprivate extension MessagesViewController {
...
+        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerWillShow(aNotification:)), name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+    }
+    
+    /// Remove observer for `UIMenuControllerWillShowMenu` notification
+    fileprivate func removeMenuControllerObservers() {
+        NotificationCenter.default.removeObserver(self, name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
+    }
+    
+    /// Show menuController and set target rect to selected bubble
+    @objc func menuControllerWillShow(aNotification:Notification) {
```
SD10: @zhongwuzw Can we just name this parameter `notification`? And add a space after the colon

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - Added `configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)` method in `MessagesDisplayDelegate` `protocol` to configure `avatarView`.
 [#416](https://github.com/MessageKit/MessageKit/pull/416) by [@zhongwuzw](https://github.com/zhongwuzw).
 
+- Added copy support for message bubble.
```
SD10: "Added copy support for image, text, and emoji messages". 

This would sound better. Sorry!

Sources/Controllers/MessagesViewController.swift
```diff
@@ -328,3 +366,35 @@ fileprivate extension MessagesViewController {
         return 0
     }
 }
+
+// MARK: - Menu Handling
+extension MessagesViewController {
+    /// Add observer for `UIMenuControllerWillShowMenu` notification
+    fileprivate func addMenuControllerObservers() {
+        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.menuControllerWillShow(aNotification:)), name: NSNotification.Name.UIMenuControllerWillShowMenu, object: nil)
```
SD10: Can we use the shorthand version of the notification names please `.UIMenuControllerWillShowMenu`?

Sources/Views/Cells/MessageCollectionViewCell.swift
```diff
@@ -131,6 +131,15 @@ open class MessageCollectionViewCell: UICollectionViewCell, CollectionViewReusab
             break
         }
     }
+    
+    /// Handle long press gesture, return true when gestureRecognizer's touch point in `messageContainerView`'s frame
+    open override func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+        let touchPoint = gestureRecognizer.location(in: self)
+        if gestureRecognizer.isKind(of: UILongPressGestureRecognizer.self) {
```
SD10: How about this? 
```Swift
guard gestureRecognizer.isKind(of: UILongPressGestureRecognizer.self) else { return false }
return messageContainerView.frame.contains(touchPoint)
```

SD10:
@zhongwuzw  Think there is any way to prevent the menu from appearing outside the content view?
screen shot 2018-01-06 at 11 18 57 pm

What about adding support for copying the image from `AvatarView`?

zhongwuzw:
> Think there is any way to prevent the menu from appearing outside the content view?

@SD10 , good question, another problem, what if message's height exceed the screen height?where should we position menu at?

SD10:
> good question, another problem, what if message's height exceed the screen height?where should we position menu at?

@zhongwuzw I think we should position it at the bottom of the message. It may work like that already.

zhongwuzw:
@SD10 , I updated the code to handle some situations, like situation below, menu would be blocked by keyboard, `JSQMessagesViewController` still has this issue. 
menu-example

SD10 (approved):
@zhongwuzw If the menu is covered by the keyboard in JSQMessagesViewController then we can merge this for now. Let's open an issue to track this behavior?

PR #423, opened 2017-12-23
Image Paste Support for InputTextView

nathantannar4:


SD10:
@nathantannar4 Thank you for working on this! Would you also mind opening a GitHub issue that matches this new feature? It may take me some time to review it with the holidays right around the corner.

Example/Sources/ConversationViewController.swift
```diff
@@ -440,12 +440,25 @@ extension ConversationViewController: MessageLabelDelegate {
...
-        let id = UUID().uuidString
-        let message = MockMessage(attributedText: attributedText, sender: currentSender(), messageId: id, date: Date())
-        messageList.append(message)
+        
+        // Each NSTextAttachment that contains an image will count as one empty character in the text: String
+        if text.count > inputBar.inputTextView.images.count {
+            
+            let attributedText = NSAttributedString(string: text, attributes: [.font: UIFont.systemFont(ofSize: 8), .foregroundColor: UIColor.blue])
+            
+            let message = MockMessage(text: text, sender: currentSender(), messageId: UUID().uuidString, date: Date())
```
zhongwuzw: Should use `attributeText` instead of `text`.

zhongwuzw (changes requested):
```
let attributedText = NSAttributedString(string: text, attributes: [.font: UIFont.systemFont(ofSize: 8), .foregroundColor: UIColor.blue])          
let message = MockMessage(text:text, sender: currentSender(), messageId: UUID().uuidString, date: Date())
```

Should use `attributeText` instead of `text` when create `MockMessage` instance.

Example/Sources/ConversationViewController.swift
zhongwuzw: Should use `attributeText` instead of `text` when create `MockMessage` instance.

zhongwuzw:
@nathantannar4 , @SD10 , yep,I think we need create a issue to discuss the behavior of copy feature.Because it has some situations to talk, like what if message mix of text and image.

Sources/Views/InputTextView.swift
```diff
@@ -49,6 +49,13 @@ open class InputTextView: UITextView {
         }
     }
     
+    /// The images that are currently stored as NSTextAttachment's
+    open var images: [UIImage] {
+        return getAttachedImages()
+    }
+    
+    open var isImagePasteEnabled: Bool = true
```
zhongwuzw: Maybe we can add a property in `MessageInputBar`? , I think put ` isImagePasteEnabled` in `inputTextView` maybe a little bit deep.

nathantannar4:
@zhongwuzw @SD10 sure. Ya the way I thought is that I just expose an images property and it’s up to the developer to support getting the image and sending it as a separate message

Sources/Views/InputTextView.swift
nathantannar4: Fair point, but I think it should still be apart of `InputTextView`. Maybe add the following in `MessageInputBar` ?

```
open var isImagePasteEnabled: Bool {
   get {
       return inputTextView.isImagePasteEnabled
   }
   set {
       inputTextView.isImagePasteEnabled = newValue
   }
}
```

zhongwuzw: :+1: That's what I think.

Sources/Views/InputTextView.swift
```diff
@@ -176,4 +184,84 @@ open class InputTextView: UITextView {
...
+        
+        let attributes: [NSAttributedStringKey:Any] = [
+            NSAttributedStringKey.font : font,
+            NSAttributedStringKey.foregroundColor : textColor,
+        ]
+
+        // Add the new image as an NSTextAttachment
+        let attributedImageString = NSAttributedString(attachment: textAttachment(using: image))
+        
+        let newAttributedStingComponent = NSMutableAttributedString(string: "")
```
zhongwuzw: Maybe we can use `init(attachment: NSTextAttachment)` in here instead of use `append`.

nathantannar4:
I have added attachment redrawing on orientation change and component parsing

zhongwuzw:
@nathantannar4 , Did you committed?

nathantannar4:
oops now i did

Sources/Views/InputTextView.swift
```diff
@@ -176,4 +197,146 @@ open class InputTextView: UITextView {
...
+        // Add a new line character after the image, this is what iMessage does
+        newAttributedStingComponent.append(NSAttributedString(string: "\n"))
+        
+        newAttributedStingComponent.addAttributes(attributes, range: NSRange(location: 0, length: newAttributedStingComponent.length))
+        
+        // Paste over selected text
+        textStorage.replaceCharacters(in: selectedRange, with: newAttributedStingComponent)
+        
+        // Advance the range to the selected range plus the number of characters added
+        selectedRange = NSRange(location: selectedRange.location + 3, length: 1)
```
zhongwuzw: The parameter `length` should be 0.

zhongwuzw: I don't know wether or not just add 3 at here is ok, because when `attributedText.length == 0`, the total attributedString's length is 2.

zhongwuzw (changes requested):

Sources/Views/InputTextView.swift
nathantannar4: Yup thanks for the catch

Sources/Views/InputTextView.swift
zhongwuzw: @nathantannar4 ,Add `isImagePasteEnabled` property in `MessageInputBar` is more friendly, and we should set `isImagePasteEnabled` property's access level to `internal` in `inputTextView` to prevent from inconsistently.

zhongwuzw: What do you think?

Sources/Views/InputTextView.swift
zhongwuzw: @nathantannar4 , I fixed a bug, see #426

Example/Sources/ConversationViewController.swift
```diff
@@ -440,12 +440,29 @@ extension ConversationViewController: MessageLabelDelegate {
...
 
     func messageInputBar(_ inputBar: MessageInputBar, didPressSendButtonWith text: String) {
-        let attributedText = NSAttributedString(string: text, attributes: [.font: UIFont.systemFont(ofSize: 8), .foregroundColor: UIColor.blue])
-        let id = UUID().uuidString
-        let message = MockMessage(attributedText: attributedText, sender: currentSender(), messageId: id, date: Date())
-        messageList.append(message)
+        
+        // Each NSTextAttachment that contains an image will count as one empty character in the text: String
+        
+        for component in inputBar.inputTextView.components {
```
SD10, line 446: This is a blind review so don't make any changes but - It may be nice to have some sort of delegate method for this?

nathantannar4:
@SD10 how might you want the delegate to be?

zhongwuzw (approved):

SD10 (approved):
@nathantannar4 Thank you for doing this 👍

Issue #402, opened 2017-12-13
inputAccessoryView has transparent bottom area


runmad:
## General Information

* MessageKit Version: `development` branch @ `769dc74`

* iOS Version(s): iOS 11

* Swift Version: Swift 4

* Devices/Simulators: iPhone X

* Reproducible in ChatExample? Yes):

![simulator screen shot - iphone x - 2017-12-13 at 12 45 22](https://user-images.githubusercontent.com/708741/33961670-cbec479c-e003-11e7-9bc5-a76016cdb444.png)
![simulator screen shot - iphone x - 2017-12-13 at 12 45 17](https://user-images.githubusercontent.com/708741/33961671-cc0785c0-e003-11e7-813b-2c9ae31915bc.png)

##  What happened?

When pulling down the keyboard with an interactive gesture, the background of the `inputAccessoryView` is transparent when the top of keyboard is lower than the `window.safeAreaLayoutGuide.bottomAnchor`.

## What did you expect to happen?

The bottom of the `inputAccessoryView` should be opaque - or same as the rest of the `inputAccessoryView`.

nathantannar4:
Yes @zhongwu and I are aware of that. It is a minor issue as it only occurs when a user very slowly drags down the collectionView. It is because we lowered the priority of the window anchor to avoid constraint being broken. Not sure how to fix this at the moment 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #425, opened 2017-12-24
Fixed iPhoneX messageInputBar transparent bottom area

zhongwuzw:
#402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


nathantannar4:
This fixes the glitch on the iPhone X when slowly dismissing the keyboard?

zhongwuzw:
@nathantannar4 ,yep, it's a trick way, we just expand `backgroundView`'s bottom edge in case of iPhoneX.

SD10:
@zhongwuzw @nathantannar4 If a PR is a simple bugfix or internal change (non-API breaking) it can go into `master`. This way we can release small patches and not have to wait for a major release. Just for future reference. I will create some documentation around this to clarify

Issue #190, opened 2017-09-28
Add support for accessory view in cell


SD10:
### Support accessory view in cell

<img width="326" alt="459654d8-f5c9-11e5-8065-6e9fdf1c62ca" src="https://user-images.githubusercontent.com/7445580/30995177-001034aa-a47e-11e7-9144-1cf65061a9af.png">

Related issues:
[Add support for an accessory view to JSQMessagesCollectionViewCell #1519](https://github.com/jessesquires/JSQMessagesViewController/issues/1519)

SD10:
Closed by #710

Issue #418, opened 2017-12-20
Custom Avatar View


jbouaziz:
Hello!

I'm using your library and need to have a custom avatar view that handles more than just the image (FYI online status).

Right now the only way is to :
1. Override the `cellForItem` in a `MessagesViewController` subclass.
2. Cast the cell to the proper type
3. Add a subview if its not there yet
3. Hide the original avatar view in plain sight by setting its properties.
4. Do some custom arrangements.

```swift

    override func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
        guard let cell = super.collectionView(collectionView, cellForItemAt: indexPath) as? MessageCollectionViewCell else {
            fatalError("Unsupported cell")
        }
        
        guard let object = messagesCollectionView.messagesDataSource?.messageForItem(at: indexPath, in: messagesCollectionView) as? MyCustomMessageDataObject else {
            return cell
        }
        
        // Set avatar infos (?hack for now?)
        cell.avatarView.isHidden = !object.shouldDisplayAvatar
        if !cell.avatarView.isHidden {
            cell.avatarView.backgroundColor = .clear
            cell.avatarView.setCorner(radius: 0)
            
            var realView = cell.avatarView.subviews.first(where: {$0 is MyModule.AvatarView }) as? Pong.AvatarView
            if realView == nil {
                realView = MyModule.AvatarView()
                cell.avatarView.addSubview(realView!)

                // Add constraints
                ...
            }
            
            // Set current online status and the remote image
            ...
        }
        
        // Set attachment image
        if let cell = cell as? MediaMessageCell, let url = object.attachment?.remoteUrl {
            cell.imageView.sd_setShowActivityIndicatorView(true)
            cell.imageView.sd_setIndicatorStyle(.gray)
            cell.imageView.sd_setImage(with: url)
        }
        
        return cell
    }
```

Would you consider transforming the `AvatarView` into a protocol conforming to `UIView` and for which specific methods would be called using the appropriates delegates & datasources ? There could also be a default implementation in the library with initials and such.

I'm seeing some improvements with #416 but it is not generic enough.

## General Information

Specs | Value
------------ | -------------
MessageKit Version | 0.12.0
iOS Version(s) | 10 & 11
Swift Version | 4
Devices/Simulators | Both
Reproducible in ChatExample? | Yes

SD10:
[...]

PR #432, opened 2017-12-31
Add customizable accessoryView besides messageContainer

hyouuu:
Add an accessoryView and give client a chance to config it to whatever they like (e.g. a button, image etc)

And I found one ;) https://github.com/MessageKit/MessageKit/issues/190

NA

Any other comments?
-------------------
Let me know how you want to approach it!

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 8+ simulator

**iOS Version:** 11

**Swift Version:** 4

**MessageKit Version:** 0.12

Example screenshot where in configCell added a UIView with green background:

screen shot 2017-12-31 at 1 01 22 pm

### DIFF
--- a/Sources/Protocols/MessagesDataSource.swift
+++ b/Sources/Protocols/MessagesDataSource.swift
@@ -81,14 +81,15 @@ public protocol MessagesDataSource: AnyObject {
     /// The default value returned by this method is `nil`.
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
 
-    /// Additional config for a cell, called in cellForItem method. e.g. customize accessoryView.   
+    /// A chance to config a cell before it's presented e.g. customize accessoryView.
     ///
     /// - Parameters:
     ///   - cell: The `MessageCollectionViewCell` that can be configured.
+    ///   - message: The `MessageType` that will be displayed by this cell.
     ///   - indexPath: The `IndexPath` of the cell.
     ///
     /// The default value returned by this method is `nil`.
-    func additionalConfig(for cell: MessageCollectionViewCell, at indexPath: IndexPath)
+    func configCell(_ cell: MessageCollectionViewCell, for message: MessageType, at indexPath: IndexPath)
 }
 
 public extension MessagesDataSource {
@@ -109,5 +110,5 @@ public extension MessagesDataSource {
         return nil
     }
 
-    func additionalConfig(for cell: MessageCollectionViewCell, at indexPath: IndexPath) {}
+    func configCell(_ cell: MessageCollectionViewCell, for message: MessageType, at indexPath: IndexPath) {}
 }
--- a/Sources/Protocols/MessagesDataSource.swift
+++ b/Sources/Protocols/MessagesDataSource.swift
@@ -81,6 +81,14 @@ public protocol MessagesDataSource: AnyObject {
     /// The default value returned by this method is `nil`.
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
 
+    /// Additional config for a cell, called in cellForItem method. e.g. customize accessoryView.   
+    ///
+    /// - Parameters:
+    ///   - cell: The `MessageCollectionViewCell` that can be configured.
+    ///   - indexPath: The `IndexPath` of the cell.
+    ///
+    /// The default value returned by this method is `nil`.
+    func additionalConfig(for cell: MessageCollectionViewCell, at indexPath: IndexPath)
 }
 
 public extension MessagesDataSource {
@@ -101,4 +109,5 @@ public extension MessagesDataSource {
         return nil
     }
 
+    func additionalConfig(for cell: MessageCollectionViewCell, at indexPath: IndexPath) {}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Fixed
+
+- Fixed message bubble tail orientation invalidation in `iOS9`.
+[#469](https://github.com/MessageKit/MessageKit/pull/469) by [@zhongwuzw](https://github.com/zhongwuzw).
+
 ## [[Prerelease] 0.13.0](https://github.com/MessageKit/MessageKit/releases/tag/0.13.0)
 
 ### Fixed
--- a/Example/Sources/InboxViewController.swift
+++ b/Example/Sources/InboxViewController.swift
@@ -28,7 +28,7 @@ import SafariServices
 
 final class InboxViewController: UITableViewController {
 
-    let cells = ["Example", "Settings", "Source Code", "Contributers"]
+    let cells = ["Example", "Settings", "Source Code", "Contributors"]
     
     // MARK: - View Life Cycle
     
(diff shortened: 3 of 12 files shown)


SD10:
@hyouuu Thanks for submitting this! This is definitely on the todo list but it could take me some time to review. This could be more difficult than it seems because it changes so much in `MessagesCollectionViewFlowLayout`. An accessory view limits the available space for things like `messageContainerView`, `cellTopLabel`, cellBottomLabel`, depending on its location.

hyouuu:
Definitely! I'll play around long message & labels etc and see where it can be improved; I'm in need of this so just made the basic version - will try to cover corner cases as I get to it and add to this PR ;)

hyouuu:
Long piece seems to be working fine as is o_O
screen shot 2017-12-31 at 12 14 20 pm

SD10:
@hyouuu It will definitely require changes to the methods:
`messageContainerMaxWidth`, `cellBottomLabelMaxWidth`, `cellTopLabelMaxWidth`, etc.

hyouuu:
Yep I'm working on them as we speak - commit coming soon

Sources/Protocols/MessagesDataSource.swift
```diff
@@ -81,6 +81,14 @@ public protocol MessagesDataSource: AnyObject {
...
     func cellBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
 
+    /// Additional config for a cell, called in cellForItem method. e.g. customize accessoryView.   
+    ///
+    /// - Parameters:
+    ///   - cell: The `MessageCollectionViewCell` that can be configured.
+    ///   - indexPath: The `IndexPath` of the cell.
+    ///
+    /// The default value returned by this method is `nil`.
+    func additionalConfig(for cell: MessageCollectionViewCell, at indexPath: IndexPath)
```
SD10: I'm looking into this, but we may not want to add it in this PR. Let's keep the scope small

SD10:
I think this resembles #418. Figuring out how to allow users to customize the `AvatarView` and what kind of `AccessoryView` is a similar problem.

Sources/Protocols/MessagesDataSource.swift
hyouuu: What's your suggestion on letting clients config the accessoryView? A specific setupAccessoryView delegate?

hyouuu:
Adjusted messageContainerMaxWidth and did some refactor e.g. made config more generic with messageType.

Seems to work fine with padding and long messages:

Message without accessoryView (size & padding both .zero):
screen shot 2017-12-31 at 12 58 13 pm

With accessoryView (size 36 & padding 14):
screen shot 2017-12-31 at 1 01 22 pm

hyouuu:
For #418, I think we could supply a switch on whether to auto config avatarView, and if it's off, people can use `messagesDataSource.configCell(cell, for: message, at: indexPath)
` to add subviews to the plain avatarView to customize it

mmuharem:
I like this! How would you add a delegate for it being tapped just as the other parts of the cell (toplabel, bottom label, avatar, etc..)?

hyouuu:
@mmuharem how this would work will be like a UITableViewCell's contentView, which serves as a view holder and you could add anything on it (e.g. a UIButton with a selector upon tapping ;)

hyouuu:
I could update the example code as soon as @SD10 thinks this is the right approach, but basically in the configCell delegate method you could do something like:
```
        let button = UIButton()
        button.frame = cell.accessoryView.bounds
        cell.accessoryView.addSubview(button)

        // This is my customized tapBlock, but you could use addTarget
        button.tapBlock = {
            // Do something
        }
```

mmuharem:
@hyouuu okay, so the part that I have a question particularly is the 

```
// This is my customized tapBlock, but you could use addTarget
        button.tapBlock = {
            // Do something
        }
```

as this would mean that you have a selector in an extension. I am using the addTarget but this is the part I was not able to figure out. 

So, where and how did you put the .tapBlock function?

hyouuu:
What's the issue with a function in an extension? You could put the method in the main class as well. Remember to add @objc if it's a selector func

mmuharem:
Well, that's what I am trying to figure out. 

Here is what I've tried in the MessagesDataSource.swift file. Is this what you mean by the 'main class'?:

```
public protocol MessagesDataSource: AnyObject {
      ....
      @objc func toggleReact(_ button: UIButton)
}

public extension MessagesDataSource {
     ....
    func configCell...{
             button.addTarget(self, action:  #selector(self.toggleReact) , for: .touchUpInside)
      }

   func toggleReact(_ button: UIButton) {
        print("button Tapped")
    }

}

```

The errors I receive:
- button.addTarget (Argument of selector refers to instance method 'toggleReact'that is not exposed to objective c. Add objc to expose this instance
- @objc func toggleReact (@objc can only be used with members of classes , @objc protocols, and concrete extensions of classes. Replace @objc with "". 

Instead, if I use MessagesCollectionView.swift as the main class, then it compiles but crashes upon click with unrecognized selector

hyouuu:
Well @mmuharem what I do is that I have a subclass of MessagesViewController called ChatVC that implements configCell as well as toggleReact:
```
class ChatVC: MessagesViewController,
    MessagesDataSource,
    MessagesDisplayDelegate,
    MessagesLayoutDelegate,
    MessageCellDelegate
{
    func configCell...{
        button.addTarget(self, action:  #selector(self.toggleReact) , for: .touchUpInside)
    }

    func toggleReact(_ button: UIButton) {
        print("button Tapped")
    }
}
```

mmuharem:
omg so simple that I overlooked that. Thank you so much for the help!

SD10:
@hyouuu To update you on this it's being blocked by some other things 😐 I need to do some stuff inside of `MessagesCollectionViewFlowLayout`

hyouuu:
Thanks for the update - hope everything goes well!

SD10:
@hyouuu Tonight I'm going to break most of the internals for `MessagesCollectionViewFlowLayout`. We're going to have to re-open this another time. It may not be possible to resolve the merge conflicts, and this PR is great, but the logic is not correct (yet). I'll use this as a basis for when we make the addition and credit you for any work I use or migrate it for you and ping you for assistance 😃 Thanks for taking the time to put this together

hyouuu:
@SD10 thanks for the update and the changes sound exciting! Hopefully we can get to this soon as otherwise I wouldn't be able to update MessageKit for my current in-dev app T_T Let me know of any assist I could help provide!

SD10:
We can move forward on this again but it really does take a lot of work to get in. 
It would affect the following methods:
```
cellContentHeight(for message: MessageType, at indexPath: IndexPath) -> CGFloat
cellTopLabelMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat
cellBottomLabelMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat
messageContainerMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat
```

Things to consider: 
- What if `AvatarView` width + `AccessoryView` width > `itemWidth`?

Pretty extreme edge case but some stuff like this is on my mind.

hyouuu:
Cool will give a try soon. For the edge case, my first thought is `AccessoryView` is more important than AvatarView because it usually takes some kind of user action rather than pure displaying (although you could argue `AvatarView` could be customized to do actions too).
What if `AvatarView` width > `itemWidth` or `AccessoryView` width > `itemWidth`? Should we display part of them, shrink them or throw an error?

PR #434, opened 2018-01-01
Input Plugin Support [In Progress]

nathantannar4:
A way to easily add plugins to the `MessageInputBar`


nathantannar4:
Closing for now, there does not seem to be a demand

PR #444, opened 2018-01-08
Fixed wrong contentInset calculation when navigationBar is hidden

zhongwuzw:


Sources/Controllers/MessagesViewController.swift
```diff
@@ -88,12 +88,14 @@ open class MessagesViewController: UIViewController {
     }
 
     open override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
```
SD10: I don't think this is necessary, Apple docs says:
> Your view controller can override this method to make changes after the view lays out its subviews. The default implementation of this method does nothing.

nathantannar4 (dismissed):

Sources/Controllers/MessagesViewController.swift
zhongwuzw: Emm, yep, default implementation does nothing, I just keep the form of call `super` when override method, I can move it.

SD10 (approved):

Issue #448, opened 2018-01-09
Cell rendering bug


Elshad:
## General Information

* MessageKit Version: 0.12.1

* iOS Version(s): 11

* Swift Version: 4

* Devices/Simulators: Tested in Simulator

* Reproducible in ChatExample? (Yes/No): Yes

##  What happened?
<img width="363" alt="screen shot 2018-01-09 at 18 14 30" src="https://user-images.githubusercontent.com/2334731/34725093-59346164-f569-11e7-9781-eaa42c7ca117.png">
<img width="364" alt="screen shot 2018-01-09 at 18 21 30" src="https://user-images.githubusercontent.com/2334731/34725324-f761e032-f569-11e7-870d-61ec4717d60c.png">

## What did you expect to happen?
<img width="366" alt="screen shot 2018-01-09 at 18 14 43" src="https://user-images.githubusercontent.com/2334731/34725112-64614ba6-f569-11e7-8230-d0f55b86df13.png">

## When
When swipe controller to half (as shown below picture) and return, cells which not visible while swiping after return rendered badly

<img width="466" alt="screen shot 2018-01-09 at 18 15 18" src="https://user-images.githubusercontent.com/2334731/34725283-da0dbab0-f569-11e7-95cc-30e2358ab9c2.png">

P.S. Sorry for my bad explanation of process

Elshad:
![screencast_480](https://user-images.githubusercontent.com/2334731/34727121-6645cd4c-f56f-11e7-9d2e-f984c519b368.gif)

<img width="395" alt="screen shot 2018-01-09 at 18 54 23" src="https://user-images.githubusercontent.com/2334731/34727168-8c75dab6-f56f-11e7-8393-f7899b366012.png">

SD10:
@Elshad more information as to how you caused this would be great 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Elshad:
@SD10 
1. I downloaded the ChatExample
2. I opened app in Simulator
3. Action causing bug i paste below as GIF
![screencast_480](https://user-images.githubusercontent.com/2334731/34731160-f2482d66-f57a-11e7-84c3-075fe8fb2011.gif)

As seen content of some cells is not displayed correctly

nathantannar4:
I dont see an issue there? Are you talking about how it slides down? That's standard UIScrollView and UINavigationController interaction

[...]

PR #461, opened 2018-01-15
Fix bubble backgroundColor deranged when swipe

zhongwuzw:
TODO:
- [ ] Add CHANGELOG

message bubble backgroundColor would deranged that operation the same as #448 

message-color-bug
message-bubble-backcolor-disorder

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -82,6 +82,12 @@ open class MessagesViewController: UIViewController {
         setupDelegates()
         addMenuControllerObservers()
     }
+    
+    open override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        view.layoutIfNeeded()
+        messagesCollectionView.collectionViewLayout.invalidateLayout()
+    }
 
     open override func viewDidLayoutSubviews() {
         // Hack to prevent animation of the contentInset after viewDidAppear


Sources/Controllers/MessagesViewController.swift
```diff
@@ -82,6 +82,12 @@ open class MessagesViewController: UIViewController {
         setupDelegates()
         addMenuControllerObservers()
     }
+    
+    open override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        view.layoutIfNeeded()
+        messagesCollectionView.collectionViewLayout.invalidateLayout()
```
SD10, line 89: The reason I haven't merged this is I'm a little concerned that the fix involves invalidating the layout on `viewWillAppear`. I have to check this one

zhongwuzw: 👌

nathantannar4:
@zhongwuzw Is this still a bug?

SD10:
@nathantannar4 It likely is, but we can't solve it this way because invalidating the layout in this place is too heavy-handed

nathantannar4:
Perhaps just reload the cells at the visible index paths?

zhongwuzw:
Emm, a long time for this issue, but I tested a moment ago, I think the bug already fixed in #707 .

SD10:
We'll close this for now then 👍

PR #466, opened 2018-01-17
Add feature for async media image

zhongwuzw:
Todo:
- [ ] Add entry in `CHANGELOG`.

### DIFF
--- a/Sources/Views/Cells/MediaMessageCell.swift
+++ b/Sources/Views/Cells/MediaMessageCell.swift
@@ -54,8 +54,13 @@ open class MediaMessageCell: MessageCollectionViewCell {
 
     open override func configure(with message: MessageType, at indexPath: IndexPath, and messagesCollectionView: MessagesCollectionView) {
         super.configure(with: message, at: indexPath, and: messagesCollectionView)
+        
+        guard let displayDelegate = messagesCollectionView.messagesDisplayDelegate else {
+            fatalError(MessageKitError.nilMessagesDisplayDelegate)
+        }
+        
         switch message.data {
-        case .photo(let image):
+        case .photo(_, let image):
             imageView.image = image
             playButtonView.isHidden = true
         case .video(_, let image):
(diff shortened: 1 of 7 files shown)


Sources/Views/Cells/MediaMessageCell.swift
```diff
@@ -64,5 +69,7 @@ open class MediaMessageCell: MessageCollectionViewCell {
         default:
             break
         }
+        
+        displayDelegate.configureMediaMessageImageView(imageView, for: message, at: indexPath, in: messagesCollectionView)
```
SD10, line 73: Interesting approach. Let me think about this some more. My concern is this then overwriting the `image` in the `.photo` enum. 🤔

Jeyhey:
@SD10 Well, the proposed approach in the pull request is similar to how the video enum is defined. In that sense it is consistent. Could be a good approach after all.

nathantannar4 (approved):

SD10:
So after we add something like the `PhotoItem` protocol from #582 we will get something very similar to this existing enum case. My question is this:

Do we want to add the following method as a hook to add images async?
```Swift
func configureMediaMessageImageView(_ imageView: UIImageView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)
```
Or should we make this more generic and just give users access to the `MessageContainerView`?
```Swift
func configureMessageContainerView(_container: MessageContainerView, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)
```
The reason being maybe users have `.custom` data types that need async support.

Let's also consider that users can achieve this without a method at all, simply by overriding `cellForItem(at indexPath:)`.

cc @zhongwuzw

zhongwuzw:
@SD10 , I don't like overriding `cellForItem` method, because user needs to judge kind of `cell` and call `super` ...
I still prefer to choose provide `imageView`, If We only provide `containerView`, users maybe cannot set image easily.

SD10:
Going to be too difficult to resolve the conflicts for this one but I'll open an issue

Issue #376, opened 2017-12-05
MessageStyle TailCorner wrong in iOS 9


shahnewaz469:
## General Information

* MessageKit Version:
0.11.0
* iOS Version(s):
iOS 9
* Swift Version:
4.0
* Devices/Simulators:
iPhone 6s
* Reproducible in ChatExample? (Yes/No):
yes

##  What happened?
MessageStyle.TailCorner always shows in the bottom right for both sender and receiver.

## What did you expect to happen?
It should be bottom left for the receiver and bottom right for the sender.

nathantannar4:
Thanks for bringing this to our attention. I can confirm that is an issue. I notice a steep degradation of scroll performance in general with the message labels @SD10

zhongwuzw:
@shahnewaz469  , Thanks for your issue, we have fixed it and will release in next version.

Issue #307, opened 2017-10-31
Use CAShapeLayer for MessageStyle and not UIImages


SD10:
The **biggest** performance hit in this project comes from the application of the `MessageStyle`.

### Thoughts on improving this:
- We use `CAShapeLayer` to create the bubble on all `TextMessageCells`
- We do not allow "tails" on media message cells (see iMessage they do the same)

(The tails also cause inconsistencies in the image of a media message based on whether the message is to or from the current sender)

SD10:
@ndonald2 Had some good suspicions that this may be less helpful than before because previously we weren't caching images and were constantly loading from disk. 

This approach is still far more customizable, however, and kind of removes the `MessageStyle.bubbleTailOutline` hack where we need to use both an `UIImage` mask and an actual `UIImage` to achieve the proper style.

Definitely worth re-evaluating before moving forward with this

PR #469, opened 2018-01-18
Fix message bubble tail orientation invalidation in iOS9

### LINKED ISSUE #468: TailCorner bottomLeft not working in iOS 9
## General Information

* MessageKit Version: 0.13.0

* iOS Version(s): iOS 9.3.5

* Swift Version: 4.0

* Devices/Simulators: iPhone/iPod

##  What happened?

In iOS 9 the ChatExample:
img_0809

As in iOS 11, ChatExample:

img_1927

zhongwuzw:
Fixed #468 , #376

### DIFF
--- a/Sources/Models/MessageStyle.swift
+++ b/Sources/Models/MessageStyle.swift
@@ -89,7 +89,7 @@ public enum MessageStyle {
             image = UIImage(cgImage: cgImage, scale: image.scale, orientation: corner.imageOrientation)
         }
 
-        return stretch(image).withRenderingMode(.alwaysTemplate)
+        return stretch(image)
     }
 
     // MARK: - Private
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Fixed
+
+- Fixed message bubble tail orientation invalidation in `iOS9`.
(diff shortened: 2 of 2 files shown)


Sources/Models/MessageStyle.swift
```diff
@@ -89,7 +89,7 @@ public enum MessageStyle {
             image = UIImage(cgImage: cgImage, scale: image.scale, orientation: corner.imageOrientation)
         }
 
-        return stretch(image).withRenderingMode(.alwaysTemplate)
```
SD10, line 92: I think `.withRenderingMode(.alwaysTemplate)` is necessary, no? We want to ignore the color of the images so that we can apply a different bubble color or border color. The image is black.

zhongwuzw: @SD10 , Nope, we just use the image to mask the `containerView`, so it's just use maskView's alpha channel.

SD10: @zhongwuzw I don't know what I would do without you 😅

SD10 (approved):

Sources/Models/MessageStyle.swift
zhongwuzw: By the way, we should optimize that and not use `mask`, it's performance penalty.

SD10: Yes, I agree. I'm trying to remove the images altogether. See #307 #324. I can't figure out a good API. I will work on this more

PR #470, opened 2018-01-19
Optimize collectionView scroll behavior when abruptly stops scrolling

### LINKED ISSUE #467: Collection view abruptly stops scrolling when dragging down keyboard
## General Information

* MessageKit Version: 0.12.0

* iOS Version(s): 10

* Swift Version: 4

* Devices/Simulators: iPhone SE

##  What happened?
When dragging down the keyboard (quick drag down and release so that the collection view continues scrolling up) the collection view suddenly stops scrolling. This happens only if `maintainPositionOnKeyboardFrameChanged` is set to true.

The solution I found is the following. In `handleKeyboardDidChangeState(_ notification: Notification)` of `MessagesViewController` change
```
if maintainPositionOnKeyboardFrameChanged {
    messagesCollectionView.setContentOffset(contentOffset, animated: false)
}
```
to
```
if maintainPositionOnKeyboardFrameChanged && differenceOfBottomInset != 0 {
    messagesCollectionView.setContentOffset(contentOffset, animated: false)
}
```

This seems to fix the issue without having side effects on the behaviour of input bar and the collectionView.

The collection view should continue scrolling, decelerate and finally slowly stop.

zhongwuzw:
Fixed #467

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Fixed
+
+- Optimize `collectionView` scroll behavior when abruptly stops scrolling.
+[#470](https://github.com/MessageKit/MessageKit/pull/470) by [@zhongwuzw](https://github.com/zhongwuzw).
+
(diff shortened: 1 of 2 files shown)


SD10 (changes requested):
@zhongwuzw Care to add a CHANGELOG entry? Appreciate your hard work 💯

PR #475, opened 2018-01-19
Fix messageInputBar sendButton not disabled in some situations

### LINKED ISSUE #473: MessageInputBar.sendButton not disabled after clear inputTextView in program
## General Information

* MessageKit Version: 0.13.0

* iOS Version(s): 11.x

* Swift Version: 4.0

* Devices/Simulators: iPhone

##  What happened?
MessageInputBar.sendButton is still enabled after sending the message. InputTextView's content is cleared in example program:

inputBar.inputTextView.text = String()

The sendButton should be disabled.

The notification .UITextViewTextDidChange was not called after clearing inputTextView content programming. Maybe just add 1 one more line to disable the sendButton after clearing.

zhongwuzw:
Fixed #473

### DIFF
--- a/Sources/Views/InputTextView.swift
+++ b/Sources/Views/InputTextView.swift
@@ -39,12 +39,14 @@ open class InputTextView: UITextView {
     
     open override var text: String! {
         didSet {
+            postTextViewDidChangeNotification()
             placeholderLabel.isHidden = !text.isEmpty
         }
     }
     
     open override var attributedText: NSAttributedString! {
         didSet {
+            postTextViewDidChangeNotification()
             placeholderLabel.isHidden = !text.isEmpty
         }
     }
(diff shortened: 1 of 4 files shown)


Sources/Views/MessageInputBar.swift
```diff
@@ -703,6 +703,5 @@ open class MessageInputBar: UIView {
     /// Invalidates each of the inputManagers
     open func didSelectSendButton() {
         delegate?.messageInputBar(self, didPressSendButtonWith: inputTextView.text)
-        textViewDidChange()
```
SD10, line 706: 👍

Sources/Views/InputTextView.swift
```diff
@@ -198,6 +200,12 @@ open class InputTextView: UITextView {
         placeholderLabelConstraintSet?.right?.constant = -placeholderLabelInsets.right
     }
     
+    // MARK: - Notification
+    
+    private func postTextViewDidChangeNotification() {
+        NotificationCenter.default.post(name: .UITextViewTextDidChange, object: self)
```
SD10, line 206: Do you think we need to post a notification here? What about keeping the behavior consistent with UIKit and just disabling the button in the `didSet`? I'm not saying this is the right approach. I want your opinion. cc @nathantannar4 please check this when you get a chance

zhongwuzw: Emm, I think it needs. because besides adjust enabled of `sendButton`, we need also handle `placeholderLabel`,`messageInputBar`'s size, and so on.

nathantannar4:
Looks fine

SD10 (changes requested):
@zhongwuzw Care to add a CHANGELOG entry for this under bugfix?

Issue #362, opened 2017-11-28
Update ChatExample settings page to toggle MessageKit functionality


SD10:
There's currently an empty `SettingsViewController` in the example app. It would be nice if through this controller we could change/enable/disable functionality in the `ConversationViewController`.

This would increase my productivity 😃 because I would be able to test things without making code changes.

Things we could do:
- [x] Set number of mock messages to appear in `ConversationViewController`
- [x] Turn Text Messages **on / off**
- [x] Turn Attributed Text Messages **on / off**
- [x] Turn Photo Messages **on / off**
- [x] Turn Video Messages **on /off**
- [ ] Select enabled `DetectorTypes`

jaredStef:
I'd be glad to pick this up if anyone has yet to do so

SD10:
Feel free @jaredStef! It would be super helpful to me 😄

nathantannar4:
Additionally can you add to the settings page, avatar (on/off), colors (sender and receiver) :) 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
@nathantannar4 I don't think updating the colors of sender / receiver do any good? I'm really only interested in things that will help debug 😅 But there's a lot on this list, I'd be happy if someone even added any subset of them

alessioarsuffi:
I'm preparing a pull request with fix of Set number of mock messages to appear in ConversationViewController. :)

Update:

- [x] Set number of mock messages to appear in `ConversationViewController` from: [0.13.0](https://github.com/MessageKit/MessageKit/releases/tag/0.13.0)

hamzaozturk:
Things we can do in Settings now:
- [x] Set number of mock messages
- [x] Turn Text Messages **on / off**
- [x] Turn Attributed Text Messages **on / off**
- [x] Turn Photo Messages **on / off**
- [x] Turn Video Messages **on /off**
- [x] Turn Emoji Messages **on /off**
- [x] Turn Location Messages **on /off**
- [x] Turn Url Messages **on /off**
- [x] Turn Phone Messages **on /off**

PR #492, opened 2018-01-28
Feature/text messages settings

alessioarsuffi:
In settings now you can hide text, attributedText, photo and video messages.

it is related to #362 

Nothing

Any other comments?
-------------------
Let me know if everything is ok for you :)

Where has this been tested?
---------------------------
**Devices/Simulators:** 
iPhone 4s (iOS 9.0)
iPhone SE, 7, 7 plus, X (iOS 11)

**iOS Version:** 11.2

**Swift Version:** 4

### DIFF
--- a/Example/.idea/vcs.xml
+++ b/Example/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$/.." vcs="Git" />
+  </component>
+</project>
--- a/Example/.idea/vcs.xml
+++ b/Example/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$/.." vcs="Git" />
-  </component>
-</project>
--- a/Example/Sources/MockMessage.swift
+++ b/Example/Sources/MockMessage.swift
@@ -32,37 +32,38 @@ struct MockMessage: MessageType {
 	var sender: Sender
 	var sentDate: Date
 	var data: MessageData
+    var type: String
 	
-    init(data: MessageData, sender: Sender, messageId: String, date: Date) {
+    init(data: MessageData, sender: Sender, messageId: String, date: Date, type: String) {
 		self.data = data
 		self.sender = sender
 		self.messageId = messageId
 		self.sentDate = date
+        self.type = type
 	}
 	
     init(text: String, sender: Sender, messageId: String, date: Date) {
-        self.init(data: .text(text), sender: sender, messageId: messageId, date: date)
+        self.init(data: .text(text), sender: sender, messageId: messageId, date: date, type: "text")
 	}
 	
     init(attributedText: NSAttributedString, sender: Sender, messageId: String, date: Date) {
-        self.init(data: .attributedText(attributedText), sender: sender, messageId: messageId, date: date)
+        self.init(data: .attributedText(attributedText), sender: sender, messageId: messageId, date: date, type: "attributedText")
 	}
 
     init(image: UIImage, sender: Sender, messageId: String, date: Date) {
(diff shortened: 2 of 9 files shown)


Sources/Models/MessageData.swift
```diff
@@ -26,7 +26,7 @@ import Foundation
 import class CoreLocation.CLLocation
 
 /// An enum representing the kind of message and its underlying data.
-public enum MessageData {
+public enum MessageData: Equatable {
```
SD10: I don't think we should make `MessageData` conform to `Equatable` or `RawRepresentable`. Making changes to the example app should never require modifications in the core library

Sources/Models/MessageData.swift
```diff
@@ -62,4 +62,67 @@ public enum MessageData {
...
 //    case placeholder
 
+    // MARK: - Equatable
+
+    public static func ==(lhs: MessageData, rhs: MessageData) -> Bool {
+        switch (lhs, rhs) {
+        case (let .text(string1), let .text(string2)):
+            return true
+
+        case (let .attributedText(string1), let .attributedText(string2)):
```
SD10: For example, this is a very light check for equality. Ideally, we would have to compare the contents of each String. But I really want to remove these conformances all together 😅

Example/Tests/SampleDataTests.swift
```diff
@@ -0,0 +1,150 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import XCTest
+import MessageKit
+import MapKit
+@testable import ChatExample
+
+class SampleDataTests: XCTestCase {
```
SD10: Let's mark this as `final` so that we have static dispatch. 
TBH I don't think we need tests for the example app but I won't argue against them.

Example/Sources/Settings+UserDefaults.swift
```diff
@@ -24,21 +24,69 @@
 
 import Foundation
 
+let messagesKey = "mockMessages"
```
SD10, line 27: It would be great to make these static properties of a `MessageKeys` enum

Example/.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SD10: We could do without this file 😃

SD10 (changes requested):
@alessioarsuffi This is a great start but I think we have to work on a few things. I appreciate you taking the time to do this. This is what I was looking for 👍

alessioarsuffi:
I understand, I will change something based on your comments :)

Example/Sources/SampleData.swift
```diff
@@ -70,7 +72,7 @@ final class SampleData {
 
     var now = Date()
 
-    let messageTypes = ["Text", "Text", "Text", "AttributedText", "Photo", "Video", "Location", "Emoji"]
+    var messageTypes = ["text", "attributedText", "photo", "video", "location", "emoji"]
```
SD10: I think the easiest way to do this would be to populate this array based on the values from `NSUserDefaults`

alessioarsuffi: I made some changes, reverting all modifications to Core. Let me know :)

Example/Sources/MockMessage.swift
```diff
@@ -32,37 +32,38 @@ struct MockMessage: MessageType {
 	var sender: Sender
 	var sentDate: Date
 	var data: MessageData
+    var type: String
 	
-    init(data: MessageData, sender: Sender, messageId: String, date: Date) {
+    init(data: MessageData, sender: Sender, messageId: String, date: Date, type: String) {
```
SD10: @alessioarsuffi I think we're getting closer. Do we really need this property? Can't we just check user defaults and add / remove messages from the `messageTypes` array depending on the keys?

alessioarsuffi: This was helpful to use in tests, anyway, I've removed everything, let me know if it okay for you :)

SD10:
@alessioarsuffi Sorry I'm going to close this one because it went stale during the v1.0 development. We're going to have to rewrite the example app one day 😢 Thank you for taking the time to submit this PR 💯

PR #497, opened 2018-01-30
AutocompleteManager Plugin

nathantannar4:
This adds an optional `AutocompleteManager` that is a plugin to the `MessageInputBar`. With it, autocomplete can easily be added to look like the following:

simulator screen shot - iphone x - 2018-01-30 at 01 00 14

## Features 
- Autocomplete from any prefix
- Colored highlighting customizable to each prefix
- Easy to use

The Example has been updated to reflect using the `AutocompleteManager` .

I believe @cwalo has already been using this in a pre-release beta version I sent him.


Example/Sources/ConversationViewController.swift
```diff
@@ -33,6 +33,15 @@ class ConversationViewController: MessagesViewController {
     var messageList: [MockMessage] = []
     
     var isTyping = false
+    
+    /// The object that manages autocomplete
+    open lazy var autocompleteManager: AutocompleteManager = { [unowned self] in
```
SD10: I hate using unowned 😱

Sources/Protocols/AutocompleteManagerDataSource.swift
```diff
@@ -0,0 +1,71 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import UIKit
+
+/// AutocompleteManagerDataSource is a protocol that passes data to the AutocompleteManager
+public protocol AutocompleteManagerDataSource: class {
```
SD10: Let's use `AnyObject` protocol

SD10 (commented):
Great work @nathantannar4, I just have a few concerns:

- We're adding a lot of complexity here, a table view, more cells, more protocols
- We still haven't finished the v1.0 requirements
- Nothing is unit tested and I'd like to stop accepting untested features after 1.0. My hopes are after the layout API is refactored to be open, we can increase test coverage
- This is almost better suited as a potential plugin to the library after version 1.0. I think it's too early to decide

cwalo:
LGTM 👍 

Indeed, I've been using a beta version in production, with some modifications - specifically, handling spaces within autocomplete text (we don't use usernames and the organizations using our app are large, so first name alone isn't sufficient). @nathantannar4 let me know if you want that considered and I'll try to put together a PR for it. I was in a rush to get @mentions working all together (tokenization/detokenization), so it's in a hacky state at the moment.

I tend to agree with @SD10 that we should get MessageKit to 1.0 before we add anything else, though. Additionally, we might want to consider marrying this to a feature that parses @mentions in message text cells - this can be client/user specific though, so it might be difficult to handle every case (see above). The work required for our case was not pleasant, haha.

Sherlouk:
Looks pretty neat but I would've quite liked to have seen support for async responses to autocomplete similar to the way we handle it in [`MessageVC`](https://github.com/GitHawkApp/MessageViewController) as I can definitely see the need to use APIs to load certain information here.

Also not a huge fan of passing 3 parameter tuples around, doesn't look fairly nice and makes the function definitions lengthy and in places hard to read. I would recommend switching these out for a struct (or at bare minimum use a type alias)

Definitely re-iterating previous remarks though this is a pretty substantial change without being sufficiently tested (we've definitely seen our autocomplete fail a few times on GH, though we're starting to get some tests down)

cwalo:
@Sherlouk good call on the tuples.

nathantannar4:
Thanks for everyones feedback. I am working on:
1. No more tuples
2. A way to handle autocompletes that have spaces
3. More flexibility
4. Better API

nathantannar4:
Made some revisions

nathantannar4:
Ok this is ready for a final review

SD10:
Still has to be blocked by MessageKit 1.0 😕

SD10:
Closing this one

PR #498, opened 2018-01-31
Add .custom(Any) case to MessageData

### LINKED ISSUE #479: Add .custom(Any) case to MessageData
We should add a `.custom(Any)` case to MessageData so that users can use this as an access point for custom cell content.

I would assign myself but I can’t on GitHawk

SD10:
This resolves #479 and provides an override point for creating custom cells

### TODO:
- [x] Fix tests
- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,27 +8,30 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Added
 
-Adds the following methods to `MessagesCollectionViewFlowLayout`:
-- `messageLabelInsets(for message: MessageType, at indexPath: IndexPath) -> UIEdgeInsets`
-- `messageContainerPadding(for message: MessageType, at indexPath: IndexPath) -> UIEdgeInsets`
-- `messageContainerMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat`
-- `messageContainerSize(for message: MessageType, at indexPath: IndexPath) -> CGSize`
-- `cellTopLabelAlignment(for message: MessageType, at indexPath: IndexPath) -> LabelAlignment`
-- `cellTopLabelSize(for message: MessageType, at indexPath: IndexPath) -> CGSize`
-- `cellTopLabelMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat`
-- `cellBottomLabelAlignment(for message: MessageType, at indexPath: IndexPath) -> LabelAlignment`
-- `cellBottomLabelSize(for message: MessageType, at indexPath: IndexPath) -> CGSize`
-- `cellBottomLabelMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat`
-- `avatarPosition(for message: MessageType, at indexPath: IndexPath) -> AvatarPosition`
-- `avatarSize(for message: MessageType, at indexPath: IndexPath) -> CGSize`
-- `cellContentHeight(for message: MessageType, at indexPath: IndexPath) -> CGFloat`
+- **Breaking Change** Adds `.custom(Any?)` case to `MessageData`.
+[#498](https://github.com/MessageKit/MessageKit/pull/498) by [@SD10](https://github.com/SD10).
+
+- Adds the following methods to `MessagesCollectionViewFlowLayout`:
+  - `messageLabelInsets(for message: MessageType, at indexPath: IndexPath) -> UIEdgeInsets`
+  - `messageContainerPadding(for message: MessageType, at indexPath: IndexPath) -> UIEdgeInsets`
+  - `messageContainerMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat`
+  - `messageContainerSize(for message: MessageType, at indexPath: IndexPath) -> CGSize`
+  - `cellTopLabelAlignment(for message: MessageType, at indexPath: IndexPath) -> LabelAlignment`
+  - `cellTopLabelSize(for message: MessageType, at indexPath: IndexPath) -> CGSize`
+  - `cellTopLabelMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat`
+  - `cellBottomLabelAlignment(for message: MessageType, at indexPath: IndexPath) -> LabelAlignment`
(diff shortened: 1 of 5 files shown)


nathantannar4 (approved):

cwalo (approved):

zhongwuzw (approved):

Sherlouk:
Feels dangerous to add a type to a project where the exact intention is it'll immediately crash?

Am I missing something?

SD10:
@Sherlouk It's to allow people to make custom cells, basically if `messageForItem(at: IndexPath)` returns a `MessageType` with `MessageData.custom(Any?)` I expect them to handle it by overriding `cellForItem(at:)` and any other sizing methods and then calling `super` for any other types

Sherlouk:
Is that documented somewhere? `If you use the Custom MessageType then you are expected to override the ...`

SD10:
`fatalError` documentation, best documentation

Sherlouk:
Nah that just tells you that you fucked up! Not how to fix it!

ash30:
This would be super useful so I gave the changes a quick go locally, but hit a snag :sweat_smile:: `Declarations from extensions cannot be overridden yet` when trying to subclass `MessagesCollectionViewFlowLayout` and overriding `messageContainerSize()`. I can see its still in progress, so just an FYI :slightly_smiling_face:

SD10:
Yeah @ash30 this will require the next release, MessageKit 0.14.0, where we already made this changes   

stephtelolahy:
Can you show a example of implementation of .custom(Any)

SD10:
@stephtelolahy We'll definitely have some example documentation but I can't get to that right now. You can do a lot with `.custom(Any)`. You basically just need to handle `cellForItem(at: IndexPath)` in `MessagesViewController` and `messageContainerSize(for:at:)` in `MessagesCollectionViewFlowLayout`.

vmalyasov:
@SD10 I tried to override ```messageContainerSize(for:at:)``` using the subclass of ```MessagesCollectionViewFlowLayout``` but I got ```Declarations from extensions cannot be overridden yet``` error. Apparently, I don't have enough knowledge about how to override this function. Could you  explain how I can do this?

zhongwuzw:
@SD10 , Oops , we can only put non-override method in extentions.

SD10:
Wow that is surprising because I'm allowed to mark them as `open` 😅 We can fix this

yokomizor:
Suggestion:

```
open func messageContainerSize(for message: MessageType, at indexPath: IndexPath) -> CGSize {
  // ...
  case .custom:
    return messageContainerSizeForCustom()
  // ...
}

@objc open func messageContainerSizeForCustom() -> CGSize {
   fatalError(MessageKitError.customDataUnresolvedSize)
}
```

This way we could add our custom behavior by overriding ```messageContainerSizeForCustom``` without touching other MessageTypes.

By marking as @objc you would solve the non-override problem I guess.

Same thing could work for ```collectionView(_:cellForItemAt indexPath: IndexPath) -> UICollectionViewCell```.

wdyt?

SD10:
@yokomizor That's an interesting suggestion. Why not just override and call super though?

yokomizor:
@SD10 calling super will check MessageType again, and throw the fatalError.

If you remove the ```case .custom```, then calling super would be fine, but we will have to check ```message.data``` twice. Super will check first. Later our override will check it again (we just want to add custom behavior for ```.custom```).

Maybe it would be even better if you delegate this to ```messagesLayoutDelegate.messageContainerSize(...)```.

wdyt?

SD10:
@yokomizor I was expecting users to `override`, handle the `.custom` case. And then call `super` afterwards to handle the remaining cases.

SD10:
@yokomizor Also keep in mind, you really should not be using `development`. I will break a ton of things on this branch all the time. I may revert all these changes to the layout until a later release.

yokomizor:
@SD10 Oh! true. We don't need messageContainerSizeForDefault.

yokomizor:
How do you expect users will set a custom collectionViewLayout?

I suppose users could pass the custom one through init:
https://github.com/MessageKit/MessageKit/blob/880cf8c67d2a5f2048d73c842bd60af7d74e903e/Sources/Views/MessagesCollectionView.swift#L62

To do so, they need to set messagesCollectionView again on MessagesViewController since the default one is created on init phase:
https://github.com/MessageKit/MessageKit/blob/880cf8c67d2a5f2048d73c842bd60af7d74e903e/Sources/Controllers/MessagesViewController.swift#L33

So, the default instance will be created anyway, even if users would not use it.

I understand this is wip. Anyway, it would be nice to know which path you are planning to follow.

btw, wdyt about the delegate suggestion? Users are used to override default behavior using ```MessagesDisplayDelegate```. It might be less disruptive.

Cheers!

yokomizor:
@SD10 btw, back to the override topic. Even calling super later, MessageType will be checked twice. First to check if it is ```.custom``` and then later super will check it again. Using ```messageContainerSizeForDefault ``` MessageType will be checked just once. Am I missing something?

zhongwuzw:
@yokomizor , If you want to use custom layout, I think override `messagesCollectionView` and implement its `set`、`get` method would be a good way.

SD10:
@yokomizor You would return before super is called in the case of `.custom`. The switch will not be executed a second time

yokomizor:
@SD10 for ```.custom``` just once. for ```.text``` twice.

yokomizor:
@zhongwuzw cool. So delegates won't be an option? Less disruptive maybe 👼

SD10:
@yokomizor 
```Swift
override func messageContainerSize(for message: MessageType, at indexPath: IndexPath) -> CGSize {
    switch message.data {
   case .custom(_): 
      // handle sizing
   default: 
      super.messageContainerSize(for: message, at: indexPath)
   }
}
```
No double checks

yokomizor:
@SD10 

```message = .text(...)```

1. ```switch message.data``` will check message type once. Then ```default:``` case will be called.

2. Inside super, ```switch message.data``` will check message type again. Then ```case .text(_):``` will be called.

My point is that we could check it just once in order to address it to the right handler. Makes sense? Calling ```switch``` twice would't be a problem, though.

SD10:
It shouldn’t make a difference in production, especially since a switch will exit at the first match. It is a good point to bring up though 👍👍

PR #502, opened 2018-02-01
MessagesCollectionViewFlowLayout refactor

zhongwuzw:
- [ ] Remove currentLayoutContext && set cache enabled default
- [ ] Subclass `NSCache`
- [ ] Optimize attributes cache strategy


Sources/Layout/MessagesCollectionViewFlowLayout+MessageContainerView.swift
```diff
@@ -119,38 +119,18 @@ extension MessagesCollectionViewFlowLayout {
...
-            let padding = messageContainerPadding(for: message, at: indexPath)
-            currentLayoutContext.messageContainerPadding = padding
-            return padding
-        }
-        return cachedPadding
+        return messageContainerPadding(for: message, at: indexPath)
     }
 
     internal func _messageContainerMaxWidth(for message: MessageType, at indexPath: IndexPath) -> CGFloat {
-        guard let cachedMaxWidth = currentLayoutContext.messageContainerMaxWidth else {
```
SD10: We had all these internal methods to not repeat calculations. We may need `messageContainerMaxWidth` in `messageContainerSize`, `cellTopLabelSize`, `cellBottomLabelSize`, so we would need to call this 3 times

SD10: That's why I put the `currentLayoutContext` at the class scope to share the values

zhongwuzw: `currentLayoutContext` only assign in `newCellLayoutContext(for message: MessageType, at indexPath: IndexPath)`, I don't know wether would leads to chaos when layout. I can change these internal method that only calculate when not in cache.

SD10: Ugh, that is a good point 👍 I remember thinking about this. I can't be sure they will always be available in the property. I would need to know more about the order of `UICollectionView` method calls.  Better to be safe...

The cache will be empty because we only store it after completing all the calculations

zhongwuzw: @SD10 , ping 😄 .

Sources/Layout/MessagesCollectionViewFlowLayout+Avatar.swift
```diff
@@ -65,22 +65,17 @@ extension MessagesCollectionViewFlowLayout {
...
 extension MessagesCollectionViewFlowLayout {
 
-    internal func _avatarPosition(for message: MessageType, at indexPath: IndexPath) -> AvatarPosition {
-
-        guard let cachedPosition = currentLayoutContext.avatarPosition else {
-            let position = avatarPosition(for: message, at: indexPath)
-            currentLayoutContext.avatarPosition = position
-            return position
+    internal func _avatarPosition(for message: MessageType, at indexPath: IndexPath, _ cache: Bool = true) -> AvatarPosition {
+        if cache, let cachedPosition = layoutContextCache.object(forKey: message.messageId as NSString)?.avatarPosition {
```
SD10: Smart to add the `cache` property 👍 👏

Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -237,7 +237,7 @@ public extension MessagesLayoutDelegate {
     }
 
     func shouldCacheLayoutAttributes(for message: MessageType) -> Bool {
-        return false
+        return true
```
SD10, line 240: I just have to check this tomorrow. We had it as `false` by default because the caching of `cellTopLabel` and `cellBottomLabel` was annoying. If you had an empty label the size would get cached as `.zero`. Then if you changed the text to another value, you would get a non-zero size. But the cache was not invalidated 😅 Maybe the new equality checking will fix this

SD10: If we can solve #504 this delegate method becomes obsolete. There is no reason to allow users to control caching in this manner. This was a hack to turn it off while the label sizes were being cached and the caching strategy was buggy. If we invalidate properly the user doesn't even need to be aware of this.

SD10:
@zhongwuzw Maybe we should just do this:
a) turn caching off by default
b) allow users to turn it on, but make them responsible for clearing cache 😅 

The users will understand how dynamic their layout is

SD10:
We're going to take a different approach here. Thanks for all your work @zhongwuzw 🙏

PR #507, opened 2018-02-03
Added PrepareReuse override for MediaMessageCell

### LINKED ISSUE #506: MessageCollectionViewCell duplication (with video content) when scrolling
## General Information

* MessageKit Version: 0.13.1

* iOS Version(s): 11

* Swift Version: 4

* Devices/Simulators: iPhone 8 plus

##  What happened?
I'm showing videos in MessageCollectionViewCell when a video cell play button is tapped. I have made sure that only one instance of avcontroller is used across the conversation, so if a new video cell is clicked while playing one, playing once will be closed and new one will play. So it works alright.
**Problem is that** sometimes, when we scroll like a page or so while video is paused or being played, another video cell would have the exact same video at the playing at the same time. Now, it cannot be an issue with my implementation of video playing, because
1. My code starts only after tapping a cell
2. All video cells doesn't appear so, only on a cell comes in after 'complete page scroll'
So, clearly cells are duplicating, it might be an effect of dequeuing.

Video cells appear as thumbnails and when clicked it should play the video, the thumbnails should be up for all video cells unless clicked after scrolling.

mohzameer:
Fixes the cell duplicating issue with videos in media cells, now in MediaMessageCell, I have implemented the override prepareForReuse to control the dequeuing, and inside the override function, I'm removing all the views and adding image and the triangle button again with constraints, because when I'm adding the video on didTapMessage delegate in MessageCellDelegate I'm adding it as a subview and I want the subviews gone on scrolling.

Yes, closes #506

Any other comments?
-------------------
I tested and it was all good, but if it affects anywhere else, let me know.

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 8 plus

**iOS Version:** 11

**Swift Version:** 4

### DIFF
--- a/Sources/Views/Cells/MediaMessageCell.swift
+++ b/Sources/Views/Cells/MediaMessageCell.swift
@@ -52,6 +52,14 @@ open class MediaMessageCell: MessageCollectionViewCell {
         setupConstraints()
     }
 
+    open override func prepareForReuse() {
+        super.prepareForReuse()
+        messageContainerView.subviews.forEach({ $0.removeFromSuperview() })
+        messageContainerView.addSubview(imageView)
+        messageContainerView.addSubview(playButtonView)
+        setupConstraints()
+    }
+    
     open override func configure(with message: MessageType, at indexPath: IndexPath, and messagesCollectionView: MessagesCollectionView) {
         super.configure(with: message, at: indexPath, and: messagesCollectionView)
         switch message.data {


nathantannar4:
Hmm, I’m not sure if we want to be removing and then re adding subviews. You should consider subclassing the cell and add the subviews you need.

SD10:
Yeah, this is definitely a fix that is more client specific than a problem with the core library. 

We definitely do not want to be adding and removing subviews.

This will end up removing a view that someone added in a subclass and wants to remain there.

mohzameer:
Ok thanks, I'll try subclassing. Sure I understand, it is not a core library problem. But I'm sure out of box, in this library cell duplication actually occurs.

SD10:
@zameerascentic What do you mean by cell duplication? Collection view cells are reused from a pool, so there may only be 5 cells for 1000 messages.

mohzameer:
I mean, you're resetting labels on prepareForReuse in MessageCollectionViewCell, but subviews attached to it may not be made nil. Not just a video view, if it's a file preview or say any views we add as a subview inside cell to show contents might not refresh correctly?

SD10:
You shouldn’t really ever add/remove subviews like that from a cell. We don’t support a video cell with a player. Maybe in the long term 😕 If you want a more complex video cell you should subclass and add the view you require, then override cellForItem(at:IndexPath) to redirect all MessageData.video messages to use your custom cell, then call super. This will be easier in the next version of course

mohzameer:
Ohh ok, I get the idea and I will definitely do that, I'm kind if finding my way around MessageKit. Thanks for the guidance, will try this for sure.

PR #515, opened 2018-02-12
Add in-memory caching for bubble outline images

ndonald2:
This changed adds an in-memory `NSCache` for stretchable message bubble images. The way `MessageStyle` is currently implemented, a unique image will be loaded from the disk every time an outline image is requested, which is a significant performance issue since this happens every time a message cell is configured (see CPU trace screenshots below). The outline images are identical for each unique bubble style, there's no reason to reload them from disk every time. 

No, but it may eventually be made obsolete by PR #324 

These are screenshots of the CPU profiler instrument run on a basic integration of MessageKit where I loaded a page of messages and scrolled up and down for a bit.

**Top CPU Traces Before Change (system frameworks hidden)**

caching-before

Note that `UIImage.init(contentsOf:)` is the top trace

**Top CPU Traces After Change (system frameworks hidden)**

caching-after

Any other comments?
-------------------
Qualitatively, this appears to have a noticeable impact on scrolling and initial collection view rendering performance on a physical device.

Where has this been tested?
---------------------------

I've tested on a physical device and in a few simulators. Also the changes here include a unit test asserting image equality for the same `MessageStyle` which fails when run without the caching enhancement.

**Devices/Simulators:** iPhone 6 device, various simulators.

**iOS Version:** iOS 11.2, iOS 10.3

**Swift Version:** Swift 4.0 (should be compatible with earlier versions)

### DIFF
--- a/Sources/Models/MessageStyle.swift
+++ b/Sources/Models/MessageStyle.swift
@@ -102,6 +102,8 @@ public enum MessageStyle {
 
     // MARK: - Private
 
+    internal static let bubbleImageCache = NSCache<NSString, UIImage>()
+
     private var imageName: String? {
         switch self {
         case .bubble:
@@ -128,6 +130,4 @@ public enum MessageStyle {
         let capInsets = UIEdgeInsets(top: center.y, left: center.x, bottom: center.y, right: center.x)
         return image.resizableImage(withCapInsets: capInsets, resizingMode: .stretch)
     }
-
-    private static let bubbleImageCache = NSCache<NSString, UIImage>()
 }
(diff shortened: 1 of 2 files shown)


SD10 (approved):
@ndonald2 This is awesome. Thanks for taking the time to do this and put together such a detailed PR 👍

SD10:

Sources/Models/MessageStyle.swift
```diff
@@ -77,6 +77,12 @@ public enum MessageStyle {
 
         guard let path = imagePath else { return nil }
 
+        let cache = MessageStyle.bubbleImageCache
+
+        if let cachedImage = cache.object(forKey: path as NSString) {
```
zhongwuzw, line 82: We cannot just return the cached image in here, because the corner maybe different.

Issue #71, opened 2017-08-16
Data detectors for MessageLabel


SD10:
If we don't use `UITextView` for the `MessageLabel` we'll have to write the data detectors in-house. 

UITextView currently supports:
- [x] Phone Number
- [x] URL
- [x] Address
- [x] Date
- [ ] Shipment Tracking Number
- [ ] Flight Number
- [ ] Lookup Suggestion

It would be nice to also support:
- [ ] Email
- [ ] Hashtags
- [ ] Mentions
- [ ] Custom detector

ulaserdegor:
Location Type cell ? @SD10 
to open it in the map applications when clicked.

martinpucik:
Support for all of these types and custom regex one should be implemented now

PR #520, opened 2018-02-15
Enhancement/transit info detection

nosarj:
This PR adds transit information detection option to message labels

Contributes but does not complete Data detectors for MessageLabel #71

Where has this been tested?
---------------------------
**Devices/Simulators:** iphone 8

**iOS Version:** … 11.2

**Swift Version:** …4

### DIFF
--- a/Sources/Views/MessageLabel.swift
+++ b/Sources/Views/MessageLabel.swift
@@ -335,7 +335,7 @@ open class MessageLabel: UILabel {
                 rangesForDetectors.updateValue(ranges, forKey: .url)
             case .transitInformation:
                 var ranges = rangesForDetectors[.transitInformation] ?? []
-                let tuple: (NSRange, MessageTextCheckingType) = (result.range, .transitInformationComponents(result.components))
+                let tuple: (NSRange, MessageTextCheckingType) = (result.range, .transitInfoComponents(result.components))
                 ranges.append(tuple)
                 rangesForDetectors.updateValue(ranges, forKey: .transitInformation)
 
@@ -406,7 +406,7 @@ open class MessageLabel: UILabel {
         case let .link(url):
             guard let url = url else { return }
             handleURL(url)
-        case let .transitInformationComponents(transitInformation):
+        case let .transitInfoComponents(transitInformation):
             var transformedTransitInformation = [String: String]()
             guard let transitInformation = transitInformation else { return }
             transitInformation.forEach { (key, value) in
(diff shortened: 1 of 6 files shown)


Sources/Views/MessageLabel.swift
```diff
@@ -306,7 +315,7 @@ open class MessageLabel: UILabel {
             switch result.resultType {
             case .address:
                 var ranges = rangesForDetectors[.address] ?? []
-                let tuple: (NSRange, MessageTextCheckingType) = (result.range, .addressComponents(result.addressComponents))
+                let tuple: (NSRange, MessageTextCheckingType) = (result.range, .addressComponents(result.components))
```
SD10: Why did we change `.addressComponents` here?

Sources/Views/MessageLabel.swift
```diff
@@ -410,11 +432,16 @@ open class MessageLabel: UILabel {
         delegate?.didSelectPhoneNumber(phoneNumber)
     }
     
+    private func handleTransitInformation(_ transitInformationComponents: [String: String]) {
```
SD10: Maybe change the parameter name `transitInformationComponents` -> `components`

Sources/Views/MessageLabel.swift
```diff
@@ -410,11 +432,16 @@ open class MessageLabel: UILabel {
...
+    }
+    
 }
 
 private enum MessageTextCheckingType {
     case addressComponents([NSTextCheckingKey: String]?)
     case date(Date?)
     case phoneNumber(String?)
     case link(URL?)
+    case transitInformationComponents([NSTextCheckingKey: String]?)
```
SD10: We could just use `transitInfoComponents`, shorter 😅

SD10 (changes requested):
Hey @nosarj 👋,
Thanks for making this PR 😄 If you could address the requested changes this LGTM. We just need to add a CHANGELOG entry under **Added** for the new `DetectorType` and list this as a breaking change.

Sources/Views/MessageLabel.swift
nosarj: It has been marked as "// Deprecated in favor of components" so I updated it

SD10: Looks good to me. I don't see this as deprecated in the Foundation docs. Just need to test that it works before merging 👍

zhongwuzw: The comments not documented in `Apple API Reference`, so please make sure this is available in `iOS9` - `iOS11`.(IMO, maybe we can do this change after documented in `API Reference` 😄 )

nosarj: happy to take it out if you prefer, but prob a good one to remember for when it does go

SD10 (approved):
thanks @nosarj

SD10:

PR #531, opened 2018-02-20
Fix wrong bubble image cache

zhongwuzw:
Fix #515

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -100,6 +100,7 @@ open class MessagesViewController: UIViewController {
         removeKeyboardObservers()
         removeMenuControllerObservers()
         removeObservers()
+        clearMemoryCache()
     }
 
     // MARK: - Methods [Private]
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -81,8 +81,9 @@ open class MessagesViewController: UIViewController {
         registerReusableViews()
         setupDelegates()
         addMenuControllerObservers()
+        addObservers()
     }
-
+    
     open override func viewDidLayoutSubviews() {
         // Hack to prevent animation of the contentInset after viewDidAppear
         if isFirstLayout {
@@ -98,6 +99,7 @@ open class MessagesViewController: UIViewController {
     deinit {
         removeKeyboardObservers()
         removeMenuControllerObservers()
+        removeObservers()
     }
 
     // MARK: - Methods [Private]
@@ -148,5 +150,18 @@ open class MessagesViewController: UIViewController {
             let trailing = messagesCollectionView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
             NSLayoutConstraint.activate([top, bottom, trailing, leading])
         }
+    }
+    
+    private func addObservers() {
+        NotificationCenter.default.addObserver(
+            self, selector: #selector(clearMemoryCache), name: .UIApplicationDidReceiveMemoryWarning, object: nil)
+    }
+    
+    private func removeObservers() {
+        NotificationCenter.default.removeObserver(self, name: .UIApplicationDidReceiveMemoryWarning, object: nil)
+    }
+    
+    @objc private func clearMemoryCache() {
+        MessageStyle.bubbleImageCache.removeAllObjects()
     }
 }
--- a/Sources/Models/MessageStyle.swift
+++ b/Sources/Models/MessageStyle.swift
@@ -28,12 +28,12 @@ public enum MessageStyle {
 
     // MARK: - TailCorner
 
-    public enum TailCorner {
+    public enum TailCorner: String {
 
-        case topLeft
-        case bottomLeft
-        case topRight
-        case bottomRight
+        case topLeft = "topLeft"
+        case bottomLeft = "bottomLeft"
+        case topRight = "topRight"
+        case bottomRight = "bottomRight"
 
         var imageOrientation: UIImageOrientation {
             switch self {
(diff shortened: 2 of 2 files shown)


Sources/Controllers/MessagesViewController.swift
```diff
@@ -149,4 +151,17 @@ open class MessagesViewController: UIViewController {
             NSLayoutConstraint.activate([top, bottom, trailing, leading])
         }
     }
+    
+    private func addObservers() {
+        NotificationCenter.default.addObserver(
+            self, selector: #selector(clearMemoryCache), name: .UIApplicationDidReceiveMemoryWarning, object: nil)
```
ndonald2, line 158: This is unnecessary - `NSCache` automatically evicts items under memory pressure. [See documentation here](https://developer.apple.com/documentation/foundation/nscache)

> The NSCache class incorporates various auto-eviction policies, which ensure that a cache doesn’t use too much of the system’s memory. If memory is needed by other applications, these policies remove some items from the cache, minimizing its memory footprint.

Sources/Models/MessageStyle.swift
```diff
@@ -79,31 +79,40 @@ public enum MessageStyle {
...
-        }
-
-        guard var image = UIImage(contentsOfFile: path) else { return nil }
-
         switch self {
         case .none, .custom:
             return nil
         case .bubble, .bubbleOutline:
-            break
+            if let cachedImage = cache.object(forKey: path as NSString) {
```
ndonald2: Rather than breaking out cache miss logic per-case, why not add a `cacheKey` computed property that returns `path` or `path + corner.rawValue` depending on the case and use that? The only thing different between these cases is the cache key so it would be more terse/DRY that way.

Sources/Models/MessageStyle.swift
```diff
@@ -79,31 +79,40 @@ public enum MessageStyle {
...
-        let stretchedImage = stretch(image)
-        cache.setObject(stretchedImage, forKey: path as NSString)
-        return stretchedImage
     }
 
+    // MARK: - Internal
+    
+    internal static let bubbleImageCache: NSCache<NSString, UIImage> = {
+        let cache = NSCache<NSString, UIImage>()
+        cache.name = "com.messagekit.MessageKit.bubbleImageCache"
```
ndonald2, line 106: Good idea 👍

ndonald2 (commented):
Thanks for fixing this @zhongwuzw! Please see my suggestions.

SD10:
Thanks for catching this @zhongwuzw. I need sleep but I'll review this in the morning. Feel free to merge if you guys come across a solution 👌

Sources/Controllers/MessagesViewController.swift
zhongwuzw: @ndonald2 , we cannot rely on the eviction policy that `NSCache` provided. It would leads to crash in some situations when low memory. So we should remove objects when receive memory warnings(`NSCache` would not remove objects when receive memory warning notification).

ndonald2: @zhongwuzw Forgive my ignorance if that is true, but I've never heard that before and it seems to contradict the documentation. Can you provide proof? If true, wouldn't that defeat the purpose, i.e. wouldn't it imply that `NSCache` and a mutable `Dictionary` instance are effectively the same thing?

ndonald2: I see [this answer](https://stackoverflow.com/a/11521471) on stack overflow that corroborates what you're saying. Regardless, I don't think it's likely that these images are the sole factor that would contribute to an OOM crash but I suppose it's better to be safe than sorry.

That said, the cache is a static instance on `MessageType` so it doesn't make sense to me to tie the memory warning observation to `MessagesViewController` lifetime. What about adding it in the static initializer?

```swift
    internal static let bubbleImageCache: NSCache<NSString, UIImage> = {
        let cache = NSCache<NSString, UIImage>()
        cache.name = "com.messagekit.MessageKit.bubbleImageCache"
        let selector = #selector(cache.removeAllObjects)
        NotificationCenter.default.addObserver(cache, selector: selector, name: .UIApplicationDidReceiveMemoryWarning, object: nil)
        return cache
    }()
```

The cache has static lifetime so there's no reason to ever remove the observation.

Sources/Models/MessageStyle.swift
zhongwuzw: Good idea 👍 .

Sources/Models/MessageStyle.swift
```diff
@@ -28,12 +28,12 @@ public enum MessageStyle {
...
     // MARK: - TailCorner
 
-    public enum TailCorner {
+    public enum TailCorner: String {
 
-        case topLeft
-        case bottomLeft
-        case topRight
-        case bottomRight
+        case topLeft = "topLeft"
```
SD10: FYI this is not necessary. If you don't provide an explicit rawValue for `String` it will default to using the case name as a `String`

SD10 (dismissed):

Sources/Controllers/MessagesViewController.swift
zhongwuzw: @ndonald2 , Thanks for your advice 😃 , I prefer to keep my own way. The reason is if `MessageViewController` is out, we need to remove all cached objects to prevent useless memory consumption. If we adapt yours, it would leads to call `removeAllObjects` each time `App` receive memory warning(even `MessageViewController` `deinit`), it's unnecessary TBO.

Sources/Models/MessageStyle.swift
zhongwuzw: 👍

ndonald2 (approved):

Issue #504, opened 2018-02-02
Improve caching strategy


SD10:
After talking with @zhongwuzw we decided it would be good to change our caching strategy to an `IndexPath` based approach. 

This puts the emphasis on scroll performance and not the speed of the layout itself. We’ve already made improvements to the layout speed by using a context for storing previously calculated attributes.

The index path strategy requires that we invalidate any affected index paths after any reload, insertion, move, or deletion of cells in the collection view.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
This is being worked on in #502

SD10:
It's unfortunate, but we're still facing that issue resolving undefined behavior by allowing users to make layouts dynamic over an `IndexPath`. So this is taking longer than expected:

```swift
func avatarSize(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGSize {
    let isEven = indexPath.section % 2 == 0
    return isEven ? .zero : CGSize(width: 20, height: 20)
}
```

### What needs invalidation
We need layout invalidation for the following scenarios:
- invalidate all indexPaths after the location of an insertion from the middle 
- invalidate all indexPaths after the location of a deletion from the middle

### What does not need invalidation
If the change only occurs at the end of the indexPaths then the attributes are fine:
- inserting N new sections at the end
- deleting N sections from the end

zhongwuzw:
Because of the dynamic of layout, maybe we should remove all cached attributes except insert or delete cells at the end.

SD10:
Same thing I spent hours complaining to @cwalo @nathantannar4 months ago. This is proof that it is impossible to cache anything without undefined behavior:

```Swift
// This is an example of how a user could reasonably write code
// that would stack bubbles with a single avatar
func avatarSize(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGSize {
        if indexPath.section > 0 {
            let previousIndexPath = IndexPath(item: 0, section: indexPath.section - 1)
            let previousMessage = messageForItem(at: previousIndexPath, in: messagesCollectionView)
            let isLastMessage = lastMessage.sender == message.sender
            if isLastMessage { return .zero }
        }
        return CGSize(width: 30, height: 30)
}
```

You can write code that is dynamic over the current `IndexPath` or `MessageType`.
You can write code that is dynamic over a past or future `IndexPath` or `MessageType`.

[...]

Issue #519, opened 2018-02-14
Making cellTopLabel and cellBottomLabel have static sizes


SD10:
We can decrease the time spent laying out each cell by ~53% if we make the `cellTopLabel` and `cellBottomLabel` have static sizes. We're getting killed by sizing attributed text to fit 2x per cell.

This is kind of a non-backward compatible change for a few reasons:
- We would likely move the `LabelAlignment` enum and users would use a combination of `UIEdgeInsets` for text and `NSTextAlignment` to position the label. Like JSQMVC.

- We would probably have to remove `AvatarPosition.Vertical`'s `cellTop` and `cellBottom` cases
because the labels would have the constant width of the cell. Like JSQMVC.

- This means you wouldn't be able to align the top edges of an `AvatarView` and `cellTopLabel` or `cellBottomLabel`. Like JSQMVC.

More on this to come. This is what has development stale lately

hyouuu:
Given the performance hit I think this is a good move. Maybe we can have the flexibility to define different static sizes for different message types if there are needs for a few different sizes.

SD10:
Yeah I'm working on some stuff surrounding this @hyouuu

SD10:
You can see the PR for this @ #580. Should reduce layout times by potentially as much a 70% actually

SD10:
Resolved by #580

Issue #499, opened 2018-01-31
Restructure the MessageCollectionViewCell inheritance hierarchy


SD10:
We need to restructure the inheritance hierarchy of `MessageCollectionViewCell` if we're going to add other types of cells. The reason is `MessageCollectionViewCell` currently stores the avatar, message container, and the top/bottom labels.

- `MessageCollectionViewCell` : `UICollectionViewCell`
This will be a bare bones cell

- `MessageContentCell` : `MessageCollectionViewCell`
This will store the avatar view, top/bottom label, and message container view

- `TextMessageCell`, `LocationMessageCell`, `MediaMessageCell` : `MessageContentCell`
These will store the specialized views

Issue #487, opened 2018-01-26
Move all cell subview origin calculations into MessagesCollectionViewCell


SD10:
Move all cell subview origin calculations back into `MessagesCollectionViewFlowLayout` so that they can be easily overriden.

SD10:
It may be better to move this back into the cell to be honest. Looking into this

SD10:
I think the following API will be better inside of `MessageCollectionViewCell`:

`open func layoutAvatarView(with attributes: MessagesCollectionViewLayoutAttributes)`
`open func layoutMessageContainer(with attributes: MessagesCollectionViewLayoutAttributes)`
`open func layoutBottomLabel(with attributes: MessagesCollectionViewLayoutAttributes)`
`open func layoutTopLabel(with attributes: MessagesCollectionViewLayoutAttributes)`

Then users can just `override` these and make them a no-op, even add their own constraints, should they choose. This also allows us to cache less information in the layout object.

SD10:
This is done in #491 and will be available in MessageKit 0.14.0

PR #555, opened 2018-03-08
[WIP] MessageKit 1.0 Staging Area

SD10:
MessageKit 1.0:

- Remove 1 message to 1 section limitation #586
- Remove MessagesLayoutDelegate methods that size cell contents #573
- Restructure the MessageCollectionViewCell inheritance hierarchy #499
- Fix MessagesCollectionViewLayoutAttributes equality checking #575
- Add method to MessagesDisplayDelegate for async images #588
- Making cellTopLabel and cellBottomLabel have static sizes #519
- Change LabelAlignment to use NSTextAlignment and offset #556
- Refactor MessagesCollectionViewFlowLayout to use MessageSizeCalculators #583
- Create PhotoItem protocol for the associated value for MessageData.photo #582
- Improve support around custom cells #396
- Enable creating custom cell types #453
- Add .custom(Any) case to MessageData #479
- Move all cell subview origin calculations into MessagesCollectionViewCell #487
- Renaming `MessageData` to `MessageKind` #107 
- Only apply textColor for .text case of MessageData #625
- Remove explicit header and footer restriction #599
- Refactor reusable view API to be section based #615
- Remove CollectionViewReusable protocol #595

Documentation:
- [ ] MessagesCollectionViewFlowLayout
- [ ] MesagesCollectionViewLayoutAttributes
- [ ] MessageSizeCalculator
- [ ] TextMessageSizeCalculator
- [ ] MediaMessageSizeCalculator
- [ ] LocationMessageSizeCalculator
- [ ] Avatar
- [ ] AvatarPosition
- [ ] DetectorType
- [ ] LabelAlignment
- [ ] LocationMessageSnapshotOptions
- [ ] MessageData
- [ ] MessageKitDateFormatter
- [ ] MessageStyle
- [ ] Sender
- [x] LocationMessageCell
- [x] MediaMessageCell
- [x] MessageCollectionViewCell
- [x] MessageContentCell
- [x] TextMessageCell
- [ ] MessageReusableView
- [ ] InsetLabel
- [ ] MessageContainerView
- [ ] MessageLabel
- [ ] MessagesCollectionView


eliburke:
@SD10 I only peripherally follow MK's development process, so excuse me if this was already suggested, but have you tried caching your height calculations instead of performing them with every request? 
I have a lightweight NSCache based cache that associates the cell's content object with its height and width: ``` func storedHeightForMessage(_ message: Message, cellWidth: CGFloat) -> CGFloat?```

As long as the supplied width is unchanged, the stored height is returned. If it changes during viewDidLoad/WillAppear or due to rotation or other resizing, I automatically recalculate and store the new width and height. 
I have NSAttributedStrings in my cells, but the calculations are infrequent and scrolling is quite zippy even with hundreds of messages.

SD10:
@eliburke Thank you for bringing this up! This is exactly what we're trying to accomplish.

The problem is that **the width is not static** and that is something that is killing the progress/development of the project. 

The problem is the available width the `messageContainer` is based on the width of the `AvatarView`and the padding around the message (see below: both of those things are dynamic.)

We retrieve the size of these things through delegate methods:
```Swift
func avatarSize(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGSize

func messagePadding(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGSize
```

In JSQMVC, these are instance properties of the layout object. So they are static/maintain state. Any changes to these properties and we can invalidate the layout. 

***This was a mistake I made when designing the MessageKit API. Everything is so dynamic it makes caching nearly impossible.***

The only way to fix these things is making non-backward compatible API changes. Any thoughts?

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -206,44 +473,26 @@ extension MessagesCollectionViewFlowLayout {
...
-        currentLayoutContext.topLabelMaxWidth = _cellTopLabelMaxWidth(for: message, at: indexPath)
-        currentLayoutContext.topLabelSize = _cellTopLabelSize(for: message, at: indexPath)
-        currentLayoutContext.bottomLabelAlignment = _cellBottomLabelAlignment(for: message, at: indexPath)
-        currentLayoutContext.bottomLabelMaxWidth = _cellBottomLabelMaxWidth(for: message, at: indexPath)
-        currentLayoutContext.bottomLabelSize = _cellBottomLabelSize(for: message, at: indexPath)
-        currentLayoutContext.itemHeight = _cellContentHeight(for: message, at: indexPath)
-        return currentLayoutContext
-    }
-}
+        guard let cachedSize = attributedTextSizeCache.object(forKey: attributedText.hash as NSNumber)?.cgSizeValue else {
```
SD10: This won't work yet. We may have seen the exact same text on another message with a different width

Sources/Supporting/MessageKit+Availability.swift
```diff
@@ -24,98 +24,40 @@
...
-extension MessagesViewController {
-    /// A Boolean value that determines whether the `MessagesCollectionView` scrolls to the
-    /// bottom on the view's first layout.
-    ///
-    /// The default value of this property is `false`.
-    @available(*, deprecated: 0.11.1, message: "Removed in MessageKit 0.11.1.")
-    open var scrollsToBottomOnFirstLayout: Bool {
-        fatalError("Fatal Error: scrollsToBottomOnFirstLayout is no longer supported")
+    @available(*, deprecated: 0.14.0, message: "Removed in MessageKit 0.14.0. MessageKit no longer caches cell attributes as it leads to undefined behavior.")
+    func shouldCacheLayoutAttributes(for message: MessageType) -> Bool {
```
SD10: We can probably add this method back now

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -45,31 +44,51 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
...
+    open var customMessageLayout: MessageCellLayout? {
+        didSet { customMessageAttributesCache.removeAllObjects() }
+    }
+
+    private let textMessageAttributesCache = NSCache<NSString, MessageIntermediateLayoutAttributes>()
+    private let attributedTextMessageAttributesCache = NSCache<NSString, MessageIntermediateLayoutAttributes>()
+    private let emojiMessageAttributesCache = NSCache<NSString, MessageIntermediateLayoutAttributes>()
+    private let photoMessageAttributesCache = NSCache<NSString, MessageIntermediateLayoutAttributes>()
+    private let videoMessageAttributesCache = NSCache<NSString, MessageIntermediateLayoutAttributes>()
+    private let locationMessageAttributesCache = NSCache<NSString, MessageIntermediateLayoutAttributes>()
```
SD10: Closer but still not right. We may only be able to handle text sizes 😕

SD10:
> I have a lightweight NSCache based cache that associates the cell's content object with its height and width: func storedHeightForMessage(_ message: Message, cellWidth: CGFloat) -> CGFloat?

@eliburke Can you share your implementation for this? I want to know how you're retrieving the size from the cache using the content + width. I guess for me to do something similar I'd have to combine the hash of the cell content + the hash of the cell width. I'm not sure if XORing the two values is safe enough here 🤔

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -45,31 +44,51 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
     open var messageLabelFont: UIFont {
```
SD10: It may be nice to move the `font` into the `TextMessageLayout` object.

Pros:
- Then attributed text would have a default font (see #559 and #563)
-  More intuitive API

Cons:
- Then users are required to set an emoji font, we lose the current 2x by default behavior

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -45,31 +44,51 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
...
-    /// Note: The default value of this property is 500.
-    open var attributesCacheMaxSize: Int = 500 {
-        didSet {
-            layoutContextCache.countLimit = attributesCacheMaxSize
-        }
+    // MARK: - Layout
+
+    open var textMessageLayout = TextMessageLayout() {
+        didSet { textMessageAttributesCache.removeAllObjects() }
+    }
```
SD10: Not sure if I want these to be open to be honest, I don't want users removing the cache clearing behavior

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -135,133 +237,326 @@ extension MessagesCollectionViewFlowLayout {
...
 
-        return attributes
+    // MARK: - Cell Bottom Label Sizing
 
+    /// Returns the size of the `cellBottomLabel` in a `MessageCollectionViewCell`
+    /// for a given `MessageType` and `IndexPath`.
+    /// - Parameters:
+    ///   - message: The `MessageType` object for the cell.
+    ///   - indexPath: The `IndexPath` for the cell.
+    open func cellBottomLabelSize(for message: MessageType, at indexPath: IndexPath) -> CGSize {
```
SD10: Not sure if I want this open yet

SD10:
@MessageKit/contributors @MessageKit/core-team it would be nice to get some eyes on this.

Essentially, we have moved the delegate methods from `MessagesLayoutDelegate` into properties of `MessagesCollectionViewFlowLayout`. This allows us to assume that the layout for a cell is static, if any changes happen to the layout configuration, we invalidate the cache using a didSet property observer.

**If you change the underlying content of a message, however, you will still have cache invalidation problems.** 

This could be:
- a) Modifying the `MessageData` of a `MessageType` ie) changing the text of a seen message
- b) Changing the `cellTopLabelText` and `cellBottomLabelText` for a cell 

In these cases, you would have to invalidate the layout attributes for the given `MessageType` using its `messageId`. 

This is a sort of "gotcha" but we would provide proper documentation and I think this is a reasonable, necessary, assumption.

Resolves #504, #573

eliburke:
@sd10 sorry I missed your message on this earlier. I'm not sure my actual implementation will shed much light as it is larded up with custom logic. But I can try and hit the high points, and maybe it will help with any part you haven't already built. Oh and ... go for breaking API changes. Short term pain, long term gain.

The cache indexes between a message ID and a content object: 
````     private var _cache = NSCache<NSNumber, MessageContent>() // NOTE: this is thread safe! ````
In my code, I nearly always go from IndexPath -> data source -> message object -> cache'd content object. Retrieving the content cache looks like: 

    func content(for message: Message) -> MessageContent {
        let cacheKey = NSNumber(value: message.messageId)
        if let messageMedia = _cache.object(forKey: cacheKey) {
            return messageMedia
        }
        var messageContent: MessageContent!
        if message.type == .message {
            // etc etc based on the message type
        } else {
             messageContent = UnknownMessageContent(message)
        }
        _cache.setObject(messageContent, forKey: cacheKey)
        return messageContent
     }

In ````heightForRowAt```` (or heightForCell) I then:

        let message = messages[indexPath.row] // get message
        let cellWidth = tableView.frame.width - AVATAR_WIDTH // determine width, which could vary per row
        // look for a stored height if the cellWidth is unchanged
        if let height = contentCache.storedHeightForMessage(message, cellWidth: cellWidth) {
            return height
        }
        // calculate and return a new stored height for the current width
        return contentCache.calculateHeightForMessage(message, cellWidth: cellWidth

The stored lookup function:

    func storedHeightForMessage(_ message: Message, cellWidth: CGFloat) -> CGFloat? {
        // if the width matches stored width, we have the correct height. return it
        let messageContent = content(for: message)
        if cellWidth == messageContent.cellWidth { return messageContent.cellHeight }
        return nil
    }

````calculateHeightForMessage```` calls determineHeight() for each different content type, and then saves the resulting value on the content cache object along with the width and any other cache-able values. For content that may require loading, it has variables and stub methods to fetching content. Fetching is triggered by the Table/Collection view's cellFor() call, with an embedded completion to reload the cell (if its currently displayed message is unchanged). Internally, when a fetch occurs that will change the rendering or size of the content, it just clears the stored width so that the next calls to heightFor will re-generate. 

I hope this is in some way helpful. I know it will be hard to translate to a more dynamic layout design, but I think if you are able to cache based on the message rather than the current width, it may get you further down the field towards the end zone.

**Comment by SD10** [comment]
@eliburke Thanks a lot for writing that up. You helped confirm that there is really little that can be done with the current dynamism of MessageKit's layout. I may be able to achieve what you described above but it may be more work than is worth it. It looks like that system assumes that the message content and the width remains unchanged. By maintaining the width as a state of the `MessagesCollectionViewLayout` object, I can confirm that width has not changed. Then we can make the assumption that message content is static, if not, users can invalidate the layout for those attributes.

TL;DR thanks for confirming some of my suspicions about the direction of this PR.

File: Sources/Layout/MessagesCollectionViewFlowLayout.swift (hunk 5 above)
Comment (SD10) [reply Sources/Layout/MessagesCollectionViewFlowLayout.swift]: We could potentially put the properties of `TextMessageLayout` and the cache into a class called `TextMessageSizeCalculator` -- similar to JSQ's `MessageBubbleSizeCalculator`. Then we can also store underlying methods in here such as `messageContainerSize` etc.

File: Sources/Layout/MessagesCollectionViewFlowLayout.swift [hunk 7]
```diff
@@ -45,31 +44,51 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
...
+        didSet { videoMessageAttributesCache.removeAllObjects() }
+    }
+
+    open var locationMessageLayout = MediaMessageLayout() {
+        didSet { locationMessageAttributesCache.removeAllObjects() }
+    }
 
-    /// The `MessageCellLayoutContext` for the current cell.
-    internal var currentLayoutContext: MessageCellLayoutContext!
+    open var customMessageLayout: MessageCellLayout? {
```
zhongwuzw: We only support one kind of custom cell?

Sources/Layout/MessagesCollectionViewFlowLayout.swift
zhongwuzw: Why we define `cache` for each `messageData`? we use the `messageID` as the key, I missed something?

Sources/Layout/MessagesCollectionViewFlowLayout.swift
SD10: I agree. I need to think about this

Sources/Layout/MessagesCollectionViewFlowLayout.swift
SD10: Well, this way we only need to invalidate the messages that changed. If you change the layout for a `.photo` message it shouldn't affect a `.text` message

zhongwuzw: `.photo` and `.text` has the different `messageID`, I think it would not affect each other.

SD10: We can't just filter out all the objects in the cache that are `.photo` messages. If cache has `.text` messages and `.photo` messages. We have to remove everything. By separating them, we can invalidate all attributes for a given `MessageData` case at once.

SD10:
Jeez, this is going to be messy but I will probably have to use this PR as a staging area for 1.0. I'll merge everything into it and add CHANGELOG entries when I can, but the final result will have to be sifted for API breaking changes

SD10:
This looks like  good indicator of what MessageKit 1.0 beta will look like.

We'll need some @MessageKit/contributors to give this a review and feedback.

Sources/Controllers/MessagesViewController+Menu.swift
```diff
@@ -57,7 +57,7 @@ extension MessagesViewController {
 
         currentMenuController.setMenuVisible(false, animated: false)
 
-        guard let selectedCell = messagesCollectionView.cellForItem(at: selectedIndexPath) as? MessageCollectionViewCell else { return }
+        guard let selectedCell = messagesCollectionView.cellForItem(at: selectedIndexPath) as? MessageContentCell else { return }
```
cwalo, line 58: I'm wondering if optionally casting to `MessageContentCell` creates a limitation that prevents subclasses of `MessageCollectionViewCell` from intercepting `menuControllerWillShow` unless you register the observer twice with a different selector.

Sources/Controllers/MessagesViewController.swift
```diff
@@ -160,17 +160,14 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
...
     }
 
     open func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
         guard let collectionView = collectionView as? MessagesCollectionView else {
             fatalError(MessageKitError.notMessagesCollectionView)
         }
-        let messageCount = collectionView.messagesDataSource?.numberOfMessages(in: collectionView) ?? 0
-        // There will only ever be 1 message per section
-        return messageCount > 0 ? 1 : 0
+        return collectionView.messagesDataSource?.numberOfItems(inSection: section, in: collectionView) ?? 0
```
cwalo, line 162: Excited about this one!

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -0,0 +1,203 @@
...
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+
+open class MessageSizeCalculator: CellSizeCalculator {
+
+    public var layout: UICollectionViewFlowLayout?
```
cwalo: Is this just for convenience if the user subclasses and wants to access the layout?

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -0,0 +1,203 @@
...
+        case .cellTrailing, .cellLeading:
+            break
+        case .natural:
+            position.horizontal = isFromCurrentSender ? .cellTrailing : .cellLeading
+        }
+        return position
+    }
+
+    internal func avatarSize(for message: MessageType) -> CGSize {
+        let dataSource = messagesLayout.messagesDataSource
```
cwalo: Oh, I see it gets used here. The internal, potentially fatal, getter works for me.

Sources/Layout/LocationMessageSizeCalculator.swift
```diff
@@ -0,0 +1,43 @@
...
+        case .location(let item):
+            let maxWidth = messageContainerMaxWidth(for: message)
+            if maxWidth < item.size.width {
+                // Maintain the ratio if width is too great
+                let height = maxWidth * item.size.height / item.size.width
+                return CGSize(width: maxWidth, height: height)
+            }
+            return item.size
+        default:
+            fatalError("messageContainerSize received unhandled MessageDataType: \(message.data)")
```
SD10, line 40: This should go in the error namespace

Sources/Layout/MediaMessageSizeCalculator.swift
```diff
@@ -0,0 +1,48 @@
...
+            }
+            return item.size
+        }
+        switch message.data {
+        case .photo(let item):
+            return sizeForMediaItem(maxWidth, item)
+        case .video(let item):
+            return sizeForMediaItem(maxWidth, item)
+        default:
+            fatalError("messageContainerSize received unhandled MessageDataType: \(message.data)")
```
SD10, line 45: Namespace

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -0,0 +1,203 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+
+open class MessageSizeCalculator: CellSizeCalculator {
```
SD10, line 27: Need to document this class

Sources/Views/InsetLabel.swift
```diff
@@ -0,0 +1,38 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import UIKit
+
+open class InsetLabel: UILabel {
```
SD10: Needs documentation

Sources/Protocols/MediaItem.swift
```diff
@@ -24,25 +24,14 @@
...
 
-    var avatarSize: CGSize?
-    var avatarPosition: AvatarPosition?
+    var image: UIImage? { get set }
 
-    var messageContainerSize: CGSize?
-    var messageContainerMaxWidth: CGFloat?
-    var messageContainerPadding: UIEdgeInsets?
-    var messageLabelInsets: UIEdgeInsets?
+    var placeholderImage: UIImage { get set }
```
cwalo: I'm thinking it might be nice to provide a placeholder view here (e.g. activity indicator), but that can be saved for a feature PR.

Sources/Protocols/MessagesDataSource.swift
```diff
@@ -79,6 +88,10 @@ public extension MessagesDataSource {
         return message.sender == currentSender()
     }
 
+    func numberOfItems(inSection section: Int, in messagesCollectionView: MessagesCollectionView) -> Int {
+        return 1
```
cwalo, line 95: Pretty sure this crashes out of the box if the default is `1`.

cwalo (approved):
A few random thoughts, but no show-stoppers. LGTM 👍

Sources/Protocols/MessagesDataSource.swift
SD10: Good catch. I bet I know where this crashes :trollface:

Sources/Protocols/MediaItem.swift
SD10: Both you and @zhongwuzw have said that now. IMHO, I always thought that this was not a good practice. We would have a data model owning a view then. It results in initializing a view for each model and also modifying the view hierarchy in `cellForItem(at:)`.

Sources/Controllers/MessagesViewController+Menu.swift
SD10: Yeah, sounds like you're correct. I think it was a sloppy code migration. I'll fix this 👌

Sources/Layout/MessageSizeCalculator.swift
SD10: Yeah pretty much. I use it to get the width of the cell from the layout. Not necessary depending on where your cell is

Sources/Layout/MessageSizeCalculator.swift
SD10: What do you mean?

cwalo: Can disregard, made the first comment about the layout property and then saw it was actually _used_ here.

Sources/Protocols/MessagesDataSource.swift
SD10: I think it's related to the following code, need to fix these.

https://github.com/MessageKit/MessageKit/blob/1abcf9757678d3d70671906cae468820fe7a479b/Sources/Controllers/MessagesViewController.swift#L240

https://github.com/MessageKit/MessageKit/blob/1abcf9757678d3d70671906cae468820fe7a479b/Sources/Controllers/MessagesViewController.swift#L256

SD10: Looks like I have to remove the `MessageType` arguments from the `header` and `footer` delegate methods.

 No way to resolve the `IndexPath` inside this message body now that there is no longer a `1` to `1` relationship between sections and messages.

SD10: This is such a rabbit hole. Requires me to refactor everything related to headers and footers

SD10:
I'm going to merge this and start a new branch

PR #564, opened 2018-03-16
Fix cicleci

zhongwuzw:


SD10 (approved):
Good catch @zhongwuzw 👍

zhongwuzw:
@SD10 , next step is to fix the config.

Issue #562, opened 2018-03-15
Keyboard dismiss bug


AndriiPetrovDev:
## General Information

* MessageKit Version:
f48315c1a279a84e5cb41fb23e4cb8b86f60cb6a
latest development
* iOS Version(s):
11.2
* Swift Version:
Swift 3
* Devices/Simulators:
Any (iPhone 8 plus)
* Reproducible in ChatExample? (Yes/No):
yes

##  What happened?
When I doing swipe back to pop view controller interactively messageCollectionView scrolls down, but keyboard stays at old plase

## What did you expect to happen?
one of 2 option:
1. Both keyboard and collection view stays at their places
2. Both keyboard and collection view scrolls down
![sharing](https://user-images.githubusercontent.com/13622708/37487590-9e7ce3f4-289a-11e8-9c7c-f9fbbb7ca7a6.GIF)

SD10:
@AndrewPetrov Can you explain to me why this is a bug? Is it because it just doesn't look nice? I'm not sure what can be done here. The `MessageInputBar` is not part of the `MessagesViewController`. It is the `inputAccessoryView`.

AndriiPetrovDev:
Yes, It's not looking nice and I can't find a solution. JSQMessagesViewController behaves as I expect. And all known to me chats too (Viber, WatsAp, Slack)

SD10:
I'm not sure why the view controller dismisses like that. The only way to fix this would probably be to:
a) Resign the keyboard
b) Stop the view from shrinking downwards

nathantannar4:
Does the same behaviour occur when a UINavigationBar isn't using large titles?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

AndriiPetrovDev:
@zhongwuzw Can't understand did you fixed this issue?  Will it me pulled to main repo? or maybe it's already there?

SD10:
@AndrewPetrov The fix is #568. You can make that change locally if you'd like. I've just been holding off because I'm not sure if removing the hardware keyboard logic is the right solution. We're basically just switching bugs. This is low priority IMO

PR #568, opened 2018-03-17
Fix issue #562

zhongwuzw:
Calculate height to judge wether or not hardware keyboard is not accuracy. Ex:
When we operate like #562, swipe to pop the controller, the `(keyboardEndFrame.origin.y + keyboardEndFrame.size.height)` would greater than `UIScreen.main.bounds.height`.

So I just remove hardware keyboard detection, and keep the same logic handler. I don't known wether it would have problems in hardware mode.


SD10:
Closing this in favor of #707

PR #577, opened 2018-03-19
Fix wrong separated components in messageInputBar

### LINKED ISSUE #569: Hybrid multi-language message be seperated
## General Information

* MessageKit Version:0.13.1

* iOS Version(s): 11.2

* Swift Version:4.0

* Devices/Simulators:iPhone6sPlus

* Reproducible in ChatExample? (Yes/No):Yes

##  What happened?
When new message contains hybrid multi-language like Chinese + English, for example，"你好Jane", the message will be seperated in two messages: "你好" and "Jane"

The sent message should keep original one message.

zhongwuzw:
Fix #569

### DIFF
--- a/Sources/Views/InputTextView.swift
+++ b/Sources/Views/InputTextView.swift
@@ -303,26 +303,50 @@ open class InputTextView: UITextView {
     private func parseForComponents() -> [Any] {
         
         var components = [Any]()
-        let range = NSRange(location: 0, length: attributedText.length)
-        attributedText.enumerateAttributes(in: range, options: []) { (object, range, _) in
-            
-            if object.keys.contains(.attachment) {
-                if let attachment = object[.attachment] as? NSTextAttachment {
-                    if let image = attachment.image {
-                        components.append(image)
-                    } else if let image = attachment.image(forBounds: attachment.bounds,
-                                                           textContainer: nil,
-                                                           characterIndex: range.location) {
-                        components.append(image)
-                    }
-                }
-            } else {
-                let stringValue = attributedText.attributedSubstring(from: range).string.trimmingCharacters(in: .whitespacesAndNewlines)
-                if !stringValue.isEmpty {
-                    components.append(stringValue)
+        var attachments = [(NSRange, UIImage)]()
+        let length = attributedText.length
+        let range = NSRange(location: 0, length: length)
+        attributedText.enumerateAttribute(.attachment, in: range) { (object, range, _) in
+            if let attachment = object as? NSTextAttachment {
+                if let image = attachment.image {
+                    attachments.append((range, image))
+                } else if let image = attachment.image(forBounds: attachment.bounds,
+                                                       textContainer: nil,
+                                                       characterIndex: range.location) {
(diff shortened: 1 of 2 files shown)


SD10 (commented):
I'm good with this but I got lost on the second half of the logic. It may be good to wait and have @nathantannar4 review this as he is the original author.

Sources/Views/InputTextView.swift
```diff
@@ -303,26 +303,50 @@ open class InputTextView: UITextView {
...
+                    attachments.append((range, image))
+                } else if let image = attachment.image(forBounds: attachment.bounds,
+                                                       textContainer: nil,
+                                                       characterIndex: range.location) {
+                    attachments.append((range,image))
+                }
+            }
+        }
+        
+        var curLocation = 0
```
SD10, line 321: @zhongwuzw Can you help me understand the code from here? I just want to understand before we merge if Nathan doesn't have time to look at this one

zhongwuzw: @SD10 , First we should enumerate all `.attachment`, get its image and range, then we can split text and image based on the range.

Sources/Views/InputTextView.swift
```diff
@@ -303,26 +303,50 @@ open class InputTextView: UITextView {
...
+        if attachments.count == 0 {
```
zhongwuzw, line 322: @SD10 If no images, just text.

Sources/Views/InputTextView.swift
```diff
@@ -303,26 +303,50 @@ open class InputTextView: UITextView {
...
+            if !text.isEmpty {
+                components.append(text)
+            }
+        }
+        else {
+            attachments.forEach { (attachment) in
+                let (range, image) = attachment
+                if curLocation < range.location {
+                    let textRange = NSMakeRange(curLocation, range.location)
+                    let text = attributedText.attributedSubstring(from: textRange).string.trimmingCharacters(in: .whitespacesAndNewlines)
```
zhongwuzw, line 333: Add text and image to `components`

Sources/Views/InputTextView.swift
```diff
@@ -303,26 +303,50 @@ open class InputTextView: UITextView {
...
                 }
-            } else {
-                let stringValue = attributedText.attributedSubstring(from: range).string.trimmingCharacters(in: .whitespacesAndNewlines)
-                if !stringValue.isEmpty {
-                    components.append(stringValue)
+                
+                curLocation = range.location + range.length
+                components.append(image)
+            }
+            if curLocation < length - 1  {
```
zhongwuzw, line 342: After last `image`, maybe still has text, so we should append it to `components`.

SD10: Why is `length - 1` here? Sorry 😓

zhongwuzw: :smile: It's because we add a new line after image, so we decrease 1 to prevent unnecessary calculation.

SD10 (approved):
@zhongwuzw Thanks for walking me through this 👍

PR #579, opened 2018-03-21
Refactoring layout to use CellSizeCalculators

### LINKED ISSUE #583: Refactor MessagesCollectionViewFlowLayout to use MessageSizeCalculators
After #573, MessageKit will need a new architecture for sizing messages. Looking at things from first principles we can determine there are only 2 main override points for any given cell:
- Determining the size of the cell
- Determining which `UICollectionViewLayoutAttributes` to pass to the cell

The goal is to introduce a single object responsible for handling the configuration of a single cell:
```Swift
open class CellSizeCalculator {

    open weak var layout: UICollectionViewFlowLayout?

    public init(layout: UICollectionViewFlowLayout? = nil) {
        self.layout = layout
    }
    open func configure(attributes: UICollectionViewLayoutAttributes) {
        // No-op default
    }
    open func sizeForItem(at indexPath: IndexPath) -> CGSize {
        return .zero
    }
}
```

The one method that routes one of these objects when configuring the cell at a given `IndexPath`:
```Swift
open func cellSizeCalculatorForItem(at indexPath: IndexPath) -> MessageSizeCalculator { 
         // return any calculator here
}
```

Benefits:
- This allows users to handle sizing any type of `MessageCollectionViewCell`
- You can handle any number of `MessageData.custom` cases with any `MessageSizeCalculator`
- You can even handle other `MessageData` cases such as `.text` with your own calculators
- The object is loosely tied to the layout object (but not mandatory) so they can be used outside the `MessagesCollectionViewFlowLayout`.
- Increases testability

MessageKit will also provide more specific subclasses such as `MessageSizeCalculator` which can handle sizing the already provided cells. You can see everything described above in #579

SD10:
Resolves #583 

cc @zhongwuzw - this is what I was going on and on about during our chat 😂 

Note: 
- I haven't done anything related to caching the attributes yet. (keeping the example simple)
- ~Access control of the underlying methods hasn't been decided yet~


zhongwuzw:
@SD10 Bravo!!! The architecture is clearness, LGTM.

zhongwuzw:
Also user can provide their own CellSizeCalculators.

SD10:
> Also user can provide their own CellSizeCalculators.

Yeah that's the goal. So a user can mix any `MessageData` case with any `MessageSizeCalculator` and any `MessagesCollectionViewCell`.

Next step would be to add caching to our personal calculators. I will save that for another PR

Sources/Layout/MessageCellLayout.swift
```diff
@@ -63,23 +59,352 @@ public struct TextMessageLayout: MessageCellLayout {
...
+    public var incomingMediaSize: CGSize = .zero
+    public var outgoingMediaSize: CGSize = .zero
+
+    open override func messageContainerSize(for message: MessageType) -> CGSize {
+        let maxWidth = messageContainerMaxWidth(for: message)
+        switch message.data {
+        case .photo(_):
+            // TODO: - Implement actual sizing
+            return CGSize(width: maxWidth, height: 240)
+        case .video(_, _):
```
SD10: I need to figure out how we're going to handle sizing `.photo` and `.video`. I think JSQMVC uses static sizes? Need to check

zhongwuzw: It uses the `JSQMessageMediaData` protocol to handle this.
Also we still need to determine wether we need the hash key to be the key of the cache.

zhongwuzw: Another issue, we still only support one cell to one section? What's the point we not choose one section to all cells if it is?

SD10: > Another issue, we still only support one cell to one section? What's the point we not choose one section to all cells if it is?

I think we can remove this limitation but I would do this in a later PR. I don't want to make all these changes all at once. I'll still give it a try before merging anything to see if this PR makes that possible or impossible

zhongwuzw: 👍

PR #580, opened 2018-03-21
Stop sizing cellTopLabel and cellBottomLabel to fit

SD10:
TODO: Better PR description

Layout time of 10,000 messages (includes time to create mock messages)

Before:
screen shot 2018-03-21 at 3 18 37 am

After:
screen shot 2018-03-21 at 3 18 08 am

cc @zhongwuzw 

TODO:
- [x] CHANGELOG entry
- [x] More sensible defaults

### DIFF
--- a/Sources/Protocols/MessagesLayoutDelegate.swift
+++ b/Sources/Protocols/MessagesLayoutDelegate.swift
@@ -81,9 +81,26 @@ public protocol MessagesLayoutDelegate: AnyObject {
 
     @available(*, deprecated: 1.0, message: "heightForLocation(message:at:with:in) has been removed in MessageKit 1.0.")
     func heightForLocation(message: MessageType, at indexPath: IndexPath, with maxWidth: CGFloat, in messagesCollectionView: MessagesCollectionView) -> CGFloat
+
+    func cellTopLabelHeight(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGFloat
+
+    func cellBottomLabelHeight(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGFloat
+
 }
 
 public extension MessagesLayoutDelegate {
+
+    // MARK: - All Messages Defaults
+
+    func cellTopLabelHeight(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGFloat {
+        // TODO: - More sensible default, this is just for testing
+        return 10
+    }
+
+    func cellBottomLabelHeight(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGFloat {
+        // TODO: - More sensible default, this is just for testing
+        return 10
+    }
 
     func headerViewSize(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGSize {
         guard let displayDelegate = messagesCollectionView.messagesDisplayDelegate else {
--- a/Sources/Protocols/MessagesLayoutDelegate.swift
+++ b/Sources/Protocols/MessagesLayoutDelegate.swift
@@ -23,7 +23,6 @@
  */
 
 import Foundation
-import AVFoundation
 
 /// A protocol used by the `MessagesCollectionViewFlowLayout` object to determine
 /// the size and layout of a `MessageCollectionViewCell` and its contents.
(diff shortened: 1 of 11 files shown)


Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -103,12 +103,26 @@ public protocol MessagesLayoutDelegate: AnyObject {
     /// The default value returned by this method is `false`.
     func shouldCacheLayoutAttributes(for message: MessageType) -> Bool
 
+    func cellTopLabelHeight(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CGFloat
```
SD10: The reason why I wanted to do this PR first is that now we can't cache label sizes since they become dynamic.

This means we can probably only cache `MessageContainerView` sizes and not the full cell height. Doesn't matter though, this tradeoff is worth it 👌

Sources/Layout/MessageCellLayout.swift
```diff
@@ -53,11 +53,11 @@ open class MessageSizeCalculator: CellSizeCalculator {
...
 
-    public var incomingCellTopLabelAlignment = LabelAlignment.messageLeading(.zero)
-    public var outgoingCellTopLabelAlignment = LabelAlignment.messageTrailing(.zero)
+    public var incomingCellTopLabelAlignment = LabelAlignment(textAlignment: .left, textInsets: .zero)
+    public var outgoingCellTopLabelAlignment = LabelAlignment(textAlignment: .right, textInsets: .zero)
 
-    public var incomingCellBottomLabelAlignment = LabelAlignment.messageTrailing(.zero)
-    public var outgoingCellBottomLabelAlignment = LabelAlignment.messageLeading(.zero)
+    public var incomingCellBottomLabelAlignment = LabelAlignment(textAlignment: .right, textInsets: .zero)
+    public var outgoingCellBottomLabelAlignment = LabelAlignment(textAlignment: .left, textInsets: .zero)
```
SD10: Migration:
`LabelAlignment.cellCenter(.zero)` = `LabelAlignment(textAlignment: .center, textInsets: .zero)`
`LabelAlignment.cellLeading(.zero)` = `LabelAlignment(textAlignment: .left, textInsets: .zero)`
`LabelAlignment.cellTrailing(.zero)` = `LabelAlignment(textAlignment: .right, textInsets: .zero)`

There's no good replacement for:
`LabelAlignment.messageLeading(.zero)` && `LabelAlignment.messageTrailing(.zero)`

Since these need the `messageContainerView` frame a reference. You can work out all the cases where the alignment is closest to the `AvatarView` because you can use `avatarView.size.width` + `messagePadding.left` for example.

You could always make these computed properties or method that depends on the `messageContainerSize`. I don't think we'll end up caching anything but message container size (if anything at all after this PR) so it should be fair game.

SD10: I mean this is ok to be honest, most chat UI's aren't aligning things with the trailing edge of the bubble because you can't predict the bubbles width.

zhongwuzw: We don't need to provide `messageLeading`, user can adjust the `textInsets` to fit, it's easy to implement.

zhongwuzw: Closer to complete refactor! 👍

SD10: Yeah, considering this reduces layout time by 70% it makes caching even less important. Regardless, I still want to merge #579 because the architecture lets users handle custom cells easier and having a less dynamic layout will allow us to explore things like caching/other optimizations if we need to in the future.

SD10: > We don't need to provide messageLeading, user can adjust the textInsets to fit, it's easy to implement.

I agree. You can do message leading if the avatar is on the left by using `AvatarView.width` + `messagePadding.left` as the inset. 

You can't do this if the avatar is on the right because you don't know how wide the bubble will be. It could be short like `lol` or long like `The quick brown fox`. I think this is fine because users likely don't want to be aligning something this way because it is so volatile.

zhongwuzw: > Yeah, considering this reduces layout time by 70% it makes caching even less important. Regardless, I still want to merge #579 because the architecture lets users handle custom cells easier and having a less dynamic layout will allow us to explore things like caching/other optimizations if we need to in the future.

Yep, a big improvement, caching/other optimization is important when scroll, we can easily add these after these refactor IMO.

>  I think this is fine because users likely don't want to be aligning something this way because it is so volatile.

I agree totally.

PR #587, opened 2018-03-23
Add MediaItem protocol

### LINKED ISSUE #582: Create PhotoItem protocol for the associated value for MessageData.photo
After #579 we will probably provide the size for `.photo` messages in the `MessageData` enum.

```Swift
protocol PhotoItem {
   var url: URL? { get set }
   var image: UIImage? { get set }
   var placeholderImage: UIImage? { get set }
   var size: CGSize { get set }
}
```

This could also become a `MediaItem` protocol to be used for other message types.

SD10:
Resolves #582 

- TODO: 
- [x] CHANGELOG entry
- [x] Fix example
- [x] Fix tests

### DIFF
--- a/Sources/Layout/MessageCellLayout.swift
+++ b/Sources/Layout/MessageCellLayout.swift
@@ -372,19 +372,15 @@ open class TextMessageSizeCalculator: MessageSizeCalculator {
 
 open class MediaMessageSizeCalculator: MessageSizeCalculator {
 
-    // TODO: - Figure out how we are going to size these
-    public var incomingMediaSize: CGSize = .zero
-    public var outgoingMediaSize: CGSize = .zero
-
     open override func messageContainerSize(for message: MessageType) -> CGSize {
         let maxWidth = messageContainerMaxWidth(for: message)
         switch message.data {
-        case .photo(_):
-            // TODO: - Implement actual sizing
-            return CGSize(width: maxWidth, height: 240)
-        case .video(_, _):
-            // TODO: - Implement actual sizing
-            return CGSize(width: maxWidth, height: 240)
+        case .photo(let item):
+            let width = min(item.size.width, maxWidth)
+            return CGSize(width: width, height: item.size.height)
+        case .video(let item):
+            let width = min(item.size.width, maxWidth)
+            return CGSize(width: width, height: item.size.height)
         default:
             fatalError("messageContainerSize received unhandled MessageDataType: \(message.data)")
         }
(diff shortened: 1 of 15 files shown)


Sources/Layout/MessageCellLayout.swift
```diff
@@ -372,19 +372,15 @@ open class TextMessageSizeCalculator: MessageSizeCalculator {
...
         let maxWidth = messageContainerMaxWidth(for: message)
         switch message.data {
-        case .photo(_):
-            // TODO: - Implement actual sizing
-            return CGSize(width: maxWidth, height: 240)
-        case .video(_, _):
-            // TODO: - Implement actual sizing
-            return CGSize(width: maxWidth, height: 240)
+        case .photo(let item):
+            let width = min(item.size.width, maxWidth)
```
SD10: @zhongwuzw This may not be a good approach, because if users set the size inside the `MediaItem` protocol. They don't know how much space is available in the `MessageContainerView`. 

You can see that I use the `min` between the two, in case they provide a size too large... 😖 They also can't use the maxWidth to determine the height

zhongwuzw: Emm, I think we should keep the ratio of the image, just like:
`let size = maxWidth < item.size.width ? CGSize(width: maxWidth, height: maxWidth * item.size.height / item.size.width) : item.size`

SD10: I guess we can do that 🤔 If someone has a problem with this they can always subclass `MediaMessageSizeCalculator` and `override`.

zhongwuzw: Pretty much.

Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -125,22 +100,6 @@ public extension MessagesLayoutDelegate {
...
-
-    func widthForMedia(message: MessageType, at indexPath: IndexPath, with maxWidth: CGFloat, in messagesCollectionView: MessagesCollectionView) -> CGFloat {
-        return maxWidth
-    }
-
-    func heightForMedia(message: MessageType, at indexPath: IndexPath, with maxWidth: CGFloat, in messagesCollectionView: MessagesCollectionView) -> CGFloat {
-        switch message.data {
-        case .photo(let image), .video(_, let image):
-            let boundingRect = CGRect(origin: .zero, size: CGSize(width: maxWidth, height: .greatestFiniteMagnitude))
-            return AVMakeRect(aspectRatio: image.size, insideRect: boundingRect).height
```
SD10, line 138: @zhongwuzw The annoying thing is here is the default implementation. Now we're providing no default implementation. People are going to be confused where to get a width from I think 🤔

zhongwuzw: Emm, I think it's ok, most of time, the image is loaded async, so user should provide size ahead.

SD10: Not gonna lie, I'm still a little bothered that they have no frame of reference for the width. It just doesn't make sense.

PR #596, opened 2018-03-27
Refactor the cell inheritance hierarchy

### LINKED ISSUE #499: Restructure the MessageCollectionViewCell inheritance hierarchy

SD10:
Resolves #499 

TODO:
- [ ] CHANGELOG entry
- [x] Fix tests


Sources/Views/Headers & Footers/MessageHeaderView.swift
```diff
@@ -24,8 +24,11 @@
 
 import UIKit
 
-open class MessageHeaderView: UICollectionReusableView, CollectionViewReusable {
-    open class func reuseIdentifier() -> String { return "messagekit.header.base" }
+open class MessageHeaderView: MessageReusableView {
```
SD10, line 27: I'm wondering if we want to get rid of `MessageHeaderView` and just keep `MessageReusableView`?
I'll check if there is any reason why we can't do that. It makes it limiting if you want to use a header as a footer. I'm not sure if that should be a 1.0 change or not.

SD10: https://github.com/MessageKit/MessageKit/issues/599

PR #601, opened 2018-03-28
Remove the limitation of 1 cell to 1 section

### LINKED ISSUE #586: Remove 1 message to 1 section limitation
This was already done in an experimental branch by PR #537.

I'm pretty sure we can safely remove the 1 message to 1 section limitation in version 1.0. We just need to supply a delegate method for users to control the number of sections. Probably need to update the `numberOfMessages(for:at:in)` to take a `section: Int` argument as well.

Of course, we will leave the underlying logic the same so no one notices a difference.

Only way for me to determine if this is possible is to try and implement it after #579.

SD10:
Resolves #586 

TODO:
- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Added
 
+- **Breaking Change** Added new `MessagesDataSource` delegate methods, `numberOfSections(in)` and `numberOfItems(inSection:in)`.
+[#601](https://github.com/MessageKit/MessageKit/pull/601) by [@SD10](https://github.com/sd10), [@zhongwuzw](https://github.com/zhongwuzw).
+
 - **Breaking Change** Added new protocol `MediaItem` as the associated value for the
 `MessageData.video` and `MessageData.photo` cases.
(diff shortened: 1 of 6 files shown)


SD10:
So I had to add 2 more methods to accomplish this. We can't have the traditional API with one method to get the number of sections and one method to get the number of items in that section.

This is because it makes it impossible to put 1 message per section by default. And I don't want to break the default behavior of the library.

The reason why it makes it impossible is because we call `numberOfMessages` in `numberOfSections`. If `numberOfMessages` has a `section` parameter, then we run into a chicken and the egg problem.

So we have to have 3 methods:
- one to get the total number of messages
- one to get the total number of sections
- one to get the total number of cells in that section

zhongwuzw:
@SD10 🤔 , Let's come to a conclusion, we support one section to `N` `MessageType`s right? if it's true, I think `numberOfSections` and `numberOfItems` is efficient, am I right?

SD10:
```Swift
func numberOfMessages(in section: Int, in messagesCollectionView: MessagesCollectionView) -> Int {
    return 1
}

func numberOfSections(in messagesCollectionView: MessagesCollectionView) -> Int {
   return numberOfMessages(in: ) // where do I get the section number from?
}
```

@zhongwuzw See the problem above. That is why we need 3 methods. To leave the default behavior

PR #606, opened 2018-03-29
Support custom CellSizeCalculator not restrict to MessageSizeCalculator

zhongwuzw:


SD10:
Good catch ☝️ LGTM

PR #607, opened 2018-03-30
Make sizeCalculator lazy && fix access level

zhongwuzw:


SD10:
I’ll have to review this when I get back to the desktop @zhongwuzw. We shouldn’t mark the constraint class or any extensions as public. I believe good frameworks do not expose things that extend any part of Apple’s framework unless absolutely necessary 

SD10:
It can lead to complications when a user has the same extension as we do. Developers often use similar extensions to abstract common tasks. We should also turn on the SwiftLint rule for explicit access control

Sources/Extensions/NSAttributedString+Extensions.swift
```diff
@@ -26,15 +26,15 @@ import Foundation
 
 extension NSAttributedString {
 
-    func height(considering width: CGFloat) -> CGFloat {
+    public func height(considering width: CGFloat) -> CGFloat {
```
SD10: These should be internal

Sources/Extensions/UIColor+Extensions.swift
```diff
@@ -26,14 +26,14 @@ import Foundation
 
 extension UIColor {
 
-    static let incomingGray = UIColor(red: 230/255, green: 230/255, blue: 235/255, alpha: 1.0)
+    public static let incomingGray = UIColor(red: 230/255, green: 230/255, blue: 235/255, alpha: 1.0)
```
SD10: Debatable but let's leave these as internal for now

Sources/Extensions/UIView+Extensions.swift
```diff
@@ -26,7 +26,7 @@ import UIKit
 
 extension UIView {
     
-    func fillSuperview() {
+    public func fillSuperview() {
```
SD10: Internal for this class too 😓

Sources/Extensions/UIEdgeInsets+Extensions.swift
```diff
@@ -26,11 +26,11 @@ import Foundation
 
 extension UIEdgeInsets {
 
-    var vertical: CGFloat {
+    public var vertical: CGFloat {
```
SD10: Internal

Sources/Models/DetectorType.swift
```diff
@@ -38,7 +38,7 @@ public enum DetectorType {
     //case hashtag
     //case custom
 
-    var textCheckingType: NSTextCheckingResult.CheckingType {
+    public var textCheckingType: NSTextCheckingResult.CheckingType {
```
SD10: Internal would be good for this too. I may have some future plans that could affect this

Sources/Models/NSConstraintLayoutSet.swift
```diff
@@ -59,7 +59,7 @@ public class NSLayoutConstraintSet {
     ///
     /// - Returns: Self
     @discardableResult
-    func activate() -> Self {
+    public func activate() -> Self {
```
SD10: To be honest, this class should not be public. MessageKit is not a constraint library. What to do about this considering it accidentally got exposed?

zhongwuzw: No exposed currently, so let's make `internal`.

zhongwuzw:
@SD10 , I updated `swiftlint`'s config.

>It can lead to complications when a user has the same extension as we do. Developers often use similar extensions to abstract common tasks. We should also turn on the SwiftLint rule for explicit access control

Emm, I think it's not lead to complications, extension belongs to different module, it would work safely. 😂

SD10:
> Emm, I think it's not lead to complications, extension belongs to different module, it would work safely. 😂 

We’ve had complications at SwifterSwift with this problem. Module A extends UIKit with the same extension as Module B. The compiler can’t resolve which is which. I even tried things like `MessageKit.UIEdgeInsets` but even if that did work it would be very bad to work with. And in some situations you get a return type that is not prefixed by the module, ie. A UIKit method. So I basically never corrupt UIKit unless I’m forced to

Sources/Models/NSConstraintLayoutSet.swift
```diff
@@ -24,18 +24,18 @@
 
 import UIKit
 
-public class NSLayoutConstraintSet {
+internal class NSLayoutConstraintSet {
```
SD10, line 27: We just need to add an API breaking CHANGELOG entry under removed for this. Then we can merge

zhongwuzw: 👌

PR #611, opened 2018-03-31
Transform CellSizeCalculator from protocol to class

zhongwuzw:
`Protocols` can't require properties to be implemented as `weak` stored properties. See [Proposal 0186](https://github.com/apple/swift-evolution/blob/master/proposals/0186-remove-ownership-keyword-support-in-protocols.md)

Currently Swift/Xcode seems has a bug that not report this issue, `Xcode9.3` fix this, would post a warning that can't use `weak` property declared in `protocol`.


SD10:
Oh wow @zhongwuzw good catch. I thought it would be ok if I made it conform to `AnyObject` (but I forgot that too). 😭

SD10 (approved):

SD10 (approved):

PR #616, opened 2018-04-03
Make computed properties of MessagesCollectionViewFlowLayout public

cwalo:
This allows cellSizeCalculatorForItem to be overridden and still call super.


SD10 (changes requested):
Can you add documentation for these?
- "The `MessagesCollectionView` that owns this layout object."
- "The `MessagesDataSource` for the layout's collection view."
- "The `MessagesLayoutDelegate` for the layout's collection view."

And also move them out of the extension into the class body with the other properties

cwalo:
@SD10 updated

SD10 (approved):
:shipit:

PR #623, opened 2018-04-05
Fix font invalid for attributedString

### LINKED ISSUE #557: New attributedText message does not display correct until scroll from visible->invisible->visible
I modified the demo code like this:
____________________

```swift
extension ConversationViewController: MessageInputBarDelegate {

    func messageInputBar(_ inputBar: MessageInputBar, didPressSendButtonWith text: String) {
        
        // Each NSTextAttachment that contains an image will count as one empty character in the text: String
        
        for component in inputBar.inputTextView.components {
            
            if let image = component as? UIImage {
                
                let imageMessage = MockMessage(image: image, sender: currentSender(), messageId: UUID().uuidString, date: Date())
                messageList.append(imageMessage)
                messagesCollectionView.insertSections([messageList.count - 1])
                
            } else if let text = component as? String {
                
                let attributedText = NSAttributedString(string: text, attributes: [.font: UIFont.systemFont(ofSize: 15), .foregroundColor: UIColor.blue])
                
                let msg9String = "Use .attributedText() to add bold, italic, colored text and more..."
                let msg9Text = NSString(string: msg9String)
                let msg9AttributedText = NSMutableAttributedString(string: String(msg9Text))
                
                msg9AttributedText.addAttribute(NSAttributedStringKey.font, value: UIFont.preferredFont(forTextStyle: .body), range: NSRange(location: 0, length: msg9Text.length))
                msg9AttributedText.addAttributes([NSAttributedStringKey.font: UIFont.monospacedDigitSystemFont(ofSize: UIFont.systemFontSize, weight: UIFont.Weight.bold)], range: msg9Text.range(of: ".attributedText()"))
                msg9AttributedText.addAttributes([NSAttributedStringKey.font: UIFont.boldSystemFont(ofSize: UIFont.systemFontSize)], range: msg9Text.range(of: "bold"))
                msg9AttributedText.addAttributes([NSAttributedStringKey.font: UIFont.italicSystemFont(ofSize: UIFont.systemFontSize)], range: msg9Text.range(of: "italic"))
                msg9AttributedText.addAttributes([NSAttributedStringKey.foregroundColor: UIColor.red], range: msg9Text.range(of: "colored"))
                
                
                let message = MockMessage(attributedText: msg9AttributedText, sender: currentSender(), messageId: UUID().uuidString, date: Date())
                messageList.append(message)
                messagesCollectionView.insertSections([messageList.count - 1])
            }
            
        }
        
        inputBar.inputTextView.text = String()
        messagesCollectionView.scrollToBottom()
    }

}
```
____________________

So whatever I send, it will send a new attributedText message.
You can see the new attributedText message will not display correct.

But after you scroll the colletionview up and down, it may display correct.

zhongwuzw:
* Fixed #557 . 
* Added `Xcode 9.3` related config file.

- [x] Added CHANGELOG.

Another issue still exist, like `textColor`, it would leads to invalidation of `attributedString`'s `color` attributes.

We need to investigate that wether remove `textColor(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> UIColor`, it also leads to invalidation of `attributedString`'s `color` attributes.

### DIFF
--- a/Sources/Views/Cells/TextMessageCell.swift
+++ b/Sources/Views/Cells/TextMessageCell.swift
@@ -44,7 +44,7 @@ open class TextMessageCell: MessageCollectionViewCell {
         super.apply(layoutAttributes)
         if let attributes = layoutAttributes as? MessagesCollectionViewLayoutAttributes {
             messageLabel.textInsets = attributes.messageLabelInsets
-            messageLabel.font = attributes.messageLabelFont
+            messageLabel.messageLabelFont = attributes.messageLabelFont
             messageLabel.frame = messageContainerView.bounds
         }
     }
@@ -79,6 +79,7 @@ open class TextMessageCell: MessageCollectionViewCell {
             switch message.data {
             case .text(let text), .emoji(let text):
                 messageLabel.text = text
+                messageLabel.applyMessageLableFont()
             case .attributedText(let text):
                 messageLabel.attributedText = text
             default:
--- a/Sources/Views/Cells/TextMessageCell.swift
+++ b/Sources/Views/Cells/TextMessageCell.swift
@@ -79,7 +79,9 @@ open class TextMessageCell: MessageCollectionViewCell {
             switch message.data {
             case .text(let text), .emoji(let text):
                 messageLabel.text = text
-                messageLabel.applyMessageLableFont()
+                if let font = messageLabel.messageLabelFont {
+                    messageLabel.font = font
+                }
             case .attributedText(let text):
                 messageLabel.attributedText = text
             default:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Fixed
+
+- Fixed font invalidation of `attributedString` in `MessageLabel`.
+[#623](https://github.com/MessageKit/MessageKit/pull/623) by [@zhongwuzw](https://github.com/zhongwuzw).
+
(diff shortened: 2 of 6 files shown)


SD10:
@zhongwuzw I have to wrap my head around this one. I think the bug is caused [by this](https://github.com/MessageKit/MessageKit/blob/c476abe5050bc7c3d4f86edc157d8c63be9de9f4/Sources/Layout/MessagesCollectionViewFlowLayout.swift#L284)?

SD10:
@zhongwuzw In Slack we were discussing changing `textColor(for:at:in)` so that it is not used on `.attributedText` case but only `.text` case.

zhongwuzw:
> I think the bug is caused by this?

Nope, this is only apply when `attributedString` not set `font` attribute, otherwise, maybe we can remove this disposition.

> In Slack we were discussing changing textColor(for:at:in) so that it is not used on .attributedText case but only .text case.

👍 , it needs.

Sources/Views/Cells/TextMessageCell.swift
```diff
@@ -79,6 +79,7 @@ open class TextMessageCell: MessageCollectionViewCell {
             switch message.data {
             case .text(let text), .emoji(let text):
                 messageLabel.text = text
+                messageLabel.applyMessageLableFont()
```
SD10: Can't we just apply the `UIFont` inside of here without this method? I don't mind repeating code. This way of doing it makes it very confusing to me.

SD10 (approved):
@zhongwuzw Thanks for fixing this. It's a pretty bad bug. After CI passes I will merge and do a release

zhongwuzw:
@SD10 , 👌 , I'll commit a fix to `move-open` branch either.

PR #625, opened 2018-04-05
Only apply textColor for .text case of MessageData

### LINKED ISSUE #624: Change textColor method behavior on MessagesDisplayDelegate.
Currently, the `textColor(for:at:in)` message of `MessagesDisplayDelegate` applies to both the `.text(String)` and `.attributedText(NSAttributedString)` cases of `MessageData`.

This is a source of confusion when an`.attributedText` message's text color get's overridden by the delegate method. I think I originally set it up like this to avoid confusion but IMO this is not good behavior as it is still confusing and limits what can be done with `.attributedText`.

**TL;DR** we need to make `textColor(for:at:in)` only apply to `.text` messages.

cwalo:
Resolves #624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,8 @@ You must now set this font explicitly through the `emojiMessageSizeCalculator` o
 - Changed the `contentMode` of the `UIImageView` for `MediaMessageCell` to be `.scaleAspectFill`.
 [#587](https://github.com/MessageKit/MessageKit/pull/587) by [@SD10](https://github.com/sd10).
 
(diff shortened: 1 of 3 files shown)


SD10 (commented):
Needs a CHANGELOG entry under changed. Non-API breaking. Should be listed at bottom of changed section as we list API breaking changes first

cwalo:
@SD10 updated 👍

SD10:
@zhongwuzw if you agree with this please merge

zhongwuzw (approved):
💯

PR #630, opened 2018-04-06
Set comp mode to whole module

SD10:
This sets compilation mode to whole module and renames `flatMap` to `compactMap` for Swift 4.1


SD10:
I don't have the energy to deal with this tonight. Going to sleep 😴

zhongwuzw:
@SD10 , circleci still has error that report `swift compiler` error, it's because `whole module optimize`. 😂

SD10:
@zhongwuzw this is a nightmare, what are we going to do? 😂 I did not push 0.13.3 to CocoaPods. I set Swift 4.1 in the file so I couldn't push it because of compile error

zhongwuzw:
🤔 , we really need to set `swift` version to `4.1`? I think `compactMap` is not backward compatibility.

Can we revert `circleci`'s swift to `4.0` and `xcode` to `9.2.0` to test? Maybe it works, I don't know 🤔 .

SD10:
@zhongwuzw no, it is not backwards compatible. I have Xcode 9.3 on my machine so when I went to release `master` I got a compile error. We could build a shim? The problem is MessageKit 1.0 needs to support 4.1 🙄

zhongwuzw:
@SD10 Can we make these changes to `move-open` branch first? my part can merge to `master` safely.

PR #633, opened 2018-04-06
Fixed bubbleTailOutline of message bubble

### LINKED ISSUE #632: Buble filled white
* MessageKit Version: 0.13.2
* iOS Version(s): 11.3
* Swift Version: 4.1
* Devices/Simulators: iPhone 6s, iPhone 8s

##  What happened?

Buble filling when I try to set "bubbleTailOutline".

zhongwuzw:
Fixed #632 .

We cannot use `bubble_outlined_tail_v2` image as the `mask`, its opacity is 0.Instead use `bubbleTail` as the temporary substitution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Fixed
+
+- Fixed `bubbleTailOutline` invalidation of message bubble.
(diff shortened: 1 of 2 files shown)


SD10:
I really hate that we're using images for this reason.... it's such a hack.

zhongwuzw:
@SD10 , yep, we can optimize it after `1.0`.

cwalo (approved):
It is what it is. 👍

SD10 (approved):

PR #635, opened 2018-04-06
 Update access control

cwalo:
Necessary to subclass these classes

### DIFF
--- a/Sources/Layout/MessageSizeCalculator.swift
+++ b/Sources/Layout/MessageSizeCalculator.swift
@@ -107,7 +107,7 @@ open class MessageSizeCalculator: CellSizeCalculator {
 
     // MARK: - Avatar
 
-    open func avatarPosition(for message: MessageType) -> AvatarPosition {
+    public func avatarPosition(for message: MessageType) -> AvatarPosition {
         let dataSource = messagesLayout.messagesDataSource
         let isFromCurrentSender = dataSource.isFromCurrentSender(message: message)
         var position = isFromCurrentSender ? outgoingAvatarPosition : incomingAvatarPosition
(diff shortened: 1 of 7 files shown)


Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -107,7 +107,7 @@ open class MessageSizeCalculator: CellSizeCalculator {
 
     // MARK: - Avatar
 
-    internal func avatarPosition(for message: MessageType) -> AvatarPosition {
+    open func avatarPosition(for message: MessageType) -> AvatarPosition {
```
SD10: Not sure if I wanted these `open` because they're accessible as properties 🤔 cc @zhongwuzw

Sources/Layout/MessagesCollectionViewLayoutAttributes.swift
```diff
@@ -29,19 +29,19 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
 
     // MARK: - Properties
 
-    internal var avatarSize: CGSize = .zero
-    internal var avatarPosition = AvatarPosition(vertical: .cellBottom)
+    public var avatarSize: CGSize = .zero
```
SD10, line 32: These are fine though

SD10:
We also have to drop the first 3 commits

Sources/Layout/MessageSizeCalculator.swift
zhongwuzw: `internal` would be fine, we use `incoming*` and `outgoing*` properties to let layout static, if we make it `open`, it's back to the original dynamic way.

zhongwuzw:
Let's `squash and merge`.

Sources/Layout/MessageSizeCalculator.swift
SD10: Exactly what I was thinking 👍

cwalo: Ah, I see. I was thinking someone might want to alter the default behavior - seemed arbitrary not to allow that.

cwalo:
Apologies for not pulling in the upstream changes first.

cwalo:
Made methods public so that users can _at least_ call them from their subclasses. IMO it's a lot of boilerplate to have to rewrite.

SD10 (approved):

PR #641, opened 2018-04-10
Feature/custom view message cell

### LINKED ISSUE #640: Add support for any UIView as a message
Hello!

I think it'd be pretty useful to have an additional MessageData case to allow for any UIView to be displayed inside a message. This would allow us to build rich content messages without needing to have lots of setup code inside something like the messageStyle(...) DisplayDelegate function. Not sure what the progress on `case custom(Any)` is, or whether that's intended to work this way, but I was picturing something like `case customView(UIView)` which would pass all the heavy lifting into the view class itself.

Thanks for all the hard work with this lib 😃

JoshRobertsDev:
Resolves #640


SD10:
Hey @JoshRobertsDev, thanks for submitting this PR 👍

Unfortunately, I'm going to have to close this because this does not fit the architecture for MessageKit 1.0. You can see most work is actually happening on the `refactor/move-open` branch and not `development`. Feel free to join our Slack channel if you have other questions regarding contributing.

Please see [my comment in #640 for more information](https://github.com/MessageKit/MessageKit/issues/640#issuecomment-380154814).

PR #645, opened 2018-04-11
Fixed wrong bubble size calculation in some fonts

### LINKED ISSUE #639: Custom Font: message is not displayed completely when using custom font for long messages
## General Information

* iOS Version(s): 10+

* Swift Version: 4

* Reproducible in ChatExample? (Yes/No): yes, attached screenshot

##  What happened?

In Message Cell Attributed text is truncated not completely loading for long messages. Attached reference screenshot from Chat Example.
customfont

It should load complete message like loading for system fonts.

zhongwuzw:
Fixed #639.
`boundingRect` would return wrong size in some fonts, so add `usesFontLeading` option to use font leading for calculating line heights.


SD10 (approved):
LGTM 👍

PR #649, opened 2018-04-12
Reduce the number of calculation for label size

zhongwuzw:


Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -186,13 +186,10 @@ open class MessageSizeCalculator: CellSizeCalculator {
     }
 
     internal func labelSize(for attributedText: NSAttributedString, considering maxWidth: CGFloat) -> CGSize {
-        let estimatedHeight = attributedText.height(considering: maxWidth)
-        let estimatedWidth = attributedText.width(considering: estimatedHeight)
+        let constraintBox = CGSize(width: maxWidth, height: .greatestFiniteMagnitude)
+        let rect = attributedText.boundingRect(with: constraintBox, options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil).integral
```
SD10, line 190: @zhongwuzw Why does this work? 🤔 I think this breaks for small messages like `"lol"`?

zhongwuzw: @SD10 , `boundingRect` would calculate `width` and `height`, for small message, the height is the height of one line based on `font`.

PR #656, opened 2018-04-13
refacor avatarView

zhongwuzw:
1. Remove `AvatarView` class, it's performance penalty.
2. Uses `Factory` to make circle image based on `JSQ`.

TODO:

- [ ] Update comments.
- [ ] Add `CHANGELOG`.


zhongwuzw:
> `cornerRadius` and `masksToBounds` do not produce `offscreen rendering` in `iOS11.3` (But it at least has `offscreen rendering` in `iOS9`, I haven't device in `iOS10`, cannot verify, the behavior is different depending on `CA`'s version), see WWDC Session [419](https://developer.apple.com/wwdc14/419)

# Instruments:
## Tests in `iOS9`, `iPad mini`:
### Refactor results:
b5b0ec5e-e85d-438f-95b9-35b145c24cce
### Original results:
image
### Conclusion:
10+fps optimization than original version.

## Tests in `iOS11`, `iPhone6S+`:
### Reactor results:
image
### Original results:
image
### Conclusion:
Because `cornerRaduis` and `masksToBounds` do not produce `offscreen rendering`, the performance is almost the same, `Refactor version` average 51.8fps, `Original version` average 50.8fps.

Example/Sources/ConversationViewController.swift
```diff
@@ -315,9 +315,11 @@ extension ConversationViewController: MessagesDisplayDelegate {
 //        return .custom(configurationClosure)
     }
     
-    func configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
+    func configureAvatarView(_ avatarView: UIImageView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
         let avatar = SampleData.shared.getAvatarFor(sender: message.sender)
```
SD10, line 319: I just wanted to point out what I was trying to say before, this is an improvement to the example app but not MessageKit itself. This benchmark was done using your specific approach which is already possible in MessageKit.

We can't control what code users put inside of `configureAvatarView`. They're just going to use `avatarView.layer.cornerRadius`. 

I really don't like the `Factory` design pattern approach. I think it adds a lot of unnecessary boilerplate that the user may not use. If they want this pattern, they can create the factory themselves

Sources/Protocols/MessageAvatarImageDataSource.swift
```diff
@@ -0,0 +1,38 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+
+public protocol MessageAvatarImageDataSource {
```
SD10, line 27: I'm thinking `Avatar` could just become a protocol instead of defining this new one

SD10:
@zhongwuzw See my comments above but basically, I don't think we can consider this an enhancement to MessageKit because it is only a guaranteed improvement if they adopt your strategy in the example app. If they just use `layer.cornerRadius`, this change is pointless.

I'd like to see MessageKit adopt a lighter weight approach here. 
My thoughts on improvements to the `AvatarView` are:
- Declare `AvatarView` using a protocol that only requires an `image: UIImage?` property. This way users can use any kind of view they want as long as they provide a way to set an image.
- Provide the methods of `MessageImageFactory` as static methods on `AvatarView` through a protocol extension. This allows users to create circular/placeholder images.
- Maybe change `Avatar` to be a `protocol` instead.

zhongwuzw:
> Declare AvatarView using a protocol that only requires an image: UIImage? property. This way users can use any kind of view they want as long as they provide a way to set an image.

You mean `AvatarView` is a `UIView`,not protocol?  If it is, we need to remove and add `AvatarView` each time the cell dequeued for reuse. How to implement async image load efficiently?

> Provide the methods of MessageImageFactory as static methods on AvatarView through a protocol extension.

`MessageImageFactory` uses `Core Graphic` API to make cornerImage? and users call it in `configure..` method? If it is, it's the `CPU-bound` work.
@SD10 Can you provide a small prototype to demonstrate your thoughts?

SD10:
> You mean AvatarView is a UIView, not protocol? If it is, we need to remove and add AvatarView each time the cell dequeued for reuse. How to implement async image load efficiently?

It would be typed as a protocol, `AvatarView`, and we would use a `UIImageView` in the cell internally. Users could subclass the cell with anything that conforms to `AvatarView`. It wouldn't affect anything related to async images or cell dequeuing, just makes subclassing more flexible

> MessageImageFactory uses Core Graphic API to make cornerImage? and users call it in configure.. method? If it is, it's the CPU-bound work.

That's what I'm trying to explain. You call your code outside the `configure` method in the example app. We can't assume that users will do the same in their own app. Ultimately, they will use the `configure` method however they want 🤔

zhongwuzw:
@SD10 , so let me try to summarize your thoughts, you want to use a `protocol` to return a `image`, just like `JSQ`, and moves the `Core Graphics` related method to this `protocol`, right?

SD10:
> @SD10, so let me try to summarize your thoughts, you want to use a protocol to return a image, just like JSQ, and moves the Core Graphics related method to this protocol, right?

@zhongwuzw Pretty much what I'm thinking. We can't control if the user uses the methods or not. I think all that MessageKit should be responsible for is providing a few utility methods for creating images on the `AvatarView` namespace. 

I'm thinking we make `AvatarView` a protocol and then add these methods through a protocol extension. We can look more into this in the next version 👍

SD10:
@zhongwuzw I'm going to close this for the reasons we discussed above. We can look into providing helper methods at a later date. But IMO, there's nothing we can, nor should do, to enforce best practices on users

PR #659, opened 2018-04-16
Add cellTopLabel and rename topLabel to messageTopLabel

zhongwuzw:
- [ ] Add CHANGELOG entry.
- [ ] Discuss the necessity to remove `LabelAlignment.textAlignment`, because the label use `NSAttributedString` to  render.
- [ ] Fix tests fail.


Sources/Views/Headers & Footers/MessageDateHeaderView.swift
```diff
@@ -1,48 +0,0 @@
...
- FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
- AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
- OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
- SOFTWARE.
- */
-
-import UIKit
-
-open class MessageDateHeaderView: MessageReusableView {
```
SD10, line 27: We still need to leave this class for people who are using 1 message : 1 section 🤔

Sources/Models/AvatarPosition.swift
```diff
@@ -46,12 +46,6 @@ public struct AvatarPosition {
     /// An enum representing the verical alignment for an `AvatarView`.
     public enum Vertical {
         
-        /// Aligns the `AvatarView`'s top edge to the cell's top edge.
-        case cellTop
```
SD10: I was thinking about leaving these in the case of a user not using the `cellTopLabel`. If they mess up and use `.cellTop` with a `cellTopLabel` then this is a user error. 

Let me know what you think?

SD10:
> Discuss the necessity to remove LabelAlignment.textAlignment, because the label use NSAttributedString to render.

Yeah, this is an interesting idea. Not sure how I feel about it yet 🤔

Sources/Views/Headers & Footers/MessageDateHeaderView.swift
zhongwuzw: We truly need to provide two options to choose for users?

Users can implement their own header/footer view easily IMO.

Sources/Models/AvatarPosition.swift
zhongwuzw: ~~I prefer to remove. I don't like overlap of views. `AvatarView` maybe overlap `topMessageLabel` and `topCellLabel`, blended layers, and also misleading of tap gesture handler. More important, I think that seldom people would needs this case. 😂~~

Holy shit, I found the most popular message app `WeChat` in China uses this case.
img_c9fe79d9da1f-1

zhongwuzw: So we should preserve .cellTop case, then, we should let avatar align the top equal to `topCellLabel`'s top or `topMessageLabel`'s top?

SD10: @zhongwuzw Yeah, I think this is a good idea. We can add back `.cellBottom` to be consistent too.

Sources/Views/Headers & Footers/MessageDateHeaderView.swift
SD10: Hmmm. I was thinking it's more for backward compatibility. Let's see if any @MessageKit/contributors have any comments.

Sources/Models/AvatarPosition.swift
MacMeDan: Yes, overlapping views is a common request for many users.

MacMeDan:
I think you are on a good thought path with this @SD10.
 One wants to preserve ease of use but allow for customization. IMO to the point of @zhongwuzw reducing overlap ( or redundancy) does help manage complexity. Tho I would say the best option would be to implement something that manages the location of all the views at a higher `layout Coordinator` level. But that as a solution introduces more complexity in an effort to reduce complexity.

SD10:
@MacMeDan Thanks for the feedback 💯 

> Tho I would say the best option would be to implement something that manages the location of all the views at a higher layout Coordinator level. But that as a solution introduces more complexity in an effort to reduce complexity.

I agree. I wish there was a better way to handle layout but we're too far down the rabbit hole at this point 😞

MacMeDan:
Ya that is my hesitation as well especially at this point.

Sources/Models/AvatarPosition.swift
zhongwuzw: @SD10 , so we should make `avatarView`'s top equal to `topMessageLabel`'s top in `.cellTop` case?

SD10: @zhongwuzw Yeah this is confusing. Should we add a `.messageLabelTop` for this?

zhongwuzw: 🤔 , I think we should, most users maybe need `.messageLabelTop`, other than `.cellTop`, I'll add these two cases.

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -75,32 +76,14 @@ open class MessageSizeCalculator: CellSizeCalculator {
...
-            cellHeight += topLabelHeight
-            cellHeight += messageContainerHeight
-            cellHeight += messageVerticalPadding
-        case .messageTop, .messageCenter, .messageBottom:
-            cellHeight += max(avatarHeight, messageContainerHeight)
-            cellHeight += messageVerticalPadding
-            cellHeight += topLabelHeight
-            cellHeight += bottomLabelHeight
-        }
+        let cellHeight: CGFloat = cellTopLabelHeight + messageTopLabelHeight
```
SD10: I'm not sure this is an accurate way to measure the height. What happens if the avatar is larger than the `messageContainerView` in some situations? Do you want me to work on this?

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -41,8 +41,8 @@ open class MessageSizeCalculator: CellSizeCalculator {
     public var incomingMessagePadding = UIEdgeInsets(top: 0, left: 4, bottom: 0, right: 30)
     public var outgoingMessagePadding = UIEdgeInsets(top: 0, left: 30, bottom: 0, right: 4)
 
-    public var incomingCellTopLabelAlignment = LabelAlignment(textAlignment: .left, textInsets: UIEdgeInsets(left: 42))
```
SD10, line 44: I think we still need `cellTopLabelAlignment`

SD10:
We need to update `MessagesDisplayDelegate` to allow users to configure `cellTopLabelAlignment`, `messageTopLabelAlignment`, and `cellBottomLabelAlignment`

Sources/Layout/MessageSizeCalculator.swift
zhongwuzw: @SD10 , 👌you can enhance it, the reason why I ignore the height of `avatarView`, is if we consider its height, we can hardly layout the views, e.x., think about `messageCenter` case, if `avatarView`'s height is enough bigger, like 1000px, the `messageContainerHeight`'s height is `10`px, for the original way, the height of cell is bigger than 1000px, and then, let's think about how to layout the subviews of cell, like `messageContainerView`, we have `padding`, it would break anything.

zhongwuzw:
> We need to update MessagesDisplayDelegate to allow users to configure cellTopLabelAlignment, messageTopLabelAlignment, and cellBottomLabelAlignment

@SD10, yep, my opinion is we need users to return the `NSAttributedString`, and provide the default alignment in layout. I don't think we still need provide these things. 😂 Can you update these things besides cell's height calculation?

SD10:
Yeah @zhongwuzw sorry I was thinking of the old MessageKit 😂 I will update this

SD10:
@zhongwuzw I handled all the edge cases for the `cellContentHeight` calculations. That was horrible 😂

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -66,18 +74,18 @@ open class MessageContentCell: MessageCollectionViewCell {
     }
 
     open func setupSubviews() {
-        contentView.addSubview(messageContainerView)
-        contentView.addSubview(avatarView)
         contentView.addSubview(cellTopLabel)
+        contentView.addSubview(messageTopLabel)
         contentView.addSubview(cellBottomLabel)
```
SD10: @zhongwuzw I think we should rename `cellBottomLabel` -> `messageBottomLabel` because it is pinned to the bottom of the message and not the cell.

zhongwuzw: Yes, I just got a glimpse, we should make it to `messageBottomLabel` and rename `cellBottom` to `messageLabelBottom` in `AvatarPosition`'s `vertical` enum.

SD10: 👍 on the rename. but `.cellBottom` actually aligns to the cell bottom and not `messageLabelBottom` 😂

zhongwuzw: Emm, maybe not, you can refer my [comments](https://github.com/MessageKit/MessageKit/pull/659#discussion_r182613839) above. 🤔

SD10: It wasn't easy but I fixed all of the layout problems. It will layout correctly even for `1000px` height.

zhongwuzw: Emm, when in `1000px` case, `.cellBottom` would not work, `avatarView`'s bottom is equal to `cellBottomLabel`'s bottom, but `cellBottomLabel`'s `minY` is just equal to `containerView`'s `padding.bottom`, not the cell's bottom.

SD10: `avatarView`'s bottom is equal to the cell's bottom. try my changes 😃

zhongwuzw: Oops, I looked at the original version. 😢

SD10:
I think all that's left to do here is:
- [x] Rename `cellBottomLabel` to `messageBottomLabel`
- [x] Add CHANGELOG entries
- [x] Fix failing tests

Will take care of this tomorrow

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -136,11 +149,26 @@ open class MessageSizeCalculator: CellSizeCalculator {
         return CGSize(width: messagesLayout.itemWidth, height: height)
     }
 
-    public func cellTopLabelAlignment(for message: MessageType) -> LabelAlignment {
+    public func cellTopLabelAlignment(for message: MessageType, at indexPath: IndexPath) -> LabelAlignment {
```
SD10: cc @zhongwuzw do we need to pass `indexPath` here?

Sources/Layout/MessagesCollectionViewLayoutAttributes.swift
```diff
@@ -30,15 +30,18 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
     // MARK: - Properties
 
     public var avatarSize: CGSize = .zero
-    public var avatarPosition = AvatarPosition(vertical: .cellBottom)
+    public var avatarPosition = AvatarPosition(vertical: .messageBottom)
```
SD10: I should revert this for consistency

SD10 (approved):
:shipit:

PR #661, opened 2018-04-21
Move LabelSize function into Extensions

PersianDevelopers:
Change-Id: Ie288df107e64d47049cf399e93226f0751c5c355

Simple Cleaning codes. moving that function into extension let others use that function in any custom cell

No

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators: Simulator iPhone 8

**iOS Version: iOS 10

**Swift Version: 3.0

**MessageKit Version: 0.13.4

### DIFF
--- a/Sources/Extensions/NSAttributedString+Extensions.swift
+++ b/Sources/Extensions/NSAttributedString+Extensions.swift
@@ -41,4 +41,19 @@ extension NSAttributedString {
         return rect.width
         
     }
+    
+    /// Returns the size required fit a NSAttributedString considering a constrained max width.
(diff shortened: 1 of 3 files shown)


SD10:
Hey @PersianDevelopers, thanks for opening this PR 👍 

Unfortunately, we've changed a lot of the layout logic in the MessageKit `development` branch so this CHANGE would be difficult to merge on `master`. I also do not think it fits well with the MessageKit API. These sort of extensions should remain internal.

Thank you for taking the time to contribute ❤️

PR #666, opened 2018-04-29
MessageKit 1.0-beta

SD10:


hyouuu:
Finally!!!

zhongwuzw:
@SD10 , I prepare to remove `Cocoapods` dependency for `Example` project, I don't want to `pod install` every time when the framework changed, I want to just open `Example` project and click run button, what do you think? 😂

SD10:
@zhongwuzw I remember we changed to this setup because we want to make sure everything is working correctly on CocoaPods

Sources/Controllers/MessagesViewController.swift
```diff
@@ -152,6 +135,138 @@ open class MessagesViewController: UIViewController {
             NSLayoutConstraint.activate([top, bottom, trailing, leading])
         }
     }
+
+    // MARK: - UICollectionViewDataSource
+
+    open func numberOfSections(in collectionView: UICollectionView) -> Int {
+        guard let collectionView = collectionView as? MessagesCollectionView else {
+            fatalError(MessageKitError.notMessagesCollectionView)
```
zhongwuzw, line 143: TBO, we have much these `fatalError`, maybe we can have a better friendly way to do this (maybe `assert` or just use some default values)? other than just stop execution in `Debug` or `Release`.

SD10: I agree, but I haven't found any good solution TBH. I had some default values before and removed them all because it just doesn't make sense 😕 We would be writing code that never gets executed. You have to set all 3 protocols `MessagesDataSource`, `MessagesDisplayDelegate`, and `MessagesLayoutDelegate` or nothing works

BasThomas: Yeah, I think the `fatalError`s here are valid. With good documentation in the superclasses this should not be an issue. And if it is; its a valid reason to crash there as there is no way to recover.

Changing these to `assert`s could be an option to prevent crashes in release builds, but I am not sure if that’s the best approach. How does it work with APIs like `UITableView` and `UICollectionView`?

SD10: Thanks for the feedback @BasThomas! 

> How does it work with APIs like UITableView and UICollectionView?

This is not standard `UITableView` or `UICollectionView` behavior. Instead of extending the UIKit protocols, we conform to them internally and then call our own MessageKit protocols inside of them.  This is to decouple the `MessagesDataSource` from `UICollectionViewDataSource`, so the project can be used with things like IGListKit. We're forced to provide a valid implementation or it is a `fatalError` 😕

BasThomas: Seems like a good usage of `fatalError` then. :)

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    public func avatarSize(for message: MessageType) -> CGSize {
+        let dataSource = messagesLayout.messagesDataSource
+        let isFromCurrentSender = dataSource.isFromCurrentSender(message: message)
+        return isFromCurrentSender ? outgoingAvatarSize : incomingAvatarSize
+    }
+
+    // MARK: - Top cell Label
+
+    public func cellTopLabelSize(for message: MessageType, at indexPath: IndexPath) -> CGSize {
```
zhongwuzw, line 145: @SD10 Can we integrate these methods to one?  it's a little bit redundancy ,`cellTopLabel`/`messageTopLabel`/`messageBottomLabel`'s `size` and `alignment` are almost the same.

SD10: I don't think so 😕 I didn't even want these to be public but people have requested them to be accessible when they subclass. The only way to combine it into 1 method is to pass it an `enum` I think? I really don't like those kinds of API

SD10:
Don't know if I want to merge this as is or ship the beta from `development`. I guess we can do the latter for future releases

Sources/Layout/MessageSizeCalculator.swift
zhongwuzw: 👌

zhongwuzw:
🤔 I prefer to choose just merge to `master`, if we ship the beta in `development`, how can we get the feedback from the user? 😂

zhongwuzw:
Oops, we can change the default branch. 😢

SD10:
ah @zhongwuzw. I meant I can push a new release to CocoaPods from the `development` branch. That's what we do for Moya. I think this release is so large that you're probably right, `master` is better

SD10:
Probs gonna merge this if CI passes and open an issue for adding better docs or testing as starter tasks

SD10:
I wish I had the free time to benchmark this vs the current `master`. I'll probably come back to it when I do. I know I reduced layout time by 70% with #519 and some amazing changes by @zhongwuzw probably reduced it by half after that

SD10:
Yeah, that just happened :shipit:

Sources/Layout/MessageSizeCalculator.swift
rnystrom: asdf

Sources/Controllers/MessagesViewController.swift
rnystrom: test

rnystrom: test

rnystrom: test

rnystromtest: test

rnystromtest: test

rnystromtest: test

rnystrom: Test

rnystromtest: Test

rob-nash:
Test comment

Issue #694, opened 2018-05-21
[Question] Removing avatar from all MessageKind


adri4silva:
Today I had to remove the Avatar View from all MessageKind cells in my app, and I faced this problem:

```swift
if let layout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout {
    layout.textMessageSizeCalculator.outgoingAvatarSize = .zero
    layout.textMessageSizeCalculator.incomingAvatarSize = .zero
    layout.emojiMessageSizeCalculator.outgoingAvatarSize = .zero
    layout.emojiMessageSizeCalculator.incomingAvatarSize = .zero
    layout.photoMessageSizeCalculator.incomingAvatarSize = .zero
    layout.photoMessageSizeCalculator.outgoingAvatarSize = .zero
    layout.locationMessageSizeCalculator.outgoingAvatarSize = .zero
    layout.locationMessageSizeCalculator.incomingAvatarSize = .zero
    layout.videoMessageSizeCalculator.outgoingAvatarSize = .zero
    layout.videoMessageSizeCalculator.incomingAvatarSize = .zero
    layout.attributedTextMessageSizeCalculator.outgoingAvatarSize = .zero
    layout.attributedTextMessageSizeCalculator.outgoingAvatarSize = .zero
}
```
Are there other ways to remove Avatar View?

SD10:
Hah, @ndonald2 made the suggestion to add a convenience method to the layout object for this. I’m open to PRs but we need to make methods for similar properties as well. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #697, opened 2018-05-22
Added convenience method to set properties for all MessageSizeCalcula…

zhongwuzw:
Support for #694 .

### DIFF
--- a/Sources/Layout/MessagesCollectionViewFlowLayout.swift
+++ b/Sources/Layout/MessagesCollectionViewFlowLayout.swift
@@ -155,4 +155,70 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
         let calculator = cellSizeCalculatorForItem(at: indexPath)
         return calculator.sizeForItem(at: indexPath)
     }
+    
+    /** Set `incomingAvatarSize` of all `MessageSizeCalculator`s */
+    open func setMessageIncomingAvatarSize(_ newSize: CGSize) {
+        messageSizeCalculators().forEach { $0.incomingAvatarSize = newSize }
+    }
+    
+    /** Set `outgoingAvatarSize` of all `MessageSizeCalculator`s */
+    open func setMessageOutgoingAvatarSize(_ newSize: CGSize) {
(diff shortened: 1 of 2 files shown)


zhongwuzw:
@SD10 Let's provide one shot method.

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -155,4 +155,70 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
         let calculator = cellSizeCalculatorForItem(at: indexPath)
         return calculator.sizeForItem(at: indexPath)
     }
+    
+    /** Set `incomingAvatarSize` of all `MessageSizeCalculator`s */
```
BasThomas: Nit, but what about a triple `/` (`///`) instead of these multiline comments?

ndonald2:
Looks good to me! Thanks for doing this.

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -155,4 +155,70 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
         let calculator = cellSizeCalculatorForItem(at: indexPath)
         return calculator.sizeForItem(at: indexPath)
     }
+    
+    /// Set `incomingAvatarSize` of all `MessageSizeCalculator`s
+    open func setMessageIncomingAvatarSize(_ newSize: CGSize) {
```
SD10: Think these should be `public` and not `open`? They're helper methods. I don't know if we should allow people to override? I don't see any major harm in leaving it but what do you think?

SD10:
🙏 I'm so lucky to have other contributors 😭

Sources/Layout/MessagesCollectionViewFlowLayout.swift
zhongwuzw: No big differences I think, I'll change this to `public`.

SD10 (approved):
Great 💯

Issue #674, opened 2018-05-05
How to hide the AvatarView with v1.0.0


janweinkauff:
I'm trying to hide the `AvatarView` with the new changes of 1.0.0-beta.1. 

Right now I use `avatarView.isHidden = true` in `MessagesDisplayDelegate.configureAvatarView` which only hides the view but will not remove the space of the `AvatarView` due to the size from `outgoingAvatarSize` or `incomingAvatarSize` of `MessageSizeCalculator`.

Is it possible to change the value of `outgoingAvatarSize` or `incomingAvatarSize`? 

I couldn't find any delegate callback or entry point for a subclass.

janweinkauff:
After a short chat in the slack channel I found a solution based on the suggestion from @creyD. 

`outgoingAvatarSize` or `incomingAvatarSize` are just properties of the `MessagesCollectionViewFlowLayout `.

```
if let layout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout {
  layout.textMessageSizeCalculator.outgoingAvatarSize = .zero
  ...
}
```

Faisal0sal:
```
if let layout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout {
layout.textMessageSizeCalculator.outgoingAvatarSize = .zero
layout.textMessageSizeCalculator.incomingAvatarSize = .zero
}
```

doing this in viewdidload ^^ still gave me an avatar on outgoing messages

janweinkauff:
I also did this change in the `viewDidLoad` method. It works on my side.
* Did you check if the cast `messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout` succeeds?
* Maybe set a breakpoint to `MessageSizeCalculator.avatarSize(for message:)` to understand from what change the `incomingAvatarSize` is set?

Faisal0sal:
it works now, just moving it to the bottom of my viewDidLoad did it. not sure how this affect it.

Faisal0sal:
Actually although this does work on text messages, once you start loading more messages it starts to flounder between having it and removing it. It also appears on media messages.

![simulator screen shot - iphone x - 2018-05-05 at 11 41 09](https://user-images.githubusercontent.com/10188948/39661407-bbbdf44e-5059-11e8-9fc7-419842814ba5.png)

janweinkauff:
[...]

PR #699, opened 2018-05-24
👻Fix avatar view layout and message top/bottom label insets

Artheyn:
Actually this PR includes two fixes appearing when you set the avatarView size to `.zero`:

1. The first issue is that when cell are reused, even if we tell the `sizeCalculator` that the `avatarView` size should be `.zero`, it still display an avatar view and doesn't take our `.zero` size into account.

You can easily reproduce it in the Example project by just setting an avatarSize to `.zero`:

Example/Sources/ConversationViewController.swift
```
  let flowLayout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout
  flowLayout?.`attributedTextMessageSizeCalculator.outgoingAvatarSize = .zero
```

Visually:

screen shot 2018-05-22 at 14 43 43

2. The second issue is, when setting an avatarSize to `.zero`, the message bottom and top label insets aren't recomputed and then appeared to be misaligned:

screen shot 2018-05-18 at 13 58 40

Yes an issue was opened and appears to talk about the same issues than I faced: https://github.com/MessageKit/MessageKit/issues/674

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Devices/Simulators:**
iPhone 7
iPhone 8
iPhone X

**iOS Version:**
10.3.1
11.0.1
11.1
11.3

**Swift Version:**
Swift 4.1

**MessageKit Version:** 
1.0.0-beta

### DIFF
--- a/Sources/Layout/MessageSizeCalculator.swift
+++ b/Sources/Layout/MessageSizeCalculator.swift
@@ -44,11 +44,21 @@ open class MessageSizeCalculator: CellSizeCalculator {
     public var incomingCellTopLabelAlignment = LabelAlignment(textAlignment: .center, textInsets: .zero)
     public var outgoingCellTopLabelAlignment = LabelAlignment(textAlignment: .center, textInsets: .zero)
 
-    public var incomingMessageTopLabelAlignment = LabelAlignment(textAlignment: .left, textInsets: UIEdgeInsets(left: 42))
-    public var outgoingMessageTopLabelAlignment = LabelAlignment(textAlignment: .right, textInsets: UIEdgeInsets(right: 42))
+    public var incomingMessageTopLabelAlignment: LabelAlignment {
+        return LabelAlignment(textAlignment: .left, textInsets: UIEdgeInsets(left: incomingMessageLeftMargin()))
+    }
 
-    public var incomingMessageBottomLabelAlignment = LabelAlignment(textAlignment: .left, textInsets: UIEdgeInsets(left: 42))
-    public var outgoingMessageBottomLabelAlignment = LabelAlignment(textAlignment: .right, textInsets: UIEdgeInsets(right: 42))
+    public var outgoingMessageTopLabelAlignment: LabelAlignment {
+        return LabelAlignment(textAlignment: .right, textInsets: UIEdgeInsets(right: outgoingMessageRightMargin()))
+    }
+
+    public var incomingMessageBottomLabelAlignment: LabelAlignment {
+        return LabelAlignment(textAlignment: .left, textInsets: UIEdgeInsets(left: incomingMessageLeftMargin()))
+    }
+
+    public var outgoingMessageBottomLabelAlignment: LabelAlignment {
+        return LabelAlignment(textAlignment: .right, textInsets: UIEdgeInsets(right: outgoingMessageRightMargin()))
+    }
 
     open override func configure(attributes: UICollectionViewLayoutAttributes) {
         guard let attributes = attributes as? MessagesCollectionViewLayoutAttributes else { return }
@@ -65,9 +75,8 @@ open class MessageSizeCalculator: CellSizeCalculator {
         attributes.cellTopLabelSize = cellTopLabelSize(for: message, at: indexPath)
         attributes.messageTopLabelSize = messageTopLabelSize(for: message, at: indexPath)
         attributes.messageTopLabelAlignment = messageTopLabelAlignment(for: message)
-
-        attributes.messageBottomLabelAlignment = messageBottomLabelAlignment(for: message)
         attributes.messageBottomLabelSize = messageBottomLabelSize(for: message, at: indexPath)
+        attributes.messageBottomLabelAlignment = messageBottomLabelAlignment(for: message)
     }
 
     open override func sizeForItem(at indexPath: IndexPath) -> CGSize {
@@ -218,6 +227,16 @@ open class MessageSizeCalculator: CellSizeCalculator {
         let rect = attributedText.boundingRect(with: constraintBox, options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil).integral
 
         return rect.size
+    }
+
+    /// - Returns: margin between the Leading edge of the Collection View and the Leading edge of the incoming message bubble.
+    public func incomingMessageLeftMargin() -> CGFloat {
+        return incomingAvatarSize.width + incomingMessagePadding.left + messagesLayout.sectionInset.left
+    }
+
+    /// - Returns: margin between the Trailing edge of the Collection View and the Trailing edge of the outgoing message bubble.
+    public func outgoingMessageRightMargin() -> CGFloat {
+        return outgoingAvatarSize.width + outgoingMessagePadding.right + messagesLayout.sectionInset.right
     }
 }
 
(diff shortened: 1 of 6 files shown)


Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -171,7 +171,10 @@ open class MessageContentCell: MessageCollectionViewCell {
     /// Positions the cell's `AvatarView`.
     /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
     open func layoutAvatarView(with attributes: MessagesCollectionViewLayoutAttributes) {
-        guard attributes.avatarSize != .zero else { return }
+        guard attributes.avatarSize != .zero else {
```
SD10, line 174: I think we need to apply the same logic to the other subviews as well

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -44,11 +44,21 @@ open class MessageSizeCalculator: CellSizeCalculator {
     public var incomingCellTopLabelAlignment = LabelAlignment(textAlignment: .center, textInsets: .zero)
     public var outgoingCellTopLabelAlignment = LabelAlignment(textAlignment: .center, textInsets: .zero)
 
-    public var incomingMessageTopLabelAlignment = LabelAlignment(textAlignment: .left, textInsets: UIEdgeInsets(left: 42))
-    public var outgoingMessageTopLabelAlignment = LabelAlignment(textAlignment: .right, textInsets: UIEdgeInsets(right: 42))
+    public var incomingMessageTopLabelAlignment: LabelAlignment {
```
SD10, line 47: The problem with these now being stored properties is that they are no longer configurable

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -219,6 +228,16 @@ open class MessageSizeCalculator: CellSizeCalculator {
 
         return rect.size
     }
+
+    /// - Returns: margin between the Leading edge of the Collection View and the Leading edge of the incoming message bubble.
+    public func incomingMessageLeftMargin() -> CGFloat {
+        return incomingAvatarSize.width + incomingMessagePadding.left + messagesLayout.sectionInset.left
```
SD10, line 234: These values are already taken into account when calling `layoutAvatarView`. We apply the padding to the origin of the `AvatarView`. 

Also, we don't need to consider the `sectionInset` because this is considered when calling `itemWidth` from the `MessagesCollectionViewFlowLayout` object.

SD10:
@Artheyn Thanks for taking the time to submit this PR 👍 I gave some feedback above. I need to give it a deeper look this weekend. 

However, I'm only really seeing the first problem you mentioned -- we need to set the `AvatarView` frame to `.zero`.

I think the first problem is what is causing the second problem you mentioned. The alignment of the other subviews of the cell use the `AvatarView` as the anchor. If the `AvatarView` isn't laid out properly, the other cell contents won't be either.

Sources/Layout/MessageSizeCalculator.swift
zhongwuzw: I don't think the dynamic margin in here is a better way, maybe user don't need this behavior, but add a delegate method in here is ok IMO, @SD10 what do you think? 🤔

SD10: @zhongwuzw Can you clarify what you mean? I don't want to make it any more dynamic than `outgoing` and `incoming`. And I want all properties to be stored properties. Or else we're heading back to where we were before

zhongwuzw: image
As screenshot from @Artheyn , because we use the `outgoing` and `incoming`, so users can not dynamic adjust the insets to make text align to the bubble.
I think add delegate for `labelAlignment` is ok, because it's only effect label's internal drawing.

SD10: Thanks for clarifying @zhongwuzw 💯 Didn't you mention before that users can add the alignment to `NSAttributedString`? I will look into the effects of adding a delegate

SD10: But also note that the two `Steven` labels above are laid out properly. The bottom label is wrong because the avatar is misplaced

Sources/Layout/MessageSizeCalculator.swift
Artheyn: Yep you're right. I was thinking about subclassing (MessageSizeCalculator or another subclass of it) to override this values, but I forgot to modify `public` to `open`.

Sources/Layout/MessageSizeCalculator.swift
Artheyn: > These values are already taken into account when calling layoutAvatarView. We apply the padding to the origin of the AvatarView.

Yeap, it was a way to dynamically compute the labels insets. There is probably a better way.

> But also note that the two Steven labels above are laid out properly. The bottom label is wrong because the avatar is misplaced

In fact, the insets (LabelAlignment) was statically set to `42` before, and I couldn't find any place where it was updated, so I think it cannot be related to the avatar position.

In my opinion we could have a default behaviour which aligns properly the top and bottom labels to the leading/trailing edge of the chat bubble and allow the integrator to update this through delegate like @zhongwuzw says or even subclassing.

SD10: > In fact, the insets (LabelAlignment) was statically set to 42 before, and I couldn't find any place where it was updated, so I think it cannot be related to the avatar position.

I calculate the frame of the label by using the frame of the `AvatarView`. So if the `AvatarView` frame is wrong, then so is the label.

Artheyn: Ok. By inspecting the frames with and without avatar, we can see that the label always take the entire cell view. So there is perhaps an issue in the label frame calculation.

screen shot 2018-05-25 at 09 54 00

zhongwuzw: @SD10 ,Yeah I mentioned before, remove `textInsets` and `alignment`, and let's user implements these things by `NSAttributedString`.

SD10: @Artheyn No, that is working correctly. Sizing the `NSAttributedString` to fit for a `UILabel` has a huge performance penalty on scrolling, so we make it a static size.

SD10: @zhongwuzw Thanks, I'm going to look into this when I get more time. I may have to redo the origin calculations for cells

zhongwuzw: @SD10 , Give you some my thoughts, remove `insetsLabel` and `*LabelAlignment`, instead uses `NSMutableParagraphStyle`'s `firstIndent`、 `tailIndent`、 `alignment `.

Artheyn: Ok. Thanks. If I can help any further please let me know.

SD10: @Artheyn No problem, this is a tricky one to fix because if we apply the `.zero` frame I think we will end up breaking the layout of the other views. This is a really good catch on your part 👍

zhongwuzw: @SD10 ,We need to fix these `.zero` issues, I think set `frame` to `.zero` is ok, do you have any concerns?

SD10: @zhongwuzw We have to set the frames to `.zero` in another PR. I agree this needs to be done. I can't merge this one because it has other breaking changes that are not in the direction MessageKit is heading.

PR #707, opened 2018-05-31
Fixed MessagesCollectionView's bottom inset when hardware keyboard is…

### LINKED ISSUE #562: Keyboard dismiss bug

### LINKED ISSUE #690: MessagesCollectionView bottom inset miscalculation when hardware keyboard is connected
👻
## General Information
Issue calculating the MessagesCollectionView bottom inset when a hardware keyboard is attached to the device/simulator.

* MessageKit Version:
v 0.13.5 (though code looks exactly the same in the v1-beta-1)

* iOS Version(s):
iOS 11.2

* Swift Version:
Swift 4

* Devices/Simulators:
Simulator 7+/8+

YES: (steps in simulator)
1) Make sure 'Simulator -> Hardware -> Keyboard -> Connect Hardware Keyboard' is checked
2) Focus the MessageInputBar
3) Type several newline characters
4.1) Delete one or more newline characters at once, OR
4.2) Tap Send
5) Observe

##  What happened?
The MessagesCollectionView inset is now too large and the collection view bottom is too far from the top of the MessageInputBar.

screen shot 2018-05-16 at 1 48 56 pm
screen shot 2018-05-16 at 1 49 04 pm
screen shot 2018-05-16 at 1 49 13 pm

The MessagesCollectionView inset is correctly calculated to match the top of the MessageInputBar

## Technical Info:
Bug location: MessagesViewController+Keyboard.handleKeyboardDidChangeState(_ notification: Notification)
Thoughts:
From what I observed, the first if statement's condition:
```swift
(keyboardEndFrame.origin.y + keyboardEndFrame.size.height) > UIScreen.main.bounds.height
```
returns false, because the two sides of the expression are equal (when the input bar grows they are equal too).

So the calculation moves onto the 'software keyboard' calculation, of which the key line is:
```swift
let afterBottomInset = keyboardEndFrame.height > keyboardOffsetFrame.height ? (keyboardEndFrame.height - iPhoneXBottomInset) : keyboardOffsetFrame.height
```
Since `afterBottomInset` eventually becomes the `MessagesCollectionView`'s bottom inset.
When the `MessageInputBar` grows, the `keyboardEndFrame.height` is greater than the `keyboardOffsetFrame.height` (aka. the old `MessageInputBar` height). So the bottom inset becomes the `keyboardEndFrame.height`.
When the `MessageInputBar` shrinks, however, the `keyboardEndFrame.height` is less than the `keyboardOffsetFrame.height` (aka. the old `MessageInputBar` height), and so the inset stays at the previous `MessageInputBar` height.

Maybe we could leverage the `MessageInputBar`'s `previousIntrinsicContentSize` and check if it is getting smaller to make more informed calculations? Or adjust the initial if condition to more accurately identify when we need to use the 'hardware' calculation.

(As a side note, if the 'hardware' `if` statement had been triggered, then I don't think the `maintainPositionOnKeyboardFrameChanged` setting would be honored).

zhongwuzw:
Closed #568 .
Fixed #562 , Fixed #690 .


Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -53,22 +53,19 @@ extension MessagesViewController {
...
-            let differenceOfBottomInset = afterBottomInset - messageCollectionViewBottomInset
-
-            if maintainPositionOnKeyboardFrameChanged && differenceOfBottomInset != 0 {
-                let contentOffset = CGPoint(x: messagesCollectionView.contentOffset.x, y: messagesCollectionView.contentOffset.y + differenceOfBottomInset)
-                messagesCollectionView.setContentOffset(contentOffset, animated: false)
-            }
-
-            messageCollectionViewBottomInset = afterBottomInset
+        
+        guard !isMessagesControllerBeingDismissed else { return }
```
SD10, line 57: @zhongwuzw Ok, I can see how this fixes the dismissal in #562. How does this also solve the hardware keyboard in #690?

zhongwuzw: @SD10 Another way to calculate the `bottomInset`.

SD10:
Sorry, GitHub UI malfunction 😂 @zhongwuzw do you think we could point this at `development` instead? I still need to release the official 1.0, just waiting for my company to migrate from our custom fork of MessageKit.

zhongwuzw:
@SD10 It's ok, not big bug. I'll switch to `development` and merge it.

SD10:
@zhongwuzw Great thanks 👌 We have to do a better job of keeping `master` and `development` in sync so that this doesn't happen, but the blame is on me

PR #709, opened 2018-06-01
Rewrite issue template to a HTML comment

### LINKED ISSUE #653: Improve the GitHub issue template
[Moya has a good example of something more lightweight](https://raw.githubusercontent.com/Moya/Moya/master/.github/Issue_template.md). I think it would be good to still mention some things that are useful to provide like MessageKit version, etc. But the entire template should be a comment. I don't want to see a bunch of empty fields when users don't fill out the proper information.

BasThomas:
Would love some feedback. This is only rewriting the issue template, not the pull request template.

Closes #653.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,26 +1,29 @@
-## General Information
+<!--
+Hi, thanks so much for opening an issue! 🤗
 
-* MessageKit Version:
+To better pinpoint (and solve) the issue you're experiencing, we could use some information on your behalf.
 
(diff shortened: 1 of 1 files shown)


SD10:
@BasThomas No feedback needed, this will work great. It's still as detailed but without all the hard requirements 💯

BasThomas:
I can’t imagine that it is perfect as-is, but we can always refine it later if needed, I suppose. 😛

PR #710, opened 2018-06-02
Add customizable accessoryView besides messageContainer

hyouuu:
Add an accessoryView and give client a chance to config it to whatever they like (e.g. a button, image)

Found one ;) https://github.com/MessageKit/MessageKit/issues/190

NA

Any other comments?
-------------------
Let me know how you want to approach it!
This was originally PR'ed for 0.12 at https://github.com/MessageKit/MessageKit/pull/432

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone X device; iPhone 8+ simulator

**iOS Version:** 11.4

**Swift Version:** 4.1

**MessageKit Version:** 1.0.0-beta.1

Example screenshot where in configCell added a UIView with green background:

screen shot 2017-12-31 at 1 01 22 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - Added convenience method to set properties for all `MessageSizeCalculator`s.
 [#697](https://github.com/MessageKit/MessageKit/pull/697) by [@zhongwuzw](https://github.com/zhongwuzw).
 
+- Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
+
 ## [[Prerelease] 1.0.0-beta.1](https://github.com/MessageKit/MessageKit/releases/tag/1.0.0-beta.1)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+### Added
+
+- Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
+
 ## [1.0.0](https://github.com/MessageKit/MessageKit/releases/tag/1.0.0)
 
 - First major release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,17 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## Upcoming release
 
+## [1.0.0](https://github.com/MessageKit/MessageKit/releases/tag/1.0.0)
+
+- First major release.
+
+## [[Prerelease] 1.0.0-beta.2](https://github.com/MessageKit/MessageKit/releases/tag/1.0.0-beta.2)
+
+### Changed
+
+- Relaxed the settable requirement for the `LocationItem` and `MediaItem` protocols.
+[#727](https://github.com/MessageKit/MessageKit/pull/716) by [@SD10](https://github.com/sd10).
+
 ### Fixed
 
 - Fixed `MessageContentCell`'s subviews `frame` when size equal to `.zero`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,6 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - Added convenience method to set properties for all `MessageSizeCalculator`s.
 [#697](https://github.com/MessageKit/MessageKit/pull/697) by [@zhongwuzw](https://github.com/zhongwuzw).
 
-- Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
-
 ## [[Prerelease] 1.0.0-beta.1](https://github.com/MessageKit/MessageKit/releases/tag/1.0.0-beta.1)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Fixed
 
+- Fixed `MessageContentCell`'s subviews `frame` when size equal to `.zero`.
+[#716](https://github.com/MessageKit/MessageKit/pull/716) by [@zhongwuzw](https://github.com/zhongwuzw).
+
 - Fixed `MessagesCollectionView`'s bottom inset when hardware keyboard is connected. Do not adjust `MessagesCollection`'s inset when being popped.
 [#707](https://github.com/MessageKit/MessageKit/pull/707) by [@zhongwuzw](https://github.com/zhongwuzw).
 
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -30,14 +30,12 @@ extension MessagesViewController {
 
     internal func addKeyboardObservers() {
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: .UIKeyboardWillChangeFrame, object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: .UIKeyboardDidChangeFrame, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleTextViewDidBeginEditing(_:)), name: .UITextViewTextDidBeginEditing, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.adjustScrollViewInset), name: .UIDeviceOrientationDidChange, object: nil)
     }
 
     internal func removeKeyboardObservers() {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillChangeFrame, object: nil)
-        NotificationCenter.default.removeObserver(self, name: .UIKeyboardDidChangeFrame, object: nil)
         NotificationCenter.default.removeObserver(self, name: .UITextViewTextDidBeginEditing, object: nil)
         NotificationCenter.default.removeObserver(self, name: .UIDeviceOrientationDidChange, object: nil)
     }
@@ -55,19 +53,18 @@ extension MessagesViewController {
     @objc
     private func handleKeyboardDidChangeState(_ notification: Notification) {
         guard let keyboardEndFrame = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }
-
+        
         guard !isMessagesControllerBeingDismissed else { return }
-
-        let newBottomInset = keyboardEndFrame.height - safeAreaBottomInset
-
+        
+        let newBottomInset = view.frame.height - keyboardEndFrame.minY - iPhoneXBottomInset
+        
         let differenceOfBottomInset = newBottomInset - messageCollectionViewBottomInset
-
+        
         if maintainPositionOnKeyboardFrameChanged && differenceOfBottomInset != 0 {
-            let contentOffset = CGPoint(x: messagesCollectionView.contentOffset.x,
-                                        y: messagesCollectionView.contentOffset.y + differenceOfBottomInset)
+            let contentOffset = CGPoint(x: messagesCollectionView.contentOffset.x, y: messagesCollectionView.contentOffset.y + differenceOfBottomInset)
             messagesCollectionView.setContentOffset(contentOffset, animated: false)
         }
-
+        
         messageCollectionViewBottomInset = newBottomInset
     }
 
@@ -88,17 +85,17 @@ extension MessagesViewController {
 
     internal var keyboardOffsetFrame: CGRect {
         guard let inputFrame = inputAccessoryView?.frame else { return .zero }
-        return CGRect(origin: inputFrame.origin,
-                      size: CGSize(width: inputFrame.width, height: inputFrame.height))
+        return CGRect(origin: inputFrame.origin, size: CGSize(width: inputFrame.width, height: inputFrame.height - iPhoneXBottomInset))
     }
 
-    /// On the iPhone with safe areas (e.g. X) the inputAccessoryView is anchored to the layoutMarginesGuide.bottom anchor
+    /// On the iPhone X the inputAccessoryView is anchored to the layoutMarginesGuide.bottom anchor
     /// so the frame of the inputAccessoryView is larger than the required offset
     /// for the MessagesCollectionView.
     ///
-    /// - Returns: The safeAreaInsets.bottom if its an iPhone with safe areas (e.g. X), else 0
-    internal var safeAreaBottomInset: CGFloat {
+    /// - Returns: The safeAreaInsets.bottom if its an iPhoneX, else 0
+    private var iPhoneXBottomInset: CGFloat {
         if #available(iOS 11.0, *) {
+            guard UIScreen.main.nativeBounds.height == 2436 else { return 0 }
             return view.safeAreaInsets.bottom
         }
         return 0
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -30,12 +30,14 @@ extension MessagesViewController {
 
     internal func addKeyboardObservers() {
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: .UIKeyboardDidChangeFrame, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleTextViewDidBeginEditing(_:)), name: .UITextViewTextDidBeginEditing, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.adjustScrollViewInset), name: .UIDeviceOrientationDidChange, object: nil)
     }
 
     internal func removeKeyboardObservers() {
         NotificationCenter.default.removeObserver(self, name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.removeObserver(self, name: .UIKeyboardDidChangeFrame, object: nil)
         NotificationCenter.default.removeObserver(self, name: .UITextViewTextDidBeginEditing, object: nil)
         NotificationCenter.default.removeObserver(self, name: .UIDeviceOrientationDidChange, object: nil)
     }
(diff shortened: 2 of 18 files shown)


Sources/Protocols/MessagesDataSource.swift
```diff
@@ -93,6 +93,14 @@ public protocol MessagesDataSource: AnyObject {
...
     func messageBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString?
 
+    /// Provide a chance to config a cell before it's presented. e.g. to customize the accessoryView.
+    ///
+    /// - Parameters:
+    ///   - cell: The `MessageContentCell` that can be configured.
+    ///   - message: The `MessageType` that will be displayed by this cell.
+    ///   - indexPath: The `IndexPath` of the cell.
+    ///
+    func configCell(_ cell: MessageContentCell, for message: MessageType, at indexPath: IndexPath)
```
SD10: I don't think this method fits well with this pull request 🤔

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -86,6 +90,9 @@ open class MessageContentCell: MessageCollectionViewCell {
         cellTopLabel.text = nil
         messageTopLabel.text = nil
         messageBottomLabel.text = nil
+        for view in accessoryView.subviews {
```
SD10: I'm not sure about this approach, to be honest. I don't like adding or removing views from the view hierarchy in `cellForRowAtIndexPath`. I think if you need to do this, it's a sign you need more cells or a better abstraction.

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -270,5 +278,25 @@ open class MessageContentCell: MessageCollectionViewCell {
...
         messageBottomLabel.frame = CGRect(origin: origin, size: attributes.messageBottomLabelSize)
     }
-    
+
+    /// Positions the cell's accessory view.
+    /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
+    open func layoutAccessoryView(with attributes: MessagesCollectionViewLayoutAttributes) {
+        guard attributes.accessoryViewSize != .zero else { return }
+
+        let y = (bounds.height - attributes.accessoryViewSize.height) / 2
```
SD10: Doesn't the `accessoryViewPadding` allow the user to adjust the y position here?

Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -81,6 +81,25 @@ public protocol MessagesLayoutDelegate: AnyObject {
...
 
+    /// Specifies the size for the `MessageContentCell`'s accessory view.
+    ///
+    /// - Parameters:
+    ///   - message: The `MessageType` that will be displayed for this cell.
+    ///   - messagesCollectionView: The `MessagesCollectionView` in which this cell will be displayed.
+    ///
+    /// - Note:
+    ///   The default value returned by this method is zero.
+    func accessoryViewSize(for message: MessageType, in messagesCollectionView: MessagesCollectionView) -> CGSize
```
SD10: I'm not sure if we want to make this a delegate method or to put it as a property of the layout object.
@zhongwuzw, this is a feature MessageKit should support so I'd like your advice?

Sources/Protocols/MessagesDataSource.swift
hyouuu: Well I know it feels too flexible, but couldn't figure out another way to customize an accessory view - thoughts?

Sources/Views/Cells/MessageContentCell.swift
hyouuu: Yes it is kind of arbitrary. Maybe providing a similar method to the configCell, a prepareForReuse call so client can do their customized preparation? Or better thoughts?

Sources/Views/Cells/MessageContentCell.swift
hyouuu: Good point. Will update

Sources/Protocols/MessagesLayoutDelegate.swift
hyouuu: I guess the delegate calls should be different by indexPath's section or item? I could make the change

hyouuu: Ah but by moving it to size calculator, we'd need another display delegate method like the

```swift
func configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)
```
, and then do something like 

 ```swift
if let layout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout {
            layout.textMessageSizeCalculator.incomingAvatarSize = CGSize(width: 52, height: 52)
            layout.textMessageSizeCalculator.outgoingAvatarSize = CGSize(width: 52, height: 52)
            layout.textMessageSizeCalculator.incomingAvatarPosition = AvatarPosition(vertical: .messageCenter)
            layout.textMessageSizeCalculator.outgoingAvatarPosition = AvatarPosition(vertical: .messageCenter)
        }
```
, which I personally feel pretty weird. Let me know what you think

File: Sources/Views/Cells/MessageContentCell.swift (hunk 3 above)
Comment (hyouuu) [reply Sources/Views/Cells/MessageContentCell.swift]: Updated

File: Sources/Protocols/MessagesLayoutDelegate.swift (hunk 4 above)
Comment (SD10) [reply Sources/Protocols/MessagesLayoutDelegate.swift]: The reason for this is I really wanted to move away from having users configure things with the delegates. I'd rather they abstract the logic for the layout of a cell into a `CellSizeCalculator` object. Then provide the layout object with the calculator.

IMO, doing it another way does not feel right. The API we're using here feels like it was designed for an earlier version of MessageKit and not version 1.0

File: Sources/Views/Cells/MessageContentCell.swift (hunk 2 above)
Comment (SD10) [reply Sources/Views/Cells/MessageContentCell.swift]: I mean I really don't want to provide any method for this. Where would `prepareForReuse` get called? They can just override `MessageContentCell` and do this themselves. I don't think we should have a method in the controller handling preparing the reuse of cells....

File: Sources/Protocols/MessagesDataSource.swift (hunk 1 above)
Comment (SD10) [reply Sources/Protocols/MessagesDataSource.swift]: All this method does is adds a layer of indirection for `cellForRowAtIndexPath`. Everyone subclasses `MessagesViewController` to use the library. They should just override `cellForRowAtIndexPath` and do this themselves

File: Sources/Views/Cells/MessageContentCell.swift (hunk 3 above)
Comment (SD10) [reply Sources/Views/Cells/MessageContentCell.swift]: I'm wondering what this looks like when we have varying label sizes, avatar sizes, etc. What would this look like if the avatar size is bigger than the message? What if there is a very large label? Lots of things to think about

File: Sources/Protocols/MessagesLayoutDelegate.swift [hunk 5]
```diff
@@ -81,6 +81,25 @@ public protocol MessagesLayoutDelegate: AnyObject {
...
+
+    /// Specifies the padding for the `MessageContentCell`'s accessory view.
+    ///
+    /// - Parameters:
+    ///   - message: The `MessageType` that will be displayed for this cell.
+    ///   - messagesCollectionView: The `MessagesCollectionView` in which this cell will be displayed.
+    ///
+    /// - Note:
+    ///   The default value returned by this method is zero.
+    func accessoryViewPadding(for message: MessageType, in messagesCollectionView: MessagesCollectionView) -> UIEdgeInsets
```
SD10: Padding would definitely need to be defined through the cell calculator objects and not via delegate

Sources/Protocols/MessagesDataSource.swift
SD10: Maybe we could add a method to configure the accessory view just like we're doing for the avatar 🤔

hyouuu: Let me try to mimic the way avatar is doing

hyouuu:
@SD10 updated and now it looks much like avatarView :D Let me know what you think!

SD10:
@hyouuu Thanks for making these changes 💯 This is _definitely_ heading in the right direction. I am still a little bit worried about the layout of the accessory view, that is something I'll need to test and review in more detail.

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -60,6 +60,9 @@ open class MessageContentCell: MessageCollectionViewCell {
         return label
     }()
 
+    // Should only add customized subviews - don't change accessoryView itself.
+    open var accessoryView: UIView = .init()
```
SD10: Can you change this to `open var accessoryView: UView = UIView()`?

SD10:
In addition to the layout of the `accessoryView` probably not being correct, this also changes the height calculations for the size of the cell. There are some scenarios where someone could set an `accessoryView` that is taller than the rest of the message content.

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -270,5 +277,27 @@ open class MessageContentCell: MessageCollectionViewCell {
 
         messageBottomLabel.frame = CGRect(origin: origin, size: attributes.messageBottomLabelSize)
     }
-    
+
+    /// Positions the cell's accessory view.
+    /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
+    open func layoutAccessoryView(with attributes: MessagesCollectionViewLayoutAttributes) {
+        guard attributes.accessoryViewSize != .zero else { return }
```
SD10: We should remove this line and set the frame regardless. Since cells are re-used, this will cause bugs. I need to fix these on the other layout methods in another PR.

Sources/Views/Cells/MessageContentCell.swift
hyouuu: sure

Sources/Views/Cells/MessageContentCell.swift
hyouuu: I removed this line alongside other zero checks - let me know if you want me to revert the change for others :)

Sources/Protocols/MessagesLayoutDelegate.swift
zhongwuzw: @SD10 , My some thoughts, wether we can provide a flag to not use `accessoryView`, not just set size to `.zero`, because I think many of the time, maybe user don't need `accessoryView` at all.

SD10: @zhongwuzw what if we just make the default size for it `.zero`?

zhongwuzw: @SD10 , `.zero` can invisible to user, I mean it would leads to unnecessary calculations, render pass (yeah, it's true, though it's invisible to user, it still has these works). I'm a performancer 😢 .
I wonder wether we need add a boolean to control this.

zhongwuzw: Maybe later we would add more feature views, I don't think we add them all on cell is a good approach. 😂 .

SD10: @zhongwuzw this is true 👍 but I've learned not to prematurely optimize. I only optimize if someone complains and then if we do benchmarks afterward. We do cell layout using frames and not auto-layout, so we already save so much. This is likely the last view we will support. I will not add views that don't exist in JSQMVC

hyouuu:
@zhongwuzw for some reason I couldn't find the comments you 2 added (but received emails) on GitHub :/ 

I'm on the side with @SD10 that we probably don't want to optimize for the performance of this specific case at the moment, but later on we could add more flexible switches not only for accessoryView, but also for cellTopLabel & cellBottomLabel, which I guess many people don't need either

SD10:
@hyouuu That's a good suggestion, but as I said to @zhongwuzw I'm really not concerned. In previous versions of MessageKit it was taking ~6 seconds to lay out 10,000 cells. We've decreased that to probably around ~1 second. Realistically, no one should have that many cells on screen at a time. So until we have people complaining about performance, I'm not going to prematurely optimize anything

PS. You can probably find the comments from @zhongwuzw in the code review above. Hit "show outdated" from the drop down.

hyouuu:
Regarding the tests, do you use Carthage to install Quick etc? Haven't used it and couldn't build the tests directly. Let me know if you want to test it yourself and correct the tests or want me to do it ;)

SD10:
Yeah @hyouuu we use Carthage to install the test dependencies. You’ll need to run `carthage bootstrap`

hyouuu:
Turned out the failed test was not related to this PR, but that we had `.natural` for avatar horizontal position, so set a default `.cellLeading`; also fixed a few formatting things from warnings

hyouuu:
Added one more commit to fix keyboard frame & insets as I was facing issues in my app - let me know if you actually want it to be in a separate PR - if so I'll remove it from my current fork right before we can push the current PR (as I need to use my fork to keep things working ;)

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -30,12 +30,14 @@ extension MessagesViewController {
 
     internal func addKeyboardObservers() {
         NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: .UIKeyboardWillChangeFrame, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: .UIKeyboardDidChangeFrame, object: nil)
```
hyouuu: Note: only listening to UIKeyboardWillChangeFrame will cause issues in certain situations where keyboard was in the process to hide/show and then we get into our msgCollectionView, then we missed the willChange event and didn't catch up with didChange event

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -85,17 +88,17 @@ extension MessagesViewController {
...
+    /// On the iPhone with safe areas (e.g. X) the inputAccessoryView is anchored to the layoutMarginesGuide.bottom anchor
     /// so the frame of the inputAccessoryView is larger than the required offset
     /// for the MessagesCollectionView.
     ///
-    /// - Returns: The safeAreaInsets.bottom if its an iPhoneX, else 0
-    private var iPhoneXBottomInset: CGFloat {
+    /// - Returns: The safeAreaInsets.bottom if its an iPhone with safe areas (e.g. X), else 0
+    internal var safeAreaBottomInset: CGFloat {
         if #available(iOS 11.0, *) {
-            guard UIScreen.main.nativeBounds.height == 2436 else { return 0 }
```
hyouuu: This doesn't make sense and is hacky - if a device has safe bottom, then we should just use it, no matter if it's an iPhone X or a future iPhone X+ or a Safe iPad

Sources/Controllers/MessagesViewController.swift
```diff
@@ -82,25 +81,24 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
...
+        messageCollectionViewBottomInset = keyboardOffsetFrame.height - safeAreaBottomInset
+        messagesCollectionView.scrollToBottom(animated: true)
     }
-    
+
     open override func viewDidLayoutSubviews() {
-        // Hack to prevent animation of the contentInset after viewDidAppear
-        if isFirstLayout {
-            defer { isFirstLayout = false }
-            addKeyboardObservers()
```
hyouuu: Since we unobserve in `deinit`, we should pair and start observation in init or viewDidLoad - not sure why we would want to do it here, so corrected

Sources/Controllers/MessagesViewController.swift
```diff
@@ -82,25 +81,24 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
...
     open override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         isMessagesControllerBeingDismissed = true
+        messageCollectionViewBottomInset = keyboardOffsetFrame.height - safeAreaBottomInset
+        messagesCollectionView.scrollToBottom(animated: true)
     }
-    
+
     open override func viewDidLayoutSubviews() {
-        // Hack to prevent animation of the contentInset after viewDidAppear
```
hyouuu: So, moved the initial bottomInset setting to viewDidAppear lol

SD10: This is too late to call it. It won’t look how I want it. I’d appreciate it if you keep this PR in scope. It makes it hard to approve because now I can reject it for keyboard related things

Sources/Controllers/MessagesViewController+Keyboard.swift
SD10: It is a hack but the logic you changed it to doesn’t work

Sources/Controllers/MessagesViewController.swift
SD10: I can’t really see the diff well for this in GitHawk, but certain keyboard events we only observe once to get the behavior we want

hyouuu:
Cool I reverted the changes in this branch, and will use another one for my project atm :) Let's figure that issue out after this one @SD10

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -270,5 +267,25 @@ open class MessageContentCell: MessageCollectionViewCell {
 
         messageBottomLabel.frame = CGRect(origin: origin, size: attributes.messageBottomLabelSize)
     }
-    
+
+    /// Positions the cell's accessory view.
+    /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
+    open func layoutAccessoryView(with attributes: MessagesCollectionViewLayoutAttributes) {
+        var y = (bounds.height - attributes.accessoryViewSize.height) / 2
+        y -= attributes.accessoryViewPadding.vertical + attributes.accessoryViewPadding.top
```
zhongwuzw, line 275: `y` maybe should align to the `center.y` of the `bubble`?
cc @SD10

SD10: yeah @zhongwuzw this is one of the reasons why I'm delaying this. everything seems to look good except the calculations for where it is located. we also have to think about having a large accessoryView but then the rest of the cell content is very small

SD10: this also cannot go in 1.0, it is a 2.0 feature for sure

hyouuu: Does bubble refer to messageContainerView? If so I could make the change

SD10: @hyouuu it's not necessarily that simple. what if the `accessoryView` is set to a height of 30pts, but the size of the other labels + the size of the message is only 20 points? if you try and center it to the `messageContainerView` your `accessoryView` will now leak out of the cell. there are a lot of edge cases here

hyouuu: @SD10 doesn't these lines https://github.com/MessageKit/MessageKit/pull/710/files#diff-ed5b75922e64407792c5a2dc5c27ebb6R107 cover the accessoryView height greater than the rest cases?

SD10: @hyouuu You're right this could work, but I'll need to test to make sure

hyouuu: Thanks! Let me know if there's anything I can help with :D

SD10 (changes requested):
@hyouuu Can you please add a CHANGELOG entry for this under the **Added** section?

hyouuu:
Done! @SD10

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 - Added convenience method to set properties for all `MessageSizeCalculator`s.
 [#697](https://github.com/MessageKit/MessageKit/pull/697) by [@zhongwuzw](https://github.com/zhongwuzw).
 
+- Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
```
SD10: @hyouuu You will have to pull in the latest changes from `master`. The CHANGELOG is outdated now

hyouuu:
Ah sorry - should be good now @SD10

SD10 (approved):

SD10:

hyouuu:
It's an honor!! Thank you :D

charvoa:
@SD10  when do you think this feature will be merged with master ?

SD10:
@charvoa Not sure unfortunately. A rough estimation would be 2-4 weeks. Right now my company is using a fork of MessageKit. That's why development has been so slow lately. Next week we're going to try to get off the fork and back to the main project.

PR #716, opened 2018-06-07
Fixed MessageContentCell's subviews frame when size equal to .zero

zhongwuzw:


Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -171,7 +171,10 @@ open class MessageContentCell: MessageCollectionViewCell {
     /// Positions the cell's `AvatarView`.
     /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
     open func layoutAvatarView(with attributes: MessagesCollectionViewLayoutAttributes) {
-        guard attributes.avatarSize != .zero else { return }
+        guard attributes.avatarSize != .zero else {
```
SD10: Actually, we want to just remove these lines completely. We still need to calculate the `X` and `Y` positions

SD10: No early exit

zhongwuzw: 🤔 What's the reason? I think each subview did not depend on other's `frame`.

SD10: @zhongwuzw they do, I lay views out relative to each other, so we have to set X and Y

zhongwuzw: Oops, I didn't look carefully.

Issue #730, opened 2018-06-15
Replace the MessageInputBar with our new MessageInputBar library


SD10:
We have decided to separate the [MessageInputBar](https://github.com/MessageKit/MessageInputBar) out into its own library. @nathantannar4 will be the lead maintainer on that project. This allows us to provide other optional subspecs to MessageKit project such as autocomplete and attachment managers without bloating the original project. 

This will be available in MessageKit 2.0 as it is a heavy new feature with API breaking changes.

cc @nathantannar4 can you submit a PR against the MessageKit `development` branch making this addition?

nathantannar4:
Will do

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #732, opened 2018-06-15
Integrate MessageInputBar as a dependency

nathantannar4:
This PR removes all source code / test related to MessageInputBar and replaces by setting `MessageInputBar/Core` as a new dependency.  This will also for separate development of each library without bloating MessageKit through the additions such as autocomplete (available in `MessageInputBar/AutocompleteManager`

If there are any Cartfile users here please let me know if the changes I made to the Cartfiles are correct.


nathantannar4:
For #730

Cartfile.private
```diff
@@ -1,2 +1,3 @@
 github "Quick/Quick" ~> 1.2.0
 github "Quick/Nimble" ~> 7.0.3
+github "MessageKit/MessageInputBar" ~> 0.2.0
```
SD10: This would need to be declared in a `Cartfile`, not `Cartfile.private`

SD10:
@nathantannar4 We'll also need to update the Circle CI configuration to build the dependencies. You can see how we do it at Moya

SD10:
@nathantannar4 Want me to take over this one?

nathantannar4:
@SD10 Ya I don't know why the build is failing, something to do with the Carthage things I think

SD10:
If any @MessageKit/contributors want to pick this one up that would be 🔥 😊 Need to link this in a way that supports both CocoaPods and Carthage

zhongwuzw:
Emm, what about make `MessageInputBar` as a git submodule?

SD10:
@zhongwuzw I'm not familiar with git submodules. I'd like to set up this project the same way as something like Moya

ndonald2:
Carthage should allow specifying other dependencies without using a submodule. My guess is that this is why it's failing (input bar dependency in the private, not public, cartfile)

https://github.com/MessageKit/MessageKit/pull/732#discussion_r195914957

SD10:
@nathantannar4 We need to create a new file `Cartfile` and list the dependency there. The `Cartfile.private` is used for dev and `Cartfile` for release

Check out the [Cartfile](https://github.com/Moya/Moya/blob/master/Cartfile) at Moya

nathantannar4:
@SD10 we do have a `Cartfile`

nathantannar4:
It seems to build it correctly now, some random test error occurred though

SD10:
@nathantannar4 There is no Cartfile

https://github.com/MessageKit/MessageKit/blob/split-messageinputbar/Cartfile

nathantannar4:
Works now

zhongwuzw (approved):
Good 👍 ,waiting for `2.0`

SD10:
Will review this weekend. I’d like to see this make it in 2.0. We’ll have to do a beta for a change this big

SD10 (approved):
Ok looks like we won't be able to do a 1.1 release but this is good to go and I need it 😓

Issue #761, opened 2018-07-14
messagesCollectionView has incorrect bottom inset when inside smaller container view


davidvpe:
### General Information

Created a ChatViewController that inherits from MessagesViewController and add it's view to smaller view inside a parent view controller, so the Chat View is actually smaller than the device. This makes the bottom content inset act weird and doesn't let me see any new message that arrives or even my own last message.

- MessageKit Version:
1.0.0

- iOS Version(s):
11.4

- Swift Version:
4

- Devices/Simulators:
All simulator devices (my own iPhone 7 does it too)

- Reproducible in ChatExample? (Yes/No):
No, because in order to reproduce it you have to present the Chat View Controller as a smaller view

### What happened?

![ezgif-5-f92089b9c5](https://user-images.githubusercontent.com/1934159/42723239-92b8bdbc-875a-11e8-8a5f-de710235255b.gif)

### What did you expect to happen?

I expect the bottom content inset to be exactly the size of the keyboard not more not less.

### My workaround

Right now, I'm altering the handleKeyboardDidChangeState function, instead of this line:

`let newBottomInset = view.frame.height - keyboardEndFrame.minY - iPhoneXBottomInset`

I am doing this:

`let convertedKeyboardEndFrame = messagesCollectionView.convert(keyboardEndFrame, from: nil)`
`let intersect = convertedKeyboardEndFrame.intersection(messagesCollectionView.bounds)`
`let newBottomInset = intersect.size.height`

SD10:
I'm going to label this a bug because it seems like the changes we made to the keyboard have caused a regression in container view controllers. Maybe we need to add one to the example app to check @nathantannar4?

nathantannar4:
@SD10 good point, I'll work it into the new example app

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

nathantannar4:
Hmm, @davidvpe Im building a new example app and when I create a custom container the bottom inset is correct. I am doing:

```
final class ConversationContainerController: UIViewController {
    
    let conversationViewController = ConversationViewController()
    
    let bannerView = UIView()
    
    override var canBecomeFirstResponder: Bool {
        return conversationViewController.canBecomeFirstResponder
    }
    
    override var inputAccessoryView: UIView? {
        return conversationViewController.inputAccessoryView
    }
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        conversationViewController.willMove(toParentViewController: self)
[...]

PR #762, opened 2018-07-14
Updated way to calculate the bottom content inset

davidvpe:
What does this implement/fix? Explain your changes.
-----------------------------------------------------
Modified the way MessagesViewController calculate the Collection View bottom content inset when showing the keyboard

Does this close any currently open issues?
------------------------------------------
https://github.com/MessageKit/MessageKit/issues/761

Any relevant logs, error output, etc?
-------------------------------------
No

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Devices/Simulators:** All simulators + Real iPhone 7

**iOS Version:** 11.3

**Swift Version:** 4.0

**MessageKit Version:** 1.0.0

Issue #755, opened 2018-07-09
Issue with MessagesViewController as child view controller and scrolling


ranhsd:
Hi, In one of my view controller in my project I need to display MessagesViewController below some custom UIView. What I did is in storyboard xib I created 2 UIView which act as content container for both the custom header view and the messages view controller view. Until today I used version 0.13.5 of MessageKit and everything worked as expected but today I upgrade the library from 0.13.5 to 1.0.0 and suddenly MessageInputBar hide the content inside my MessageCollectionView and I also cannot scroll to the end of the collection view. 

This is how I add MessagesViewController as child view: (ChatViewController is a subclass of MessageViewController and **chatContainerView** is the container of it.

```
    lazy var chatViewController : ChatViewController! = {[weak self] in
        guard let strongSelf = self else {return nil}
        let chatViewController = ChatViewController()
        chatViewController.setInputBarStyle()
        chatViewController.messagesCollectionView.messagesDataSource = strongSelf
        chatViewController.messagesCollectionView.messagesLayoutDelegate = strongSelf
        chatViewController.messagesCollectionView.messagesDisplayDelegate = strongSelf
        chatViewController.messageInputBar.delegate = strongSelf
        strongSelf.addChildViewController(chatViewController)
        chatViewController.view.frame = strongSelf.chatContainerView.bounds
        strongSelf.chatContainerView.addSubview(chatViewController.view)
        chatViewController.didMove(toParentViewController: strongSelf)

        return chatViewController
        }()
```

Somebody have a clue why I have this issue after upgrading to 1.0? 

Thanks.

SD10:
Hey @ranhsd,
We changed the bottom inset calculation in 1.0.0 -- see #707 -- maybe this will help you resolve it.
Do you think you may need to account for some padding on the `collectionView.contentInset.top`?

ranhsd:
Hi @SD10 I think that this change may work for part of the scenarios but not to all of them. a good option is to be able to modify the contentInset (bottom part) of the messages collection view so the scroll will have some padding

SD10:
@ranhsd Have you tried `extendedLayoutIncludesOpaqueBars = true` property of `MessagesViewController`? Just 2 days ago I finally upgraded my app to MessageKit 1.0 -- and I set to set this to `true` to fix the same bug you are experiencing. 

I will look into giving users better control of the bottom inset 😢

ranhsd:
[...]

PR #787, opened 2018-07-30
Adjust contentInset for the keyboard frame properly

andreyvit:
I know that modifying the keyboard handling logic is all the rage these days, and wanted to get in on the action. :-) More seriously, I've been figuring out some keyboard handling issues in my app with a more complex structure of view controllers, and produced this PR to fix the issues found.

I believe this PR makes the code cleaner and more logical. It handles undocked keyboards and a bunch of other scenarios, and does *not* introduce weird hacks, animation blocks and the like.

Commit desciption follows:

This change provides better handling of keyboard frames. It copes with different values of UIViewController.edgesForExtendedLayout, UIScrollView.contentInsetAdjumentBehavior, view controller containment scenarios and undocked keyboards, and uses cleaner code.

Improvements and fixes in this change:

1. Translates keyboard frame from screen coordinates into view controller coordinates. They aren’t the same thing when containment is involved, or edgesForExtendedLayout does not include .top.

2. Avoids special casing any iPhone models and avoids dealing with safeAreaInsets manually. In iOS 11, UIScrollView may behave differently depending on contentInsetAdjumentBehavior, and the old code incorrectly assumed that the mode is .always (while it is in fact is .automatic) and that no extra safe area insets have been introduced at the bottom edge. Best part, we don’t have to deal with any of this at all, we can offload the logic to UIScrollView itself by looking at the difference between adjustedContentInset and contentInset.

3. Handles view controller containment and undocked keyboards by taking into account an intersection of the scroll view frame and the keyboard’s frame. I think Apple demoes this in one of the WWDC videos, but I couldn’t find which one (or if I’m misremembering). Regardless, this technique makes sure we compute and use a sensible number.

#296 and likely #761.

Where has this been tested?
---------------------------

**Devices/Simulators:** Device: iPhone 8, Simulators: iPhone 8, iPhone X, iPhone SE, iPhone 6 Plus, iPad Pro 10", iPad Air 2.

**iOS Version:** 11.4, 10.3.1, 9.3

**Swift Version:** 4

**MessageKit Version:** on top of commit 57165b7, development branch, on top of 1.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
 
+- Added `extraBottomInset` property that allows to adjust the bottom content inset automatically set on the messages collection view by the view controller. [#787](https://github.com/MessageKit/MessageKit/pull/787) by [@andreyvit](https://github.com/andreyvit)
+
+### Fixed
+
+- Fixed bottom content inset adjustment when using an undocked keyboard on iPad, or when `edgesForExtendedLayout` does not include `.top`, or when a parent container view controller adds extra views at the top of the screen. [#787](https://github.com/MessageKit/MessageKit/pull/787) by [@andreyvit](https://github.com/andreyvit)
+
+
 ## [1.0.0](https://github.com/MessageKit/MessageKit/releases/tag/1.0.0)
 
 - First major release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
 
-- Added `extraBottomInset` property that allows to adjust the bottom content inset automatically set on the messages collection view by the view controller. [#787](https://github.com/MessageKit/MessageKit/pull/787) by [@andreyvit](https://github.com/andreyvit)
+- Added `additionalBottomInset` property that allows to adjust the bottom content inset automatically set on the messages collection view by the view controller. [#787](https://github.com/MessageKit/MessageKit/pull/787) by [@andreyvit](https://github.com/andreyvit)
 
 ### Fixed
 
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -115,15 +115,15 @@ extension MessagesViewController {
         if intersection.isNull || intersection.maxY < messagesCollectionView.frame.maxY - coordinatePrecision /* never compare floats without a tolerance */ {
             // The keyboard is hidden, is a hardware one, or is undocked and does not cover the bottom of the collection view.
             // Note: intersection.maxY may be less than messagesCollectionView.frame.maxY when dealing with undocked keyboards.
-            return 0
+            return max(0, extraBottomInset - automaticallyAddedBottomInset)
         } else {
-            return max(0, intersection.height - automaticallyAddedBottomInset)
+            return max(0, intersection.height + extraBottomInset - automaticallyAddedBottomInset)
         }
     }
 
     internal func requiredInitialScrollViewBottomInset() -> CGFloat {
         guard let inputAccessoryView = inputAccessoryView else { return 0 }
-        return max(0, inputAccessoryView.frame.height - automaticallyAddedBottomInset)
+        return max(0, inputAccessoryView.frame.height + extraBottomInset - automaticallyAddedBottomInset)
     }
 
     /// iOS 11's UIScrollView can automatically add safe area insets to its contentInset,
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -62,6 +62,21 @@ extension MessagesViewController {
             return
         }
         
+        // Note that the check above does not exclude all notifications from an undocked keyboard, only the weird ones.
+        //
+        // We've tried following Apple's recommended approach of tracking UIKeyboardWillShow / UIKeyboardDidHide and ignoring frame
+        // change notifications while the keyboard is hidden or undocked (undocked keyboard is considered hidden by those events).
+        // Unfortunately, we do care about the difference between hidden and undocked, because we have an input bar which is at the
+        // bottom when the keyboard is hidden, and is tied to the keyboard when it's undocked.
+        //
+        // If we follow what Apple recommends and ignore notifications while the keyboard is hidden/undocked, we get an extra inset
+        // at the bottom when the undocked keyboard is visible (the inset that tries to compensate for the missing input bar).
+        // (Alternatives like setting newBottomInset to 0 or to the height of the input bar don't work either.)
+        //
+        // We could make it work by adding extra checks for the state of the keyboard and compensating accordingly, but it seems easier
+        // to simply check whether the current keyboard frame, whatever it is (even when undocked), covers the bottom of the collection
+        // view.
+        
         guard let keyboardEndFrameInScreenCoords = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }
         let keyboardEndFrame = view.convert(keyboardEndFrameInScreenCoords, from: view.window)
         
@@ -98,7 +113,9 @@ extension MessagesViewController {
         let intersection = messagesCollectionView.frame.intersection(keyboardFrame)
         
         if intersection.isNull || intersection.maxY < messagesCollectionView.frame.maxY - 1e-6 /* never compare floats without a tolerance */ {
-            return 0  // keyboard is undocked/hardware, or just does not cover the bottom of the collection view
+            // The keyboard is hidden, is a hardware one, or is undocked and does not cover the bottom of the collection view.
+            // Note: intersection.maxY may be less than messagesCollectionView.frame.maxY when dealing with undocked keyboards.
+            return 0
         } else {
             return max(0, intersection.height - automaticallyAddedBottomInset)
         }
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -93,7 +93,8 @@ extension MessagesViewController {
     }
 
     private func computeScrollViewBottomInset(forKeyboardFrame keyboardFrame: CGRect) -> CGFloat {
-        // we only need to adjust for the part of the keyboard that covers (i.e. intersects) our collection view
+        // we only need to adjust for the part of the keyboard that covers (i.e. intersects) our collection view;
+        // see https://developer.apple.com/videos/play/wwdc2017/242/ for more details
         let intersection = messagesCollectionView.frame.intersection(keyboardFrame)
         
         if intersection.isNull || intersection.maxY < messagesCollectionView.frame.maxY - 1e-6 /* never compare floats without a tolerance */ {
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -52,6 +52,8 @@ extension MessagesViewController {
 
     @objc
     private func handleKeyboardDidChangeState(_ notification: Notification) {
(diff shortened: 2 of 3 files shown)


Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -52,12 +52,20 @@ extension MessagesViewController {
...
+            // ignore this notification.
+            return
+        }
         
-        guard !isMessagesControllerBeingDismissed else { return }
+        guard let keyboardEndFrameInScreenCoords = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }
+        let keyboardEndFrame = view.convert(keyboardEndFrameInScreenCoords, from: view.window)
         
-        let newBottomInset = view.frame.height - keyboardEndFrame.minY - iPhoneXBottomInset
+        guard !isMessagesControllerBeingDismissed else { return }
```
nathantannar4: Lets keep this above the `keyboardEndFrame` calculations as that extra computing time isn't needed if we are just going to exit out. I would place this at the top of the function.

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -81,23 +92,32 @@ extension MessagesViewController {
         }
     }
 
-    // MARK: - Helpers
+    private func computeScrollViewBottomInset(forKeyboardFrame keyboardFrame: CGRect) -> CGFloat {
```
nathantannar4: For consistency with other functions in the project, could this function name be changed to `requiredScrollViewBottomInset(for keybaordFrame: CGRect) -> CGFloat`

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -68,8 +76,11 @@ extension MessagesViewController {
         messageCollectionViewBottomInset = newBottomInset
```
nathantannar4, line 91: We also wanted to add a new property to `MessagesViewController` so users can add additional inset. Something like, `open var additionalBottomInset: CGFloat = 0` and then add this with `newBottomInset`

nathantannar4 (changes requested):
Thanks for contributing! Im going to test this in my app

nathantannar4:
From my tests, this seems to work on all iPhone models. Don't have an iPad to test the split keyboard.

zhongwuzw:
@nathantannar4 In #762 , you said it has issue when use `iPhoneX`, but the `PR` author said it's ok. #762 don't take the bottom inset into account. Would you figure out any reason?

andreyvit:
@zhongwuzw In #762, the additional auto-added inset isn't taken into account, so it will have issues on iPhone X where safe area insets are non-zero.

andreyvit:
@nathantannar4 Thanks for the review, will address your points.

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -52,12 +52,20 @@ extension MessagesViewController {
 
     @objc
     private func handleKeyboardDidChangeState(_ notification: Notification) {
-        guard let keyboardEndFrame = notification.userInfo?[UIKeyboardFrameEndUserInfoKey] as? CGRect else { return }
+        guard let keyboardStartFrameInScreenCoords = notification.userInfo?[UIKeyboardFrameBeginUserInfoKey] as? CGRect else { return }
+        guard !keyboardStartFrameInScreenCoords.isEmpty else {
+            // WORKAROUND for what seems to be a bug in iPad's keyboard handling in iOS 11: we receive an extra spurious frame change
+            // notification when undocking the keyboard, with a zero starting frame and an incorrect end frame. The workaround is to
+            // ignore this notification.
```
SD10, line 61: Interesting. Nice trick

SD10:
Hey @nathantannar4, when you did your tests were you just using the MessageKit example app or did you also look at child controllers?

Sources/Controllers/MessagesViewController+Keyboard.swift
SD10: Yeah I think this is a really good idea as a fallback if a user needs to adjust things on their own

nathantannar4:
I did so using the new example app I have which uses child controllers

SD10:
@zhongwuzw Any input here? I've yet to test it myself but I'm pretty confident in this approach

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -81,23 +92,32 @@ extension MessagesViewController {
         }
     }
 
-    // MARK: - Helpers
+    private func computeScrollViewBottomInset(forKeyboardFrame keyboardFrame: CGRect) -> CGFloat {
+        // we only need to adjust for the part of the keyboard that covers (i.e. intersects) our collection view
+        let intersection = messagesCollectionView.frame.intersection(keyboardFrame)
+        
+        if intersection.isNull || intersection.maxY < messagesCollectionView.frame.maxY - 1e-6 /* never compare floats without a tolerance */ {
```
zhongwuzw: What's the purpose here to compare `intersection.maxY < messagesCollectionView.frame.maxY - 1e-6`?
Is this situation, `intersection.maxY == min(messagesCollectionView.frame.maxY, keyboardFrame.maxY`, I think we only need to check `isNull`, if it fails, go to `else` statement.

andreyvit: The equality does not always hold. The undocked keyboard still generates frame change notifications, and they go through the same code, so it's important to only apply the inset when the keyboard covers the bottom of the collection view.

I do agree that the check feels a bit weird. I've tried a different approach, one recommended by Apple: track whether the keyboard is visible using show/hide events, and ignore frame changes when it's not visible. (The undocked keyboard is considered hidden.) Unfortunately, I couldn't get this to work properly, because the input bar is a keyboard accessory, so it's crucial to differentiate between 2 cases — the keyboard being hidden (and the input bar being at the bottom) and the keyboard being undocked (and the input bar being tied to the keyboard). It turns out that just using the keyboard frame in all cases allows us to handle them all transparently without needing to branch.

I'll add a comment about this to the code.

zhongwuzw: @andreyvit  So what's problem if we remove `intersection.maxY < messagesCollectionView.frame.maxY - 1e-6` check?

andreyvit:
The new update should address all the issues above. Introduced `extraBottomInset` property, moved code around and renamed functions as suggested by Nathan, added more explanations for the points raised by Wu Zhong, and filled in a `CHANGELOG`. Let me know if there's more I should change/add/explain. Also, do you want the commits squashed before merging?

SD10:
@andreyvit I see that you named the property `extraBottomInset`, can we rename it `additionalBottomInset`? The reason is that's the same name JSQMVC uses and I'd like to mirror that API

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Added customizable `accessoryView`, with a new `MessagesDisplayDelegate` function `configureAccessoryView`, and corresponding size & padding properties in `MessageSizeCalculator`. [#710](https://github.com/MessageKit/MessageKit/pull/710) by [@hyouuu](https://github.com/hyouuu)
 
+- Added `extraBottomInset` property that allows to adjust the bottom content inset automatically set on the messages collection view by the view controller. [#787](https://github.com/MessageKit/MessageKit/pull/787) by [@andreyvit](https://github.com/andreyvit)
```
SD10: Just a friendly reminder to rename this as well

Sources/Controllers/MessagesViewController.swift
```diff
@@ -59,6 +61,18 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
...
 
+    /// A CGFloat value that adds to (or, if negative, subtracts from) the automatically
+    /// computed value of `messagesCollectionView.contentInset.bottom`. Meant to be used
+    /// as a measure of last resort when the built-in algorithm does not produce the right
+    /// value for your app. Please let us know when you end up having to use this property.
+    open var extraBottomInset: CGFloat = 0 {
+        didSet {
+            let delta = extraBottomInset - oldValue
+            guard abs(delta) > coordinatePrecision else { return }
+            messageCollectionViewBottomInset += delta
```
SD10: What's the reasoning behind not just applying the `newValue` to the bottom inset directly? I'm concerned this is unexpected behavior

Sources/Controllers/MessagesViewController+Keyboard.swift
SD10: Re-raising this comment so it's not hidden

SD10: @zhongwuzw asks:
> @andreyvit 
>  So what's problem if we remove `intersection.maxY < messagesCollectionView.frame.maxY - 1e-6` check?

Re-raising this comment

I really hate GitHub?

Sources/Controllers/MessagesViewController.swift
```diff
@@ -59,6 +61,18 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
         return false
     }
 
+    /// A CGFloat value that adds to (or, if negative, subtracts from) the automatically
+    /// computed value of `messagesCollectionView.contentInset.bottom`. Meant to be used
+    /// as a measure of last resort when the built-in algorithm does not produce the right
+    /// value for your app. Please let us know when you end up having to use this property.
+    open var extraBottomInset: CGFloat = 0 {
```
SD10: Do you think we can support an `additionalTopInset` as well? This can be used to allow users to overlay a view over the `MessagesCollectionView`. See #755

Sources/Controllers/MessagesViewController.swift
andreyvit: @SD10 Because the newValue here is the _additional_ bottom inset, not the entire bottom inset. :-)

SD10:
I've tested this and it seems to be working well -- the only problem is when you modally present another controller the bottom inset seems to be off if the keyboard isn't showing

Sources/Controllers/MessagesViewController.swift
zhongwuzw: @SD10 🤔 Emm, how can we use `additionalTopInset`? If user set `additionalTopInset`, how we handle this? add to `contentInset`? maybe I think we don't need this currently.

Sources/Controllers/MessagesViewController.swift
```diff
@@ -59,6 +61,18 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
...
+        didSet {
+            let delta = extraBottomInset - oldValue
+            guard abs(delta) > coordinatePrecision else { return }
```
zhongwuzw: @andreyvit Seems like you are very careful to the comparison of `double` type. 😂 TBO these things maybe not needed, I think `1e-6` is only applied to compare the equality of two `double`, in our situations, we don't need accurate precision, just get the delta and assign to the `messageCollectionViewBottomInset`.

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -81,23 +107,35 @@ extension MessagesViewController {
         }
     }
 
-    // MARK: - Helpers
+    private func requiredScrollViewBottomInset(forKeyboardFrame keyboardFrame: CGRect) -> CGFloat {
+        // we only need to adjust for the part of the keyboard that covers (i.e. intersects) our collection view;
+        // see https://developer.apple.com/videos/play/wwdc2017/242/ for more details
+        let intersection = messagesCollectionView.frame.intersection(keyboardFrame)
+        
+        if intersection.isNull || intersection.maxY < messagesCollectionView.frame.maxY - coordinatePrecision /* never compare floats without a tolerance */ {
```
zhongwuzw: `intersection.isNull` has take the precision into consideration. `intersection.maxY < messagesCollectionView.frame.maxY - coordinatePrecision` is not needed.

zhongwuzw:
@andreyvit Thanks for your great `PR`, I removed the precision when compare double(If you think it would leads to problem, please let me know), and fix the insets calculation issue when present then dismiss vc.

SD10 (approved):
Thanks @andreyvit @zhongwuzw @nathantannar4 💯 🥇

SD10:

ljs19923:
So which pod versions i need to install to have this update ? thanks @SD10

SD10:
I'll be doing a MessageKit 1.1 release in a couple hours @JulienLevallois

ljs19923:
Okay thanks. Because i tried to install "development" branch 2 min ago, 
and i have the same bug again :/ Its normal ?

andreyvit:
Hey, thanks for merging this. Do you still want me to do any changes you mentioned?

Regarding comparisons, I'm not sure if it's _currently_ going to be a problem or not, but it's a rule when comparing floating-point numbers: never rely on them being exactly the same (or being exactly less or greater). It absolutely can be a problem in practice, although it's doesn't _often_ occur in UI code because UI code doesn't do much arithmetics.

zhongwuzw:
@andreyvit Thanks for your `PR`, we'll do a release after SD10 checks everything is ok.

SD10:
Yeah it looks like there is still issues, could be related to the `contentInset.top` code now. Honestly, getting the inset to work properly for everyone is so frustrating. Debating starting from scratch

ljs19923:
Is this problem now fixed?
Thanks @SD10

Issue #770, opened 2018-07-17
Add MessageCellDelegate method for didTapAccessoryView


SD10:
We need to add another delegate method that handles touches in the accessory view

SD10:
@hyouuu Want to take a look at this? It's blocking the 1.1 release

hyouuu:
Super busy lately - will work on it when get a chance!

SD10:
@MessageKit/contributors This is a good issue for people looking to contribute, it's holding up the next release 😉

nathantannar4:
Ill work on this. I added the delegate method but I think the layout of the accessory view should be tweaked. Currently we are setting it to the middle of the cell but I think we should be aligning it to the middle of the `MessageContainerView`

SD10:
@nathantannar4 Thanks 🙏 I agree changing the alignment makes sense. We’ll have to review the layout and sizing logic if we make this change

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

nathantannar4:
Resolved!

PR #834, opened 2018-08-31
Add didTapAccessoryView Delegate method and adjust layout

nathantannar4:


nathantannar4:
Solves #770

nathantannar4:
simulator screen shot - iphone x - 2018-08-31 at 11 00 47
simulator screen shot - iphone x - 2018-08-31 at 10 57 48

Example/Sources/View Controllers/AdvancedExampleViewController.swift
```diff
@@ -311,6 +316,14 @@ extension AdvancedExampleViewController: MessagesDisplayDelegate {
         avatarView.layer.borderColor = UIColor.primaryColor.cgColor
     }
     
+    func configureAccessoryView(_ accessoryView: UIView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) {
+        let button = UIButton(type: .infoLight)
+        button.tintColor = .primaryColor
+        accessoryView.addSubview(button)
```
SD10, line 332: This subview will have to be removed somewhere because cells are reused?

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -271,7 +273,10 @@ open class MessageContentCell: MessageCollectionViewCell {
     /// Positions the cell's accessory view.
     /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
     open func layoutAccessoryView(with attributes: MessagesCollectionViewLayoutAttributes) {
-        var y = (bounds.height - attributes.accessoryViewSize.height) / 2
+        
+        // Accessory view aligned to the middle of the messageContainerView
+        var y = (messageContainerView.bounds.height - attributes.accessoryViewSize.height) / 2
+        y += messageContainerView.frame.origin.y
```
SD10: Pretty sure the `cellContentHeight` method will be incorrect due to this new alignment

SD10 (commented):
@nathantannar4 Really appreciate you stepping up and taking this on 🥇 💯 💪 Just a few things

Sources/Views/Cells/MessageContentCell.swift
SD10: Maybe not actually 🤔 Need to review in more depth

nathantannar4:
@SD10 💪

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -97,7 +97,8 @@ open class MessageSizeCalculator: CellSizeCalculator {
         let avatarVerticalPosition = avatarPosition(for: message).vertical
         let accessoryViewHeight = accessoryViewSize(for: message).height
         let accessoryViewVerticalPadding = accessoryViewPadding(for: message).vertical
-        let accessoryViewTotalHeight = accessoryViewHeight + accessoryViewVerticalPadding
+        let accessoryViewYOriginInset = cellTopLabelHeight + messageTopLabelHeight + messageVerticalPadding + messageContainerPadding(for: message).top
+        let accessoryViewTotalHeight = accessoryViewHeight + accessoryViewVerticalPadding + accessoryViewYOriginInset
```
SD10: Care to explain what you're doing here? I'll review later but the logic was already super confusing before this 🤔

nathantannar4: Adding the extra distance that the accessoryView's origin starts at

SD10: `messageVerticalPadding` already includes `messageContainerPadding(for: message).top`. Not sure if I really understand this approach? Shouldn't the logic be the same for resolving the height against the `AvatarView` when it is positioned `.messageCenter`? 

If you think you have a better way of calculating it -- I'm open to suggestions -- but I can't understand what trick you're using so I'm skeptical 🤔

nathantannar4: Idk if this is a better way. I was just trying to come to a solution. I can adjust

Sources/Views/Cells/MessageContentCell.swift
SD10: @nathantannar4 I think your logic for this is:
```swift
var y = messageContainerView.frame.midY - (attributes.accessoryViewSize.height/2)
```
This is a tough area of the codebase I know

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -97,7 +97,8 @@ open class MessageSizeCalculator: CellSizeCalculator {
         let avatarVerticalPosition = avatarPosition(for: message).vertical
         let accessoryViewHeight = accessoryViewSize(for: message).height
         let accessoryViewVerticalPadding = accessoryViewPadding(for: message).vertical
-        let accessoryViewTotalHeight = accessoryViewHeight + accessoryViewVerticalPadding
+        let accessoryViewYOriginInset = cellTopLabelHeight + messageTopLabelHeight + messageContainerPadding(for: message).top
+        let accessoryViewTotalHeight = accessoryViewHeight + accessoryViewVerticalPadding + accessoryViewYOriginInset
```
SD10: This should just be:
```swift
let accessoryViewTotalHeight = accessoryViewHeight + accessoryViewVerticalPadding
```
No need for `accessoryViewYOriginInset`, see the reason why below

SD10: The reason is because the `cellContentHeight` logic was already fine as is 😆. Do you mind cleaning up the git history by squashing these commits? Since we flip flopped back and forth

EDIT: I think I may have to make a slight change actually... jeez

nathantannar4: Reverted

SD10:
Alright, fixing this now 😓

Sources/Views/Cells/MessageContentCell.swift
nathantannar4: @SD10 Made this change

Example/Sources/View Controllers/AdvancedExampleViewController.swift
nathantannar4: Added with a comment to remind devs of this

nathantannar4:
@SD10 I think this is fairly solid in its current state. Layout gets messed up when absurdly large accessory view sizes are set, but that could be listed as a known bug so we can push the 1.1 release which will fix the bottom inset issues. Then we could update to Swift 4.2 for a 1.2 update

SD10:
@nathantannar4 Thanks for taking lead here 👍 The problem here is we shouldn't ship an API where the accessory view has a `UIEdgeInsets` for padding because they should not be able to set any vertical padding -- only left and right. As it is, the current layout logic for the accessory view's Y position doesn't make any sense

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -275,8 +275,7 @@ open class MessageContentCell: MessageCollectionViewCell {
     open func layoutAccessoryView(with attributes: MessagesCollectionViewLayoutAttributes) {
         
         // Accessory view aligned to the middle of the messageContainerView
-        var y = (messageContainerView.bounds.height - attributes.accessoryViewSize.height) / 2
-        y += messageContainerView.frame.origin.y
+        var y = messageContainerView.frame.midY - (attributes.accessoryViewSize.height/2)
```
SD10: This part is perfect

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -275,8 +275,7 @@ open class MessageContentCell: MessageCollectionViewCell {
...
         y -= attributes.accessoryViewPadding.vertical + attributes.accessoryViewPadding.top
```
SD10: This code isn't doing anything except ruining the calculation above. This is what I mean by no vertical padding

Example/Sources/View Controllers/AdvancedExampleViewController.swift
```diff
@@ -97,6 +97,11 @@ final class AdvancedExampleViewController: ChatViewController {
         layout?.setMessageIncomingAvatarSize(CGSize(width: 30, height: 30))
         layout?.setMessageIncomingMessagePadding(UIEdgeInsets(top: -outgoingAvatarOverlap, left: -18, bottom: outgoingAvatarOverlap, right: 18))
         
+        layout?.setMessageIncomingAccessoryViewSize(CGSize(width: 30, height: 30))
+        layout?.setMessageIncomingAccessoryViewPadding(UIEdgeInsets(top: 0, left: 8, bottom: 0, right: 0))
```
SD10: An unfortunate API at this point:
```
incomingAccessoryViewLeftPadding
incomingAccessoryViewRightPadding
outgoingAccessoryViewLeftPadding
outgoingAccessoryViewRightPadding
```

nathantannar4:
Current state of this PR:
1) Under fair sizing of the accessoryView, ex: 30x30 the layout looks as follows
simulator screen shot - iphone xr - 2018-10-11 at 22 20 32

2) Under sizing that breaks layout, ie: developer error
simulator screen shot - iphone xr - 2018-10-11 at 22 20 49

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -95,40 +95,37 @@ open class MessageSizeCalculator: CellSizeCalculator {
...
-        let accessoryViewVerticalPadding = accessoryViewPadding(for: message).vertical
-        let accessoryViewTotalHeight = accessoryViewHeight + accessoryViewVerticalPadding
 
         switch avatarVerticalPosition {
         case .messageCenter:
             let totalLabelHeight: CGFloat = cellTopLabelHeight + messageTopLabelHeight
                 + messageContainerHeight + messageVerticalPadding + messageBottomLabelHeight
             let cellHeight = max(avatarHeight, totalLabelHeight)
-            return max(accessoryViewTotalHeight, cellHeight)
+            return cellHeight
```
SD10: Are we sure this was the correct change? 🤔 Seems like it was correct before

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -231,7 +228,7 @@ open class MessageSizeCalculator: CellSizeCalculator {
         let messagePadding = messageContainerPadding(for: message)
         let accessoryWidth = accessoryViewSize(for: message).width
         let accessoryPadding = accessoryViewPadding(for: message)
-        return messagesLayout.itemWidth - avatarWidth - messagePadding.horizontal - accessoryWidth - accessoryPadding.horizontal
+        return messagesLayout.itemWidth - avatarWidth - messagePadding.horizontal + accessoryWidth + accessoryPadding.horizontal
```
SD10: These should be subtracted, not added?

Sources/Models/HorizontalEdgeInsets.swift
```diff
@@ -0,0 +1,55 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+
+/// A varient of `UIEdgeInsets` that only has horizontal inset properties
+public struct HorizontalEdgeInsets {
```
SD10, line 28: 💯 🥇

Sources/Layout/MessageSizeCalculator.swift
nathantannar4: Reverted, I goofed

Sources/Layout/MessageSizeCalculator.swift
nathantannar4: Ok, I thought this was one of the things you wanted reverted and if it messed up the height it was a developer error, back to how I originally had it :D

PR #874, opened 2018-09-20
Updating "apps using this library"

marcocapano:
Adding my latest work in the list of apps using this library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ Interested in contributing to MessageKit? Click here to join our [Slack](https:/
 Add your app to the list of apps using this library and make a pull request.
 
 - [MediQuo](https://www.mediquo.com)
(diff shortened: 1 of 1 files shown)


marcocapano:
PS: please provide attribution if you use this. It's pretty clear that this is not used by only one app lol

According to Cocoapods this library was integrated in 2,586 apps. The number of published apps has to be at least bigger than 500😊

nathantannar4 (approved):

nathantannar4:
@SD10 Is this ok to be put into the `master` branch?

SD10 (approved):
Thanks @marcocapano 😄 I think CocoaPods stopped reporting stats but I don't care for them 🤷‍♂️  

Yeah I'm fine with this @nathantannar4 👍

PR #879, opened 2018-09-27
MessagesDataSource & MessagesLayoutDelegate methods for MessageType.custom

RealBonus:
Simpler solution for working with custom cells: Delegate methods. No needs to override or implement custom FlowLayout's.

Any other comments?
-------------------
👻

Where has this been tested?
---------------------------
**iOS Version:**  11, 12
Not a major logic refactoring, should work anywhere.

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -180,6 +180,8 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
         return collectionView.messagesDataSource?.numberOfItems(inSection: section, in: collectionView) ?? 0
     }
 
+    /// Note:
+    ///   If you override this method, remember to call MessagesDataSource's customCell(for:at:in:) for MessageKind.custom messages, if necessary
     open func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
 
         guard let messagesCollectionView = collectionView as? MessagesCollectionView else {
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -206,7 +206,7 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
             cell.configure(with: message, at: indexPath, and: messagesCollectionView)
             return cell
         case .custom:
-            fatalError(MessageKitError.customDataUnresolvedCell)
+            return messagesDataSource.customCell(for: message, at: indexPath, in: messagesCollectionView)
         }
     }
 
--- a/Sources/Layout/MessagesCollectionViewFlowLayout.swift
+++ b/Sources/Layout/MessagesCollectionViewFlowLayout.swift
@@ -151,7 +151,7 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
         case .location:
             return locationMessageSizeCalculator
         case .custom:
-            fatalError("Must return a CellSizeCalculator for MessageKind.custom(Any?)")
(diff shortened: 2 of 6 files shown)


Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -80,6 +80,17 @@ public protocol MessagesLayoutDelegate: AnyObject {
...
+    /// Cell size calculator for messages with MessageType.custom.
+    ///
+    /// - Parameters:
+    ///   - message: The custom message
+    ///   - indexPath: The `IndexPath` of the cell.
+    ///   - messagesCollectionView: The `MessagesCollectionView` in which this cell will be displayed.
+    ///
+    /// - Note:
+    ///   The default implementation will throw fatalError(). You must override this method if you are using messages with MessageType.custom.
+    func cellSizeCalculator(for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> CellSizeCalculator
```
SD10: Hey @RealBonus, this is an interesting PR. Thanks for taking the time to submit it 👍 I'm a little concerned that this method is called `cellSizeCalculator` but is only used for the `.custom` case of `MessageKind`. Maybe `customCellSizeCalculator` is a better name? Aside from that, I'll let some other team members review this

RealBonus: Good point, agreed

nathantannar4: I think this does a good job making it easier for people to get started with using custom cells with, as you said, not having to subclass the layout. And if people still want to subclass the layout they can, this doesn't raise any breaking changes.

nathantannar4 (approved):

Sources/Layout/MessagesCollectionViewFlowLayout.swift
```diff
@@ -151,7 +151,7 @@ open class MessagesCollectionViewFlowLayout: UICollectionViewFlowLayout {
         case .location:
             return locationMessageSizeCalculator
         case .custom:
-            fatalError("Must return a CellSizeCalculator for MessageKind.custom(Any?)")
+            return messagesLayoutDelegate.customCellSizeCalculator(for: message, at: indexPath, in: messagesCollectionView)
```
SD10, line 156: I think we need to add a comment to tell the user that if they override
`func cellSizeCalculatorForItem(at indexPath: IndexPath) -> CellSizeCalculator`
they will then be responsible for calling the `customCellSizeCalculator` method if necessary

Sources/Controllers/MessagesViewController.swift
```diff
@@ -206,7 +206,7 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
             cell.configure(with: message, at: indexPath, and: messagesCollectionView)
             return cell
         case .custom:
-            fatalError(MessageKitError.customDataUnresolvedCell)
+            return messagesDataSource.customCell(for: message, at: indexPath, in: messagesCollectionView)
```
SD10, line 211: Same here, I'm worried someone overrides `cellForItem(at:)` and then doesn't handle calling `customCell`. Then they are using the delegate and expecting it to be called

SD10:
@RealBonus We also need to add a CHANGELOG entry for these changes

nathantannar4:
@RealBonus Can you please add a CHANGELOG entry and add the comments Steven requested so we can get this in the next release!

RealBonus:
Sorry for delay guys, out of town out of business.
CHANGELOG and comments: Done!

nathantannar4:
🔥 💯

SD10:

PR #880, opened 2018-09-27
update to swift 4.2

moldovaniosif:
What does this implement/fix? Explain your changes.
---------------------------------------------------
Update project to support swift 4.2.

Does this close any currently open issues?
------------------------------------------
No.

Any relevant logs, error output, etc?
-------------------------------------

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 7, Simulator 7, X;

**iOS Version:** 12.0

**Swift Version:** 4.2

**MessageKit Version:** 1.0.0

PR #883, opened 2018-09-29
Swift 4.2

nathantannar4:
Updates the codebase to Swift 4.2, specifies Swift 4.2 version of `MessageInputBar`, update CircleCi config

Everyone requesting Swift 4.2 😊 

Any other comments?
-------------------
Sorry to those who have submitted Swift 4.2 PRs already. #880 was branched from `master`, not `development` and we have already made changes in `development`

For those who want early access, use the branch `swift-4.2`

This will segue `MessageKit` to its 1.1 release


SD10 (approved):
@nathantannar4 Please add a CHANGELOG entry for the upcoming release - "Added Swift 4.2 support"

Sources/Views/PlayButtonView.swift
```diff
@@ -28,7 +28,7 @@ open class PlayButtonView: UIView {
 
     // MARK: - Properties
 
-    open let triangleView = UIView()
+    public let triangleView = UIView()
```
zhongwuzw, line 31: Emm, how about make it `open` and `var`?

nathantannar4: What would be the use case for allowing someone to set a new value?

zhongwuzw: Maybe users want to do some custom things? Make it `open` have no side-effect, right?

nathantannar4: Before it was open but a let. So technically they can't change it. But if it's changed to var someone can set it to something. But if they set it that won't automatically add the new view to the superview or lay it out.

hyouuu:
This is awesome! Please update MessageInputBar as well :D

hyouuu:
Oh I see the repo is updated, but cocoapod didn't update it - probably because the podspec didn't specify a version. I had to add MessageInputBar directly in my pod file, then it was updated from 0.2.2 to 0.4.0

talanov:
@nathantannar4 PodSpec invalid, MessageKit's InputBar automatically install 0.2.2 version instead of 0.4.

nathantannar4:
@talanov I didn't officially released a new pod version for MessageInputBar bar since this is wasn't a release yet. You will need to specify the version until official release. 

SD10:
@nathantannar4 You can release `MessageInputBar 0.4.0` though? It shouldn't be blocked by a MessageKit release

nathantannar4:
Sure, released just now

PR #890, opened 2018-10-02
Fix UIButton highlight behaviour when contained by MessagesCollectionView

marius-serban:
Because the tap gesture recognizer on `MessagesCollectionView` delays the touch began events, `UIButton` subviews can't properly respond by highlighting.
I didn't see a good reason for `delaysTouchesBegan` to be on so I removed it.

nope

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …

### DIFF
--- a/Sources/Views/MessagesCollectionView.swift
+++ b/Sources/Views/MessagesCollectionView.swift
@@ -71,7 +71,6 @@ open class MessagesCollectionView: UICollectionView {
     
     private func setupGestureRecognizers() {
         let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleTapGesture(_:)))
-        tapGesture.delaysTouchesBegan = true
         addGestureRecognizer(tapGesture)
     }
     


SD10:
@zhongwuzw Pinging you because I believe you added this method. Do you know why we were using `delayTouchesBegan`?

nathantannar4:
@marius-serban Please change destination branch to `development` and add a CHANGELOG entry for upcoming release

zhongwuzw:
@SD10 Sorry for late reply, I'm having National Day holidays, may have no time to review it in depth 😂 , currently, we may have issues using custom tap gesture, it would intercept touch events and block any subviews touch handler, and another issue is like `imageView` of media cell and `messageContainerView`, they are `UIImageView`, `isUserInteractionEnabled` is false, it also block the touch events.

SD10:
@zhongwuzw No worries! Thanks for taking the time to reply ❤️ Enjoy the holidays

marius-serban:
Indeed, `isUserInteractionEnabled` is `false` for `messageContainerView`, but that's something that users can customize when subclassing `MessageContentCell`. The problem with the tap gesture recognizer is that it isn't exposed by `MessagesCollectionView`, maybe it should be? But for now, as a minimal change, I just made it not interfere with other gesture recognizers by delaying touches began.

zhongwuzw:
I think we may should adjust touch logic.
1. `CollectionView` already has built-in tap gesture, so we should move custom tap gesture to cell to prevent issues.
2. Set `Container` views `isUserInteractionEnabled` to `true`.
3. `delaysTouchesBegan` can remove and keep the default value.

nathantannar4:
@marius-serban Would you be interested in refactoring the touch logic?

marius-serban:
I'm sorry @nathantannar4 but I can't commit to doing it, I'm just very limited on time and I don't want to block you folks by saying I'm going to do it and actually doing it 2 months later.

SD10:
@marius-serban Don't ever apologize for not contributing to open source 👍 We work on these things as we have time. @MessageKit/contributors anyone want to look into this?

PR #892, opened 2018-10-04
Add support for audio messages.

moldovaniosif:
Add support for audio messages.

No.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** : iPhone 7, iPhone 4S, Simulator iPhone 7

**iOS Version:**  12.0, 9.3.5

**Swift Version:** 4.2

**MessageKit Version:** 1.0


nathantannar4:
Excited to see someone submit a PR for this feature! 🔥 

Example/Sources/View Controllers/BasicExampleViewController.swift
```diff
@@ -26,6 +26,7 @@ import UIKit
 import MapKit
 import MessageKit
 import MessageInputBar
+import AVFoundation
```
moldovaniosif: there's no need for this import

Example/Sources/View Controllers/ChatViewController.swift
```diff
@@ -25,14 +25,15 @@ SOFTWARE.
 import UIKit
 import MessageKit
 import MessageInputBar
+import AVFoundation
```
moldovaniosif: there's no need for this import

Sources/Controllers/AudioController.swift
```diff
@@ -0,0 +1,172 @@
...
+    /// The `Timer` that update playing progress
+    fileprivate var progressTimer: Timer?
+
+    /// The `AudioMessageCell` that is currently playing sound
+    fileprivate var playingCell: AudioMessageCell?
+
+    /// The `MessageType` that is currently playing sound
+    fileprivate var playingMessage: MessageType?
+
+    fileprivate weak var messageCollectionView: MessagesCollectionView?
```
moldovaniosif: messageCollectionView property should not be optional.

Documentation/FAQs.md
```diff
@@ -1,6 +1,13 @@
 # MessageKit Frequently Asked Questions
 
-**Why doesn't the `MessageInputBar` appear in my controller?**
```
nathantannar4: Thanks for updating the FAQ 💯

Example/ChatExample.xcodeproj/project.pbxproj
```diff
@@ -676,7 +773,7 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 5HU545WEHV;
```
nathantannar4: Please remove development team signing

Example/Sources/Data Generation/SampleData.swift
```diff
@@ -174,15 +183,6 @@ final internal class SampleData {
...
     }
-    
-    func getMessages(count: Int, allowedSenders: [Sender], completion: ([MockMessage]) -> Void) {
-        var messages: [MockMessage] = []
-        for _ in 0..<count {
-            let message = randomMessage(isCustomEnabled: false, allowedSenders: allowedSenders)
-            messages.append(message)
-        }
-        completion(messages)
-    }
```
nathantannar4: Is there a reason you removed this?

Example/Sources/Models/MockMessage.swift
```diff
@@ -53,6 +54,22 @@ private struct ImageMediaItem: MediaItem {
...
+
+    var url: URL
+    var size: CGSize
+    var duration: Float
+
+    init(url: URL) {
+        self.url = url
+        self.size = CGSize(width: 160, height: 35)
+        // compute duration
+        let audioAsset = AVURLAsset.init(url: url)
```
nathantannar4: `let audioAsset = AVURLAsset(url: url)`

Sources/Controllers/AudioController.swift
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            if playingCell != nil { // there is a sound on other cell that is already playing or in pause - prepare to stop playing sound and play the sound for receive message
+                stopAudioInCurrentPlayingCell()
+            }
+            // play sound
+            switch message.kind {
+            case .audio(let item):
+                playingCell = audioCell
+                playingMessage = message
+                audioPlayer = try? AVAudioPlayer.init(contentsOf: item.url)
```
nathantannar4: `audioPlayer = try? AVAudioPlayer(contentsOf: item.url)`

Sources/Controllers/AudioController.swift
```diff
@@ -0,0 +1,172 @@
...
+        audioCell.playButton.isSelected = false // show play button on audio cell since the sound is on pause
+        progressTimer?.invalidate()
+        // call stop audio delegate
+        if let cell = playingCell {
+            cell.delegate?.didPauseAudio(in: cell)
+        }
+    }
+
+    internal func stopAudioInCurrentPlayingCell() {
+        // call stop audio delegate
```
nathantannar4: If the audio player is nil then we don't need to go through the stopping logic
```
// call stop audio delegate
guard let player = audioPlayer else { return }
if let cell = playingCell {
             cell.delegate?.didStopAudio(in: cell)
         }
audioPlayer.stop()
audioPlayerDidFinishPlaying(audioPlayer, successfully: true)
```
Also remove comments, the code is self explanatory

Sources/Controllers/AudioController.swift
```diff
@@ -0,0 +1,172 @@
...
+            cell.delegate?.didStartAudio(in: cell)
+        }
+    }
+
+}
+
+// MARK: - AVAudioPlayerDelegate
+extension AudioController: AVAudioPlayerDelegate {
+
+    internal func audioPlayerDidFinishPlaying(_ player: AVAudioPlayer, successfully flag: Bool) {
```
nathantannar4: `flag: Bool` is not used, also why are we passing in a `player: AVAudioPlayer` parameter when we have access to the `audioPlayer` in the `AudioController`

Sources/Controllers/MessagesViewController.swift
```diff
@@ -71,12 +71,19 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
...
+        audioController.stopAudioInCurrentPlayingCell()
+    }
+
     private var isFirstLayout: Bool = true
     
     internal var isMessagesControllerBeingDismissed: Bool = false
 
     internal var selectedIndexPathForMenu: IndexPath?
 
+    internal var audioController: AudioController!
```
nathantannar4: I think this should be open to allow people to set their own `AudioController`

Sources/Controllers/AudioController.swift
```diff
@@ -0,0 +1,172 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import UIKit
+import AVFoundation
+
+internal class AudioController: NSObject {
```
nathantannar4: Make `AudioController` a protocol to allow people to set their own `AudioController` in the `MessagesViewController`, then make this an open class like `BasicAudioController`

This would allow people to customize  the experience to which is one of the MessageKit goals. We try to make everything open/public unless it will break the framework

Sources/Protocols/AudioCellDelegate.swift
```diff
@@ -0,0 +1,42 @@
...
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+
+/// A protocol used by `AudioMessageCell` to notify about taps on the play/pause button.
+internal protocol AudioCellDelegate: AnyObject {
```
nathantannar4: `public`, see earlier comments

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,171 @@
...
+    internal weak var audioCellDelegate: AudioCellDelegate?
+
+    /// The play button view to display on audio messages.
+    open lazy var playButton: UIButton = {
+        let playButton = UIButton.init(type: .custom)
+        let playImage = AudioMessageCell.getImageWithName("play")
+        let pauseImage = AudioMessageCell.getImageWithName("pause")
+        playButton.setImage(playImage?.withRenderingMode(.alwaysTemplate), for: .normal)
+        playButton.setImage(pauseImage?.withRenderingMode(.alwaysTemplate), for: .selected)
+
```
nathantannar4: remove newline

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,171 @@
...
+/// A subclass of `MessageContentCell` used to display video and audio messages.
+open class AudioMessageCell: MessageContentCell {
+
+    /// The `AudioCellDelegate` delegate
+    internal weak var audioCellDelegate: AudioCellDelegate?
+
+    /// The play button view to display on audio messages.
+    open lazy var playButton: UIButton = {
+        let playButton = UIButton.init(type: .custom)
+        let playImage = AudioMessageCell.getImageWithName("play")
```
nathantannar4: Should use an enum here rather than a raw string

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,171 @@
...
+        durationLabel.font = UIFont.systemFont(ofSize: 14)
+        durationLabel.text = "0:00"
+
+        return durationLabel
+    }()
+
+    open lazy var progressView: UIProgressView = {
+        let progressView = UIProgressView.init(progressViewStyle: .default)
+        progressView.progress = 0.0
+
```
nathantannar4: remove newline

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,171 @@
...
+        return playButton
+    }()
+
+    /// The time duration lable to display on audio messages.
+    open lazy var durationLabel: UILabel = {
+        let durationLabel = UILabel.init(frame: CGRect.zero)
+        durationLabel.textAlignment = .right
+        durationLabel.font = UIFont.systemFont(ofSize: 14)
+        durationLabel.text = "0:00"
+
```
nathantannar4: remove newline

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,171 @@
...
+    open lazy var progressView: UIProgressView = {
+        let progressView = UIProgressView.init(progressViewStyle: .default)
+        progressView.progress = 0.0
+
+        return progressView
+    }()
+
+    // MARK: - Internal Methods
+    internal class func durationString(from duration: Float) -> String {
```
nathantannar4: Not sure a `class` func is the right way to approach this, is there a better way? At the very least this should be `open`

nathantannar4 (changes requested):
A lot of changes to look though here. General feedback is that I would like more classes/methods to be `open` so they are overridable/subclassable. That way users can customize it. The design of the `AudioMessageCell` is great! And seems to work without issue.

Sources/Controllers/AudioController.swift
moldovaniosif: The audioPlayerDidFinishPlaying(::) is a delegate method that notify when AVAudioPlayer finished play. I can not modify the structure of this method since this one is from the OS system. I think is better to implement a new method without parameters an call it from audioPlayerDidFinishPlaying(::) and stopAudioInCurrentPlayingCell() methods.

Example/Sources/Data Generation/SampleData.swift
moldovaniosif: This method is not used any more, there is other method " func getMessages(count: Int, completion: ([MockMessage]) -> Void) " that does the same thing.

Sources/Controllers/AudioController.swift
moldovaniosif: The messageCollectionView can be set a constant using let.

Documentation/FAQs.md
moldovaniosif: I souldn't be credit for updating FAQ, I had no ideea why FAQ had been updated in my PR.

Sources/Views/Cells/AudioMessageCell.swift
moldovaniosif: @nathantannar4 you are right, this class method is not the best approach. I had removed it from AudioMessageCell and instead I created a new delegate methods on MessageDisplayDelegate. In this case user can proide it's own implementation for formatting the audio sound duration or can use the default implementation.

Sources/Controllers/MessagesViewController.swift
moldovaniosif: You are right, I wasn't thinking about how other users can customize the AudioController.

Sources/Controllers/AudioController.swift
moldovaniosif: @nathantannar4 greate ideea.

Sources/Protocols/AudioCellDelegate.swift
moldovaniosif: I've remove the `AudioCellDelegate` protocol and move the didTapPlayButton() delegate method in the `MessageCellDelegate` protocol. In this way the user can take any action when audio play button was pressed.

moldovaniosif:
@nathantannar4 thank you for yhe details review. I had apply the changes.

Example/Sources/View Controllers/ChatViewController.swift
```diff
@@ -207,6 +207,20 @@ extension ChatViewController: MessageCellDelegate {
...
     }
 
+    func didTapPlayButton(in cell: AudioMessageCell) {
+        guard let indexPath = messagesCollectionView.indexPath(for: cell),
+            let message = messagesCollectionView.messagesDataSource?.messageForItem(at: indexPath, in: messagesCollectionView) else {
+                print("Failed to identify message when audio cell receive tap gesture")
+                return
+        }
+        // check if should play sound or pause
+        if cell.playButton.isSelected == true { // sound is playing - prepare to pause sound
```
nathantannar4: If this is how playing sound is determined its not a very intuitive way. I think something like `isPlaying: Bool` should be accessible in the AudioController as its what controls the sound, not the cell

Sources/Controllers/BasicAudioController.swift
```diff
@@ -25,55 +25,79 @@
...
     /// The `AVAudioPlayer` that is playing the sound
-    fileprivate var audioPlayer: AVAudioPlayer?
-
-    /// The `Timer` that update playing progress
-    fileprivate var progressTimer: Timer?
+    open var audioPlayer: AVAudioPlayer?
 
     /// The `AudioMessageCell` that is currently playing sound
-    fileprivate var playingCell: AudioMessageCell?
+    open var playingCell: AudioMessageCell?
```
nathantannar4: I think this should be weak

Sources/Controllers/BasicAudioController.swift
```diff
@@ -25,55 +25,79 @@
...
 
     /// The `MessageType` that is currently playing sound
-    fileprivate var playingMessage: MessageType?
+    open var playingMessage: MessageType?
```
nathantannar4: I think this should be weak

Sources/Controllers/BasicAudioController.swift
```diff
@@ -25,55 +25,79 @@
...
 
-    fileprivate weak var messageCollectionView: MessagesCollectionView?
+    // The `MessagesCollectionView` where the playing cell exist
+    public let messageCollectionView: MessagesCollectionView
```
nathantannar4: This should defiantly be weak

`open weak var messageCollectionView: MessagesCollectionView!`

Sources/Controllers/BasicAudioController.swift
```diff
@@ -25,55 +25,79 @@
...
+    /// Used to start or resume the audio sound
+    ///
+    /// - Parameters:
+    ///   - message: The `MessageType` that contain the audio item to be played.
+    ///   - audioCell: The `AudioMessageCell` that needs to be updated while audio is playing.
+    open func playSound(for message: MessageType, in audioCell: AudioMessageCell) {
         if playingMessage?.messageId == message.messageId { // resume curently pause audio
             self.resumeSound()
         } else {
             if playingCell != nil { // there is a sound on other cell that is already playing or in pause - prepare to stop playing sound and play the sound for receive message
```
nathantannar4: I don't think this is a good way of determining if a sound is playing. As earlier mentioned the AudioController should have an accessor for if its currently playing

Sources/Controllers/BasicAudioController.swift
```diff
@@ -25,55 +25,79 @@
...
             case .audio(let item):
                 playingCell = audioCell
                 playingMessage = message
-                audioPlayer = try? AVAudioPlayer.init(contentsOf: item.url)
-                audioPlayer?.delegate = self
+                audioPlayer = self.createAudioPlayer(url: item.url)
                 audioPlayer?.prepareToPlay()
+                audioPlayer?.delegate = self
                 audioPlayer?.play()
                 audioCell.playButton.isSelected = true  // show pause button on audio cell since the sound is currently playing
```
nathantannar4: As before mentioned, not a good way of managing `isPlaying`

Sources/Controllers/BasicAudioController.swift
```diff
@@ -95,31 +124,46 @@ internal class AudioController: NSObject {
...
             if currentMessage != nil && currentMessage?.messageId == playingMessage?.messageId {
-                // messages are the same updte cell content
+                // messages are the same update cell content
                 cell.progressView.progress = (player.duration == 0) ? 0 : Float(player.currentTime/player.duration)
-                cell.durationLabel.text = AudioMessageCell.durationString(from: Float(player.currentTime))
+                guard let displayDelegate = messageCollectionView.messagesDisplayDelegate else {
+                    fatalError(MessageKitError.nilMessagesDisplayDelegate)
+                }
+                cell.durationLabel.text = displayDelegate.formatDuration(Float(player.duration), for: cell, in: messageCollectionView)
             }
```
nathantannar4: I think that if the currentMessage has changed and is no longer the stored message audio playing and the timer should terminate

Sources/Controllers/BasicAudioController.swift
```diff
@@ -143,30 +187,23 @@ internal class AudioController: NSObject {
         }
     }
 
+    private func createAudioPlayer(url: URL) -> AVAudioPlayer? {
```
nathantannar4: This can be removed and where needed use:
```
guard let audioPlayer = audioPlayer = try? AVAudioPlayer(contentsOf: url) else { return }
```
Since if we can't create the player we just don't play anything rather than returning nil and going about as if we did create a player

Sources/Controllers/BasicAudioController.swift
```diff
@@ -95,31 +124,46 @@ internal class AudioController: NSObject {
         }
     }
 
-    internal func stopAudioInCurrentPlayingCell() {
-        // call stop audio delegate
+    /// Stops any ongoing audio playing if exists
+    open func stopAnyOngoingPLaying() {
```
nathantannar4: typo

Sources/Controllers/BasicAudioController.swift
```diff
@@ -143,30 +187,23 @@ internal class AudioController: NSObject {
...
-        progressTimer?.invalidate()
-        progressTimer = nil
-        audioPlayer = nil
-        playingMessage = nil
-        playingCell = nil
+extension BasicAudioController: AVAudioPlayerDelegate {
+
+    open func audioPlayerDidFinishPlaying(_ player: AVAudioPlayer, successfully flag: Bool) {
+        self.stopAnyOngoingPLaying()
     }
```
nathantannar4: I would add the other delegate methods for the case where there is some audio playing corruption

Sources/Controllers/BasicAudioController.swift
```diff
@@ -143,30 +187,23 @@ internal class AudioController: NSObject {
...
-        if let item = audioItem {
-            playingCell?.durationLabel.text = AudioMessageCell.durationString(from: item.duration)
-        }
-        // reset properties since audio finished
-        progressTimer?.invalidate()
-        progressTimer = nil
-        audioPlayer = nil
-        playingMessage = nil
-        playingCell = nil
+extension BasicAudioController: AVAudioPlayerDelegate {
```
nathantannar4: Dont make this an extension, it means the methods cannot be overridden even though they are open

Sources/Controllers/MessagesViewController.swift
```diff
@@ -60,6 +60,9 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
         return false
     }
 
+    /// The `BasicAudioController` controll the AVAudioPlayer state (play, pause, stop) and udpate audio cell UI accordingly.
+    open var audioController: BasicAudioController!
```
nathantannar4: Make this lazy, as those who don't use audio messages will never need it to be created
`open lazy var audioController = BasicAudioController(messageCollectionView: messageCollectionView)`

Sources/Protocols/MessagesDisplayDelegate.swift
```diff
@@ -184,6 +173,33 @@ public protocol MessagesDisplayDelegate: AnyObject {
...
+    ///   - duration: The audio sound duration is seconds.
+    ///   - audioCell: The `AudioMessageCell` that ask for formated duration.
+    ///   - messagesCollectionView: The `MessagesCollectionView` in which this cell will be displayed.
+    ///
+    /// - Note:
+    ///   The default value is computed like fallow:
+    ///     1. return the time as 0:ss if duration is up to 59 seconds                         (e.g. 0:03     means 0 minutes and 3 seconds)
+    ///     2. return the time as m:ss if duration is greater than 59 and lower than 3600      (e.g. 12:23    means 12 mintues and 23 seconds)
+    ///     3. return the time as h:mm:ss for anything longer that 3600 seconds                (e.g. 1:19:08  means 1 hour 19 minutes and 8 seconds)
+    func formatDuration(_ duration: Float, for audioCell: AudioMessageCell, in messageCollectionView: MessagesCollectionView) -> String
```
nathantannar4: `audioProgressTextFormat` is a bit more discoverable/descriptive

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -160,12 +138,16 @@ open class AudioMessageCell: MessageContentCell {
 
     // MARK: - Helpers
 
-    private class func getImageWithName(_ imageName: String) -> UIImage? {
+    private class func getImageWithName(_ imageName: ImageName) -> UIImage? {
```
nathantannar4: This should be open so people could subclass the cell and get the image we have

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -160,12 +138,16 @@ open class AudioMessageCell: MessageContentCell {
...
         let assetBundle = Bundle.messageKitAssetBundle()
-        let imagePath = assetBundle.path(forResource: imageName, ofType: "png", inDirectory: "Images")
+        let imagePath = assetBundle.path(forResource: imageName.rawValue, ofType: "png", inDirectory: "Images")
         let image = UIImage(contentsOfFile: imagePath ?? "")
-
         return image
     }
 
+    private enum ImageName: String {
```
nathantannar4: Public and place at the top of the class

nathantannar4 (changes requested):
Thanks for making the previous changes! I think some improvements could still be made

Sources/Controllers/BasicAudioController.swift
moldovaniosif: You are right.

Sources/Controllers/BasicAudioController.swift
moldovaniosif: The `playingMessage ` can not be made weak because `MessageType` is not a class bound object.

moldovaniosif:
Second review changes had been applied.

moldovaniosif:
Hi @nathantannar4 is everything ok with this latest changes?
Is there anything that could be improved?

Sources/Controllers/BasicAudioController.swift
```diff
@@ -25,21 +25,37 @@
...
+public enum PlayerState {
+
+    /// The audio controller is currently playing a sound
+    case playing
+
+    /// The audio controller is currently in pause state
+    case pause
+
+    /// The audio controller is not playing any sound and audioPlayer is nil
+    case none
```
nathantannar4: Perhaps `stopped` is a better name for this, a `none` state doesn't make sense grammatically

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,152 @@
...
+    open lazy var durationLabel: UILabel = {
+        let durationLabel = UILabel.init(frame: CGRect.zero)
+        durationLabel.textAlignment = .right
+        durationLabel.font = UIFont.systemFont(ofSize: 14)
+        durationLabel.text = "0:00"
+        return durationLabel
+    }()
+
+    open lazy var progressView: UIProgressView = {
+        let progressView = UIProgressView.init(progressViewStyle: .default)
```
nathantannar4: Please cleanup these `.init` as its not needed

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -0,0 +1,152 @@
...
+        messageContainerView.addSubview(durationLabel)
+        messageContainerView.addSubview(progressView)
+        setupConstraints()
+    }
+
+    open override func prepareForReuse() {
+        super.prepareForReuse()
+        progressView.progress = 0
+        playButton.isSelected = false
+        self.durationLabel.text = "0:00"
```
nathantannar4: Please remove `self` whenever possible

Sources/Controllers/BasicAudioController.swift
```diff
@@ -0,0 +1,223 @@
...
+            audioPlayer?.prepareToPlay()
+            audioPlayer?.delegate = self
+            audioPlayer?.play()
+            state = .playing
+            audioCell.playButton.isSelected = true  // show pause button on audio cell since the sound is currently playing
+            startProgressTimer()
+            // call start audio delegate
+            audioCell.delegate?.didStartAudio(in: audioCell)
+        default:
+            print("Error - can not play sound for a message that is not of kind .audio")
```
nathantannar4: I believe we have a file for MessageKit errors, can you please move this there and follow the convention we use for printing errors?

Sources/Controllers/BasicAudioController.swift
```diff
@@ -0,0 +1,223 @@
...
+        progressTimer?.invalidate()
+        progressTimer = nil
+        audioPlayer = nil
+        playingMessage = nil
+        playingCell = nil
+    }
+
+    /// Resume a currently pause audio sound
+    open func resumeSound() {
+        guard let player = audioPlayer else {
```
nathantannar4: We should also be guarding the `cell` here

nathantannar4 (changes requested):
Getting to a much better state! In general, can you please cleanup your commenting so things are more readable. Please note that as this is a larger change we won't be able to include it in v1.1. It would be great if others could also review this, I still haven't had time to test how customizable this is. 

It would also be great if there were some tests for the AudioController 😊 

Thanks for your continued contribution

moldovaniosif:
I've added some test cases for AudioController. If i've miss some cases please let me know.
I had no ideea why circleUI tests are falling, on my mashine are allways pass. It may be because I've added a resource to test target?

nathantannar4:
CircleCI probably complaining about xcode/Swift version. Did you migrate this PR to Swift 4.2 yet? You might need to rebase it on the latest `development` commit 

Sources/Protocols/AudioControllerDelegate.swift
```diff
@@ -0,0 +1,45 @@
...
+    ///     2. progresssView progress;
+    ///     3. durationLabel text;
+    ///
+    /// - Parameters:
+    ///   - cell: The `AudioMessageCell` that needs to be configure.
+    ///   - message: The `MessageType` that configures the cell.
+    ///
+    /// - Note:
+    ///   This protocol method is called by MessageKit every time an audio cell needs to be configure
+    func configureCell(_ cell: AudioMessageCell, message: MessageType)
```
SD10: Do you mind naming this `configureAudioCell`? So that it's clearer as to what kind of cell

SD10:
@moldovaniosif This is looking pretty cool ❤️ Thanks for all the effort 🥇 And @nathantannar4 thanks for not letting this slip through the cracks and doing the review. I reviewed it briefly now but plan to take a deeper look

moldovaniosif:
Current PR is using swift 4.2 and yesterday I rebased it to the lastest development commit. CircleCI is complaing to the test cases I've just added in the AudioControllerTests, some true asserts are falling. I don't know how CircleCI is working or how can debug it, if I runt tests from Xcode everything is working fine.

talanov (approved):

Tests/ControllersTest/AudioControllerTests.swift
```diff
@@ -0,0 +1,155 @@
...
+
+    func testPauseSound() {
+        let audioMessage = getAudioMessage()
+        let audioCell = AudioMessageCell(frame: CGRect.zero)
+        audioController.playSound(for: audioMessage, in: audioCell)
+        audioController.pauseSound(for: audioMessage, in: audioCell)
+        XCTAssertNotNil(audioController.audioPlayer)
+        XCTAssertNotNil(audioController.playingCell)
+        XCTAssertNotNil(audioController.playingMessage)
+        XCTAssertEqual(audioCell.durationLabel.text, "0:00")
```
talanov:
```suggestion
```

talanov (changes requested):
Tests are failing.

audioPlayer isn't nil, yet, isPlaying returns false.

Tests/ControllersTest/AudioControllerTests.swift
moldovaniosif: But why when I run the test from Xcode isPlaying returns true, and from CirecleCi interface returns false? I don't understant what is wrong or what is the diference between Xocde and CircleCI

talanov:
Yep, all tests are passed @ Xcode.

talanov (approved):

nathantannar4:
@moldovaniosif I'd like to get this in the 2.1 release 😊 There is some merge conflicts and we still need a CHANGELOG entry for this. You can place in under upcoming release as a breaking change

moldovaniosif:
I had resolved the merge conflicts, update the change log and remove the aserts that breake the Circle CI.

Sources/Protocols/AudioControllerDelegate.swift
```diff
@@ -0,0 +1,45 @@
...
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ SOFTWARE.
+ */
+
+import Foundation
+
+/// A standard protocol representing an audio controller.
+/// Use this protocol to create your own audio controller object to be used by MessageKit.
+/// MessageKit provide a default implmentation for AudioControllerDelegate in BasicAudioController
+public protocol AudioControllerDelegate: AnyObject {
```
SD10: Does this need to be a separate delegate or can it be part of `MessagesDisplayDelegate`?

SD10 (commented):
@moldovaniosif Thanks for your hard work on this. I don't want it to get swept under the rug. Taking a bit of time to get around to it. Originally, I didn't want any audio player to be part of the library (just the cell UI). Can you make a case for why it needs to be? Thanks again 👍

Sources/Protocols/AudioControllerDelegate.swift
moldovaniosif: The `AudioControllerDelegate` contains the mandatory methods which an AudioController must implement, so any one hwo want a custom implmentation should conform with this protocol. That's way i make it a separate protocol.
I was thinking because `BasicAudioController` updates the `AudioMessageCell` UI it should be part of MessageKit. But it can be moved outside the framework also. If you have any ideea of how the arhitecture for audio can be improved please let me know.

moldovaniosif:
Hi, is there any thing else to change for this pull request? Should I move the AudioController outside the MessageKit?

nathantannar4:
@moldovaniosif Im ok with having a basic Audio controller, but I think there still needs to be some more work on making a more protocol oriented design for it if we keep it, as your design only works for local audio clips, not an audio stream from a server URL. The UI is defiantly on point, perhaps the best way to go about this is if you take out the AudioController parts of this PR and just leave the UI? Then we can make a separate PR for the AudioController

Example/Sources/View Controllers/ChatViewController.swift
```diff
@@ -64,6 +64,8 @@ class ChatViewController: MessagesViewController, MessagesDataSource {
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         MockSocket.shared.disconnect()
+
```
iceman201: remove empty line

Sources/Controllers/MessagesViewController.swift
```diff
@@ -208,6 +211,11 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
             let cell = messagesCollectionView.dequeueReusableCell(LocationMessageCell.self, for: indexPath)
             cell.configure(with: message, at: indexPath, and: messagesCollectionView)
             return cell
+        case .audio:
+            let cell = messagesCollectionView.dequeueReusableCell(AudioMessageCell.self, for: indexPath)
+            cell.configure(with: message, at: indexPath, and: messagesCollectionView)
+            audioController.configureAudioCell(cell, message: message) // this is needed especily when the cell is reconfigure while is playing sound
```
SD10: @moldovaniosif Stopping by again because I gave this more thought tonight. The only thing that concerns me here is `BasicAudioController` from a maintenance perspective. The reason this feature has had a lower priority to me is because I will never use it personally.

1. Is there any way to add Audio cells to MessageKit without the burden of the player?
2. If there's no way around this, can you be the code owner for this feature of MessageKit?

moldovaniosif: 1. I think it should be possible to move the `AudioController` outside the framework.
2. Yes, I can take the responsibility for this feature.

When I will find some time I will try to pull out the `AudioController` as @nathantannar4 mention letting inside the framework only the audio cell UI. Then I can make a new pull request for the `AudioController`. If you have some thoughts about how the `AudioController` should behave or how should be structured I am opened to any suggestions.

moldovaniosif:
I've moved the `BasicAudioController` from MessageKit framework in to the example app. Currently the framework only implement the audio cell UI while the app example treats playing the audio sound.

alexBotteri:
@zhongwuzw do you have an estimate date for this PR review? TY!

zhongwuzw:
@alexBotteri Emm, sorry, I may don't have much time to review this. cc @SD10 @nathantannar4 Hi guys, any thoughts about this `PR`?

bmt2018:
Any idea how far out this is from being merged into the project?  I have a project that I'm working on which requires audio message support and this looks like a perfect fit!

nathantannar4:
@zhongwuzw I've reviewed this quite heavily. Now that it's just the UI elements which are being added it's good. 

bmt2018:
@SD10 @nathantannar4 @zhongwuzw any idea how far away an official release with this feature merged in might be?  I just started using MK and was trying to understand how to create a custom cell to support audio messages when I found this PR.  This is exactly what I need to implement but I'm not sure how I can leverage it until it's been merged into a release?

nathantannar4:
@bmt2018 you can specify @moldovaniosif branch to use this

bmt2018:
> @bmt2018 you can specify @moldovaniosif branch to use this

@nathantannar4 thanks for the suggestion!  I'm sorry for the dumb question but I'm somewhat new to IOS development and still learning.  The project I'm working on uses Cocoapods to manage dependencies so in order to use the branch from @moldovaniosif, I would need to edit my Podfile to specify that, right?  Currently, I just use this:

**pod 'MessageKit'**

which I thought would need to be changed to something like:

**pod 'MessageKit', :git => 'https://github.com/TAGonSoft/MessageKit/MessageKit.git', :branch => 'development'**

but that doesn't seem to work.  Am I going about this completely wrong?

PR #893, opened 2018-10-04
[Accessoryview] fix

JulienKode:
With an Accessoryview of 100 x 300

img_5915

#834

### DIFF
--- a/Sources/Views/Cells/MessageContentCell.swift
+++ b/Sources/Views/Cells/MessageContentCell.swift
@@ -225,7 +225,12 @@ open class MessageContentCell: MessageCollectionViewCell {
                 fallthrough
             }
         default:
-            origin.y = attributes.cellTopLabelSize.height + attributes.messageTopLabelSize.height + attributes.messageContainerPadding.top
+            if attributes.accessoryViewSize.height > attributes.messageContainerSize.height {
+                let messageHeight = attributes.messageContainerSize.height + attributes.messageContainerPadding.vertical
+                origin.y = (attributes.size.height / 2) - (messageHeight / 2)
+            } else {
+                origin.y = attributes.cellTopLabelSize.height + attributes.messageTopLabelSize.height + attributes.messageContainerPadding.top
+            }
         }
 
         switch attributes.avatarPosition.horizontal {
@@ -276,7 +281,6 @@ open class MessageContentCell: MessageCollectionViewCell {
         
         // Accessory view aligned to the middle of the messageContainerView
         var y = messageContainerView.frame.midY - (attributes.accessoryViewSize.height/2)
-        y -= attributes.accessoryViewPadding.vertical + attributes.accessoryViewPadding.top
 
         var origin = CGPoint(x: 0, y: y)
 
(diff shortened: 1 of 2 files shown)


nathantannar4:
@JulienKode Nice! So if the width is set to large your just shrinking it? I think that makes sense so that it fits the overall message cell padding. @SD10 what do you think?

JulienKode:
>@JulienKode Nice! So if the width is set to large your just shrinking it? I think that makes sense so that it fits the overall message cell padding. @SD10 what do you think? ...

@nathantannar4 it’s supposed to do that as we take care of the size of the accessoryView `return messagesLayout.itemWidth - avatarWidth - messagePadding.horizontal - accessoryWidth - accessoryPadding.horizontal`

JulienKode:
We can adjust some part if you want 

nathantannar4 (approved):

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -225,7 +225,12 @@ open class MessageContentCell: MessageCollectionViewCell {
                 fallthrough
             }
         default:
-            origin.y = attributes.cellTopLabelSize.height + attributes.messageTopLabelSize.height + attributes.messageContainerPadding.top
+            if attributes.accessoryViewSize.height > attributes.messageContainerSize.height {
```
SD10, line 228: Can you explain this to me a little bit more? I don't know why it's necessary

JulienKode: We are doing that here because usually the origin.y of the view is just under the others items 

But If the accessoryView is bigger than the messageContentCell, 
It will cause the accessory view to render outisde the cell 

As the accessory view is at the center of the messageContentCell if the accessoryView is bigger, he will go outside of the cell

So if the accessoryView is bigger, we render the messageContentCell, at the middle of the cell then the accessoryView fit perfectly the cell

JulienKode: GitHawk Upload by JulienKode here is what happened if we do not manage this case

SD10: Hey @JulienKode, thanks for explaining this to me 👍 ~I think the real issue is in the height calculation for the cell? I'm sure if this PR solves the issue the correct way but I totally understand what you're saying is the problem~

SD10: There may actually not be a great way to solve this as I think more about it

JulienKode: I agree that’s not a good way to handle this new case, what did you suggest ? 

May be add an enum somewhere

nathantannar4: @SD10 @JulienKode I think the current state is good, it doesnt break the cell layout as all subviews are rendered within the bounds. I think if we want to develop the reliability for rendering options in this case it could be tabled for a future release.

I think we really need to release 1.1, particularly because of the content inset bug people keep facing and the Swift 4.2 update

nathantannar4:
@SD10 Do you think anything still needs to be changed here? This change along with my other changes will ensure the accessory size is centered to the message content view and that large accessory views don't mess up the layout. 

nathantannar4:
@JulienKode This does a good job aligning the accessoryView to the center, but we still need to make some changes. Im gonna merge this to my branch and then work on it.

TODO:
1) Remove vertical padding
2) Always layout the accessoryView to the size specified, even if its bigger than the cell
> If that occurs it is a developer error and we should not auto resize it

JulienKode:
@nathantannar4  

Thanks, I'll take a look at this

PR #897, opened 2018-10-08
Fix intrinsic content size for multi-line text inside MessageLabel

marius-serban:
The superclass implementation of `intrinsicContentSize` doesn't take `textInsets` into consideration when breaking lines . The workaround is to manually implement the size calculation and take `preferredMaxLayoutWidth` into consideration as the line breaking width.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …

### DIFF
--- a/Sources/Views/MessageLabel.swift
+++ b/Sources/Views/MessageLabel.swift
@@ -114,7 +114,9 @@ open class MessageLabel: UILabel {
     }
 
     open override var intrinsicContentSize: CGSize {
-        var size = super.intrinsicContentSize
+        let maxWidth = preferredMaxLayoutWidth - textInsets.horizontal
+        let constraintBox = CGSize(width: maxWidth, height: .greatestFiniteMagnitude)
+        var size = textStorage.boundingRect(with: constraintBox, options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil).integral.size
         size.width += textInsets.horizontal
         size.height += textInsets.vertical
         return size


talanov (approved):

Sources/Views/MessageLabel.swift
```diff
@@ -114,7 +114,9 @@ open class MessageLabel: UILabel {
     }
 
     open override var intrinsicContentSize: CGSize {
-        var size = super.intrinsicContentSize
+        let maxWidth = preferredMaxLayoutWidth - textInsets.horizontal
+        let constraintBox = CGSize(width: maxWidth, height: .greatestFiniteMagnitude)
+        var size = textStorage.boundingRect(with: constraintBox, options: [.usesLineFragmentOrigin, .usesFontLeading], context: nil).integral.size
```
SD10, line 119: @marius-serban You're finding these issues because you're using autolayout/xibs with MessageKit? 

I can see how this is necessary to solve this issue 👍 This is a really heavy operation though, so I'd like to see what impact this has on performance.

marius-serban: Indeed, I expect `intrinsicContentSize` was calling it too but maybe only in a multi-line width constrained scenario and probably caching the result. Some optimisations could definitely be done here, this is a bare-bones not-that-efficient but correct implementation.

The reason I run into this issues is because I'm using a custom cell, non-xib but laid out with autolayout. The content is dynamic, each label could go multi-line and I didn't want to worry about computing the size manually at this point, feels like an early optimisation for my use case.

screen shot 2018-09-27 at 14 00 55

SD10: Still haven't had time to investigate this but I wanted to tag #824. These two methods should be consistent if it gets merged

nathantannar4:
@marius-serban Thank you for raising this issue. As this implementation would really add a cost to the layout of each MessageLabel I don't think this is the right approach. We want to keep the performance high and that means not relying on auto layout.

I believe if a user wants to implement a cell like you have shown they should implement a custom cell

PR #898, opened 2018-10-08
allow dynamic avatar size per message size calculator

marius-serban:
instead of having the same avatar size per size calculator, allow a subclass to override it so it can provide multiple avatar sizes depending on the message

…

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …

### DIFF
--- a/Sources/Layout/MessageSizeCalculator.swift
+++ b/Sources/Layout/MessageSizeCalculator.swift
@@ -120,7 +120,7 @@ open class MessageSizeCalculator: CellSizeCalculator {
 
     // MARK: - Avatar
 
-    public func avatarPosition(for message: MessageType) -> AvatarPosition {
+    open func avatarPosition(for message: MessageType) -> AvatarPosition {
         let dataSource = messagesLayout.messagesDataSource
         let isFromCurrentSender = dataSource.isFromCurrentSender(message: message)
         var position = isFromCurrentSender ? outgoingAvatarPosition : incomingAvatarPosition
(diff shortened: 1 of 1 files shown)


talanov:
How exactly did you implement your avatarView?

I had no problems with it, when used it in xib, size automatically changed.

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -134,7 +134,7 @@ open class MessageSizeCalculator: CellSizeCalculator {
         return position
     }
 
-    public func avatarSize(for message: MessageType) -> CGSize {
+    open func avatarSize(for message: MessageType) -> CGSize {
```
SD10, line 137: We originally made these non-overrideable because allowing dynamic layouts can lead to performance and cache invalidation issues. I think all of these `public` properties can now become `open` tbh. It should be proceed at your own risk at this point

marius-serban:
To be honest, in my case, I just wanted to hide the avatar view in certain scenarios.

Sources/Layout/MessageSizeCalculator.swift
marius-serban: 👍I like the idea of having a couple of good optimised defaults for the basic types of cells, and being able to deeply customise others. Thanks for explaining!

SD10:
@marius-serban For hiding the `AvatarView` you can also use `MessagesDisplayDelegate`:
```swift
func configureAvatarView(_ avatarView: AvatarView, for message: MessageType, at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView)
```

This is a good way to hide the avatar but still retain its spacing when you want to show only one avatar on a stack of messages

marius-serban:
Yup, saw that, but I wanted to get rid of the spacing as well.

Sources/Layout/MessageSizeCalculator.swift
SD10: I still think all functions above `// MARK: - Helpers` should be made from `public` -> `open`. If you want to make that change, I'll accept the PR

marius-serban: sure thing!

SD10: cc @zhongwuzw @nathantannar4, I just wanted to ping you on this one. 

We've essentially come full circle. Originally, we made this public to prevent the user from creating a dynamic layout because it broke our internal cache and had perf implications. But after @zhongwuzw's awesome optimizations 🤗 we ditched the cache (and I don't intend to add it back because the logic for cache invalidation is way too client specific)

Thus, I see no reason for preventing users from doing this any longer

SD10 (approved):

Sources/Layout/MessageSizeCalculator.swift
zhongwuzw: Emm, if we decide that would not do any cache, maybe we can move to `open`.

nathantannar4:
@marius-serban which spacing? Have you looked at the new Advanced example i released? I got rid of some spacing there 

Sources/Layout/MessageSizeCalculator.swift
SD10: Yeah @zhongwuzw, we're not going to add a cache. The client should be responsible for this. It's too client specific. How should we invalidate the cache if a section is reloaded? Other sections could depend on changes in that section (ie. stacking avatars)

SD10:
@nathantannar4 He meant he wanted to keep the size of the avatar view in the layout, just hide it

PR #909, opened 2018-10-13
Make Sender a protocol, renamed SenderType

### LINKED ISSUE #628: Change Sender from struct to protocol
I'm thinking that the `Sender` type is better suited as a protocol. It's common that people want to make an existing user object such as `FIRUser` or `PFUser` conform to `Sender`.

This is out of scope for 1.0 but good to keep track of.

nathantannar4:
This change would allow models such as a `User` to conform to `SenderType` rather than creating a `Sender` struct for each model. This moves to a more protocol oriented design similar to how `MessageType` works.

Closes #628


zhongwuzw:
Good to make it protocol-oriented.

SD10:
@nathantannar4 Please try to put something in the body of your PRs 😂 Can you please add a CHANGELOG entry?

nathantannar4:
Just threw this together, wasn't sure if moving to a protocol was still wanted

Sources/Models/Sender.swift
```diff
@@ -25,7 +25,8 @@
 import Foundation
 
 /// An object that groups the metadata of a messages sender.
-public struct Sender {
+@available(*, deprecated: 2.0, message: "`Sender` has been replaced with the `SenderType` protocol in 2.0")
```
SD10: This needs to become 3.0. Let's add a CHANGELOG entry please 👍 

### Changed
- **Breaking Change** Deprecated the `Sender` struct in favor of the `SenderType` protocol.

JulienKode:
@nathantannar4  really nice to make it as a protocol

SD10:
@nathantannar4 Fix this up and add that CHANGELOG entry when you feel up to it. I think this is a good 3.0 candidate

Sources/Protocols/MessagesDataSource.swift
```diff
@@ -97,7 +97,7 @@ public protocol MessagesDataSource: AnyObject {
 public extension MessagesDataSource {
 
     func isFromCurrentSender(message: MessageType) -> Bool {
-        return message.sender == currentSender()
+        return message.sender.isEqualTo(currentSender())
```
SD10: Can we just do `message.sender.is == currentSender().id`?

SD10:
@nathantannar4 What's the status on this one? Can we freshen this up, add a changelog entry, so we can review again and merge?

nathantannar4:
@SD10 yes I'll do that

jjatie:
To follow Swift convention (I believe introduced in Swift 3, might have been Swift 4), `SenderType` should be named `SenderProtocol`. Refer to `IteratorProtocol` or `StringProtocol`

PR #912, opened 2018-10-15
[MessageLabel] Change privacy of properties and function to use allow…

JulienKode:
Change privacy of properties and function to use allow people to use it in their own cell custom

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Changed
 
+- Change `handleGesture(touchLocation:)` in `MessageLabel` as open now. 
+[#912](https://github.com/MessageKit/MessageKit/pull/912) by [@julienkode](https://github.com/JulienKode)
+
 - Fixed the way that the Strings and UIImages are parsed in the `InputTextView` to prevent crashes in `parseForComponents()`. 
 [#791](https://github.com/MessageKit/MessageKit/pull/791) by [@nathantannar4](https://github.com/nathantannar4)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Changed
 
-- Change `handleGesture(touchLocation:)` in `MessageLabel` as open now. 
+- Change acl of `handleGesture(touchLocation:)` in `MessageLabel` from internal to open. 
 [#912](https://github.com/MessageKit/MessageKit/pull/912) by [@julienkode](https://github.com/JulienKode)
 
 - Fixed the way that the Strings and UIImages are parsed in the `InputTextView` to prevent crashes in `parseForComponents()`. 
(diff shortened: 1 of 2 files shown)


nathantannar4:
What would be the use case for allowing modification of these properties? They are kept private to prevent unexpected behavior. Moreover, they are set this way to align with the MessagesCollectionViewLayout. So any changes here could mean vastly overriding the layout. 

JulienKode:
As example if we want to create our own TextMessageCell that use the messageLabel, we need 

- messageLabelFont: to give the font to the messageLabel
- handleGesture: in the case the user touch the cell

In addition if we want to add our own Attributes in ranges we need:
- textStorage

May be I'm wrong and other possibilities are possible, I just do not know what is the things to do, if we want to use this view in customs cells

nathantannar4:
So you don't need access to messageLabelFont. That's internal for a reason, used to increase performance. Setting it won't actually change the font. UILabel has a font property for that. 

Not sure about text storage as well. I feel like it should be kept private as if someone wants to create a custom TextCell they don't need access to the text storage. If you look at the file you can were we have made attributes have a didSet that will effect the text storage. Thinik of it this way, it users don't have access to text storage in a UILabel, why should they get access in MessageLabel?

JulienKode:
Ok I understand why we cannot do this for the text storage and the messageLabelFont
TextStorage was only if we want to add some attributes in the same,
But if we allow user to do that via regex it solve this problem 

For `handleGesture ` we cannot put it as public ?

nathantannar4:
`handleGesture` is for internal use, so it would not serve someone who is implementing something custom very well. Should be kept the same. Thanks for the suggestions, but ultimately we have to keep a tight access control on things that when changed will lead to unexpected behaviour. Makes our job at solving issues submitted to GitHub much harder when they are ones such as, "Tapping address in message doesn't work"

JulienKode:
Ok I understand makes sense too,
I’ll try with public properties

SD10:
@nathantannar4 In our app, we made `handleGesture` public because we need it for detector types like mentions -- which MessageKit does not support. We should look into a better solution and also consider that StyledTextView does not have support for detectors (yet) either. Overall, I agree that limiting undefined behavior for now is a good idea

JulienKode:
@SD10 Thanks for theses details

nathantannar4 (changes requested):
Make messageLabelFont and textStorage private again, make handleGesture open

JulienKode:
@nathantannar4 I've did it

nathantannar4:
Looks good, maybe just add a changeling entry and then it can be merged

JulienKode:
>Looks good, maybe just add a changeling entry and then it can be merged

@nathantannar4 ok I’ll add it 😊

nathantannar4 (approved):

JulienKode:
@nathantannar4 theses changes will be in the `2.0` or `2.1` ?

nathantannar4:
2.0 they are non breaking

JulienKode:
@nathantannar4 ok thanks

SD10:

SD10:
@nathantannar4 Thanks for taking lead here, I'm extremely busy. Please ping me when you decide to merge something though. Maybe I could offer some input.

IMO, new features/new changes should not be added to betas. Only bugfixes. This should've been something scheduled for a 2.1 release

JulienKode:
Thanks @SD10 👍

PR #913, opened 2018-10-16
[DetectorType][MessageLabel] Add custom, mention and hashtag detector type 👻

JulienKode:
Adding other `DetectorTypes` such as:

- `.mention`
- `.hashtag`
- `.custom(regex: NSRegularExpression)`

Others elements of the enum are made with `NSDataDetector` which inherit from `NSRegularExpression`

Theses three new element use only `NSRegularExpression`

We can discuss of the API or changes to make here


Sources/Models/DetectorType.swift
```diff
@@ -31,12 +31,11 @@ public enum DetectorType: Hashable {
...
-    case custom(pattern: String)
+    case custom(regex: NSRegularExpression)
 
-    // MARK: - Not supported yet
-
-    //case mention
-    //case hashtag
+    // swiftlint:disable force_try
+    public static let hashtag = DetectorType.custom(regex: try! NSRegularExpression(pattern: "#[a-zA-Z]+", options: []))
+    public static let mention = DetectorType.custom(regex: try! NSRegularExpression(pattern: "@[a-zA-Z]+", options: []))
```
nathantannar4: Numbers should be allowed for mention and hashtag

nathantannar4 (changes requested):
We need some ample test cases to ensure the regex matches all cases for hashtag and mention

nathantannar4:
Please make sure your branch, `JulienKode:CustomDetectorType` is up to date with the latest `development` branch

Sources/Models/DetectorType.swift
JulienKode: Ok I’ll add it thanks

Sources/Models/DetectorType.swift
```diff
@@ -24,19 +24,18 @@
...
+    case custom(regex: NSRegularExpression)
 
-    // MARK: - Not supported yet
-
-    //case mention
-    //case hashtag
-    //case custom
+    // swiftlint:disable force_try
+    public static let hashtag = DetectorType.custom(regex: try! NSRegularExpression(pattern: "#[a-zA-Z]+", options: []))
+    public static let mention = DetectorType.custom(regex: try! NSRegularExpression(pattern: "@[a-zA-Z]+", options: []))
```
SD10: What if I want to provide a different regex for hashtag or mention? For example, this mention regex does not meet the requirements for my app

SD10:
This looks promising ❤️ Thanks for working on this @JulienKode 🥇 💯

Sources/Models/DetectorType.swift
cwalo: I would also argue that the defaults should be alpha-numeric.

JulienKode: @SD10 In fact if the user want another regular expression he have to make a custom one currently, what did you suggest to improve that ?

We use theses two static variable to detect that it's a mention and it's an hashtag to call the right delegate method

SD10: maybe the `hashtag` and `mention` properties can become `var`s so that they're mutable?

JulienKode: @SD10 In fact it can be nice to do that

JulienKode:
@nathantannar4 Thanks it's now sync with the `development` branch

@SD10 I've change it into a var 

There now the testing to do I've see that there are already test for `DetectorType` in `MessageLabelSpec` but they're all commented. Anyone know the reason ? 

If I uncomment theses tests they do not pass but they are right.
If any body know a limitation about that I'm interested 
I'll try to fix that soon

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Changed
 
+- Added  `.hashtag`, .`mention` to detect theses pattern inside the `messageLabel`.
+- Added `.custom(pattern: YOUR_PATTERN)` to `DetectorType` to manage and deal with your own regular expression.
+[#913](https://github.com/MessageKit/MessageKit/pull/913) by [@JulienKode](https://github.com/julienkode).
+
```
SD10: There should only be one entry for this. It should also go under the `Added` section under `Upcoming Release`

JulienKode: @SD10 Thanks I've just fix it

talanov (changes requested):
Could you, please, squash commits into 1?

JulienKode:
@talanov for sure

Sources/Models/DetectorType.swift
```diff
@@ -24,19 +24,18 @@
...
+    case custom(regex: NSRegularExpression)
 
-    // MARK: - Not supported yet
-
-    //case mention
-    //case hashtag
-    //case custom
+    // swiftlint:disable force_try
+    public static var hashtag = DetectorType.custom(regex: try! NSRegularExpression(pattern: "#[a-zA-Z0-9]+", options: []))
+    public static var mention = DetectorType.custom(regex: try! NSRegularExpression(pattern: "@[a-zA-Z0-9]+", options: []))
```
SD10: Do you think we should make these have a minimum of 4 characters to be detected?

CHANGELOG.md
SD10: Sorry but two more things:
- this is also an API breaking change. Can you mark it as **Breaking Change** like the other entries?
- this needs an `### Added` section above the entry

Sources/Models/DetectorType.swift
```diff
@@ -45,6 +44,30 @@ public enum DetectorType {
...
+        case .date:
+            return 1
+        case .phoneNumber:
+            return 2
+        case .url:
+            return 3
+        case .transitInformation:
+            return 4
+        case .custom(let regex):
+            return regex.hashValue
```
SD10: Nit: Can you put the `return` on the same line as the `case` statement?

Sources/Views/MessageLabel.swift
```diff
@@ -301,10 +319,10 @@ open class MessageLabel: UILabel {
 
     private func parse(text: NSAttributedString) -> [NSTextCheckingResult] {
         guard enabledDetectors.isEmpty == false else { return [] }
-        let checkingTypes = enabledDetectors.reduce(0) { $0 | $1.textCheckingType.rawValue }
-        let detector = try? NSDataDetector(types: checkingTypes)
         let range = NSRange(location: 0, length: text.length)
-        let matches = detector?.matches(in: text.string, options: [], range: range) ?? []
+        let matches = enabledDetectors
+            .compactMap { parse($0, for: text, in: range) }
```
SD10: This is worrisome to me 🤔 The old detector code took in all the checking types up front. We're now iterating and parsing the text each time. I'm not sure how `NSDataDetector` operates internally, but if this has gone from parsing 1 time to parsing N times -- we can't handle that overhead

Sources/Models/DetectorType.swift
JulienKode: For sure :)

Sources/Views/MessageLabel.swift
JulienKode: @SD10 May be I can refactor that part to put all `NSTextCheckingResult.CheckingType` and call `NSDataDetector` once. Like before

`NSDataDetector` is a subclass of `NSRegularExpression`

Sources/Models/DetectorType.swift
JulienKode: It can be a nice to do that but I do not know the standard in mention and hashtag. I can add it ? 
And if users want something more customisable they can add their own mention and hashtag

CHANGELOG.md
JulienKode: @SD10 Ok I've just added it

Sources/Views/MessageLabel.swift
```diff
@@ -320,9 +320,22 @@ open class MessageLabel: UILabel {
...
+        let regexs = enabledDetectors
+            .compactMap { parse(custom: $0, for: text, in: range) }
+            .joined()
+        matches.append(contentsOf: regexs)
+
+        // Get all Checking Types of detectors, except for .custom because they contain their own regex
+        let detectorCheckingTypes = enabledDetectors
+            .filter({ !isCustomDetector(for: $0) })
+            .reduce(0) { $0 | $1.textCheckingType.rawValue }
+        if detectorCheckingTypes > 0, let detector = try? NSDataDetector(types: detectorCheckingTypes) {
```
JulienKode, line 348: @SD10 we check `detectorCheckingTypes` because if the value is 0 it cause `NSDataDetector` crash

Sources/Views/MessageLabel.swift
```diff
@@ -320,9 +320,22 @@ open class MessageLabel: UILabel {
...
+            let detectorMatches = detector.matches(in: text.string, options: [], range: range)
```
JulienKode, line 349: @SD10 now `NSDataDetector` is run once for all `NSTextCheckingResult.CheckingType`

Sources/Models/DetectorType.swift
cwalo: @SD10 @JulienKode 4 characters seems arbitrary. what's your thinking @SD10?

Sources/Views/MessageLabel.swift
```diff
@@ -322,6 +353,36 @@ open class MessageLabel: UILabel {
...
 
+    /**
+     Simply check if the detector type is a .custom
+     All .custom contains an `NSRegularExpression`
+
+     - Parameters: detector: `DetectorType` that you want to execute
+
+     - Returns: Bool true if the detector type is a .custom, otherwise false
+     */
+    private func isCustomDetector(for detector: DetectorType) -> Bool {
```
JulienKode: @nathantannar4 may be we can do this operation in one line of code

Sources/Views/MessageLabel.swift
```diff
@@ -301,10 +319,23 @@ open class MessageLabel: UILabel {
...
+
+        // Get matches of all .custom DetectorType and add it to matches array
+        let regexs = enabledDetectors
+            .compactMap { parse(custom: $0, for: text, in: range) }
+            .joined()
+        matches.append(contentsOf: regexs)
+
+        // Get all Checking Types of detectors, except for .custom because they contain their own regex
+        let detectorCheckingTypes = enabledDetectors
+            .filter({ !isCustomDetector(for: $0) })
```
nathantannar4: Remove unneeded brackets, `.filter { !isCustomDetector(for: $0) }`

Sources/Views/MessageLabel.swift
nathantannar4: I think this should be a property of `DetectorType`, not a method in `MessageLabel`

```
enum DetectorType {
...

     var isCustom: Bool {
                   ...
     }
```

Sources/Views/MessageLabel.swift
```diff
@@ -322,6 +353,36 @@ open class MessageLabel: UILabel {
...
+     - Parameters: text: where we apply the regular expression
+     
+     - Returns: an array of `NSTextCheckingResult` that contains all maching for this detector or nil if it fails
+     */
+    private func parse(custom detector: DetectorType, for text: NSAttributedString, in range: NSRange) -> [NSTextCheckingResult]? {
+        switch detector {
+        case .custom(let regex):
+            return regex.matches(in: text.string, options: [], range: range)
+        default:
+            return nil
```
nathantannar4: I think this should be a `fatalError()` as the method is only to be used with custom detector types

Sources/Views/MessageLabel.swift
```diff
@@ -322,6 +353,36 @@ open class MessageLabel: UILabel {
...
+
+    /**
+     Take a custom detector and apply the matching to the text
+     
+     - Parameters: detector: `DetectorType` that you want to execute
+     - Parameters: text: where we apply the regular expression
+     
+     - Returns: an array of `NSTextCheckingResult` that contains all maching for this detector or nil if it fails
+     */
+    private func parse(custom detector: DetectorType, for text: NSAttributedString, in range: NSRange) -> [NSTextCheckingResult]? {
```
nathantannar4: `parseForMatches(with detector: DetectorType, in text: NSAttributedString, for range: NSRange) -> [NSTextCheckingResult]`

nathantannar4:
Very close! I'd like to see the Example updated so people can see these detectors in action 😊

Sources/Views/MessageLabel.swift
JulienKode: Nice idea @nathantannar4

JulienKode:
@nathantannar4 I’ve added examples 

sonnyio:
Hey guys, in MessageLabel.swift, it's calling `handleHashtag` when it should be calling `handleMention`.  Looking forward to this commit being pulled in!

https://github.com/MessageKit/MessageKit/blob/77e9ccda4697d4cafb9854dffc88ffdbdbed7a71/Sources/Views/MessageLabel.swift#L493

JulienKode:
@strujillojr Thanks, It's fixed now

nathantannar4 (changes requested):
@JulienKode can you please add the delegate methods in the Example so users know they exist?

```
func didSelectHashtag(_ hashtag: String) {
        print("Hashtag selected: \(hashtag)")
    }

    func didSelectMention(_ mention: String) {
        print("Mention selected: \(mention)")
    }

    func didSelectCustom(_ pattern: String, match: String?) {
        print("Custom data detector patter selected")
    }
```

nathantannar4:
After example changes it LGTM 💯 @SD10 Should we include this in 2.0.0?

JulienKode:
@nathantannar4 I've update the branch to the latest `development` branch and updated the examples

Sources/Models/DetectorType.swift
SD10: I'm behind the times. I guess 4 characters is arbitrary lol

SD10:
This can't be in 2.0 because we already did a beta. You can't have anything other than bugfixes in a beta or it defeats the purpose of the beta lol. This is an API breaking change so I changed it to be part of the 3.0 milestone.

Example/Sources/Data Generation/SampleData.swift
```diff
@@ -40,8 +40,10 @@ final internal class SampleData {
         case Location = 5
         case Url = 6
         case Phone = 7
-        case Custom = 8
-        
+        case Hashtag = 8
+        case Mention = 9
+        case Custom = 10
```
SD10: Let's make all these cases lowercased, it is not very Swifty 😆 If you could take care of all of them 💯

Sources/Views/MessageLabel.swift
```diff
@@ -313,10 +331,24 @@ open class MessageLabel: UILabel {
 
     private func parse(text: NSAttributedString) -> [NSTextCheckingResult] {
         guard enabledDetectors.isEmpty == false else { return [] }
-        let checkingTypes = enabledDetectors.reduce(0) { $0 | $1.textCheckingType.rawValue }
-        let detector = try? NSDataDetector(types: checkingTypes)
         let range = NSRange(location: 0, length: text.length)
-        let matches = detector?.matches(in: text.string, options: [], range: range) ?? []
+        var matches =  [NSTextCheckingResult]()
```
SD10: Nit: fix this double space please

Sources/Views/MessageLabel.swift
```diff
@@ -334,6 +366,24 @@ open class MessageLabel: UILabel {
...
+    /**
+     Take a custom detector and apply the matching to the text
+     
+     - Parameters: detector: `DetectorType` that you want to execute
+     - Parameters: text: where we apply the regular expression
+     - Parameters: range: where we apply the regular expression
+     
+     - Returns: an array of `NSTextCheckingResult` that contains all maching for this detector or nil if it fails
+     */
+    private func parseForMatches(with detector: DetectorType, in text: NSAttributedString, for range: NSRange) -> [NSTextCheckingResult] {
```
SD10, line 369: Docs are nice for me to understand the code, but fwiw we don't _need_ to document non-public APIs. I can't complain though 🥇

Sources/Models/DetectorType.swift
```diff
@@ -45,6 +45,32 @@ public enum DetectorType {
...
+        }
+    }
+
+    ///The hashValue of the `DetectorType` so we can conform to `Hashable` and be sorted.
+    public var hashValue: Int {
+        return self.toInt()
+    }
+
+    /// Return an 'Int' value for each `DetectorType` type so `DetectorType` can conform to `Hashable`
+    private func toInt() -> Int {
```
SD10, line 66: Are `toInt()` and `hashValue` being used somewhere?  I reviewed fast. Maybe I'm missing something

Sources/Models/DetectorType.swift
```diff
@@ -24,19 +24,19 @@
...
 
-public enum DetectorType {
+public enum DetectorType: Hashable {
 
     case address
     case date
     case phoneNumber
     case url
     case transitInformation
+    case custom(regex: NSRegularExpression)
```
SD10: I'm fine dropping the name on the associated value. Seems annoying 😛

Sources/Views/MessageLabel.swift
```diff
@@ -313,10 +331,24 @@ open class MessageLabel: UILabel {
...
         let range = NSRange(location: 0, length: text.length)
-        let matches = detector?.matches(in: text.string, options: [], range: range) ?? []
+        var matches =  [NSTextCheckingResult]()
+
+        // Get matches of all .custom DetectorType and add it to matches array
+        let regexs = enabledDetectors
+            .filter { $0.isCustom }
+            .map { parseForMatches(with: $0, in: text, for: range) }
+            .joined()
+        matches.append(contentsOf: regexs)
```
SD10, line 342: So it looks like we still parse all the regex and all the other detector types separately? I wonder if it's possible to register our own custom text checking type using `NSDataDetector`.

Fwiw, I love what's happening here -- I just have no time to benchmark what effect this will have on the framework and scroll performance. I have a serious suspicion --- but without benchmarks it's just suspicion

Example/Sources/Data Generation/SampleData.swift
JulienKode: That’s true, It’s not really swift, may I changes all of them in this PR or create another one ? To change others

Sources/Models/DetectorType.swift
JulienKode: It's use because since we have an enum with a type and we store theses `DectectorType` into a dictionary we need to implement the `Hashable` protocol

Sources/Views/MessageLabel.swift
JulienKode: I've did it, thanks

Sources/Views/MessageLabel.swift
JulienKode: Ok I'll remove it in the update

JulienKode:
I think it should be ok now ? @nathantannar4 @SD10

Example/Sources/Data Generation/SampleData.swift
SD10: It can all be done in this PR 👍

JulienKode:
@SD10 @nathantannar4 

I've made some benchmarks here is the results

Here is the configuration of the first test:

img_6597
withoutpr

Here is the configuration of the second test
img_6598
withpr

What do you think ?

nathantannar4:
What are the graphs showing? FPS? CPU %?

JulienKode:
@nathantannar4 FPS in purple 
and upper it's the CPU

JulienKode:
Any update ?

JulienKode:
I've sync this branch with the latest `development` branch

SD10:
@JulienKode It does look like adding those code has some major spikes in the CPU 😞

nathantannar4:
It is optional, personally in my apps I wouldn't use this, I would use attributed strings. But this is a nice to have feature for ease of use

JulienKode:
Ok, I understand, but It’s only in the worst case, if you just handle hashtag the perf is the same 

I know it's a sensitive part of the library

If you have any idea of change to improve the performance. I want to know it

Sources/Models/DetectorType.swift
btomtom5: My opinion would be to adopt something closer to twitter's convention for hashtag because that is what people are used to.

**Cases not handled by the current regex**
#1234 pure numbers can be tags which can often be extensions
#MIT 4 is a pretty arbitrary requirement for the tag length. esp when people tag acryonyms
#swift_ios _ is a pretty common way to delimit words in a tag
discord#chat_room_number  discord uses # inside of their handle to share profiles
#hello#world : Matter of taste but twitter doesn't register this as a tag.

_Suggested Regex:_ "\B#\w*[a-zA-Z]+\w*"

The above regex handles the edge cases listed above except for the last one. It's not possible to implement without a look behind which is a little more complex of a regex.

nathantannar4:
Going to merge this into the 3.0.0-beta branch. While it does come at a performance hit I am going to leave it to the end user to decide if they want this since without using the new detectors there is no difference.

nathantannar4 (approved):

JulienKode:
@nathantannar4 ok it makes sense, if someone have an idea to improve the speed, we also can improve it

Issue #769, opened 2018-07-17
Add another bottom label to the cell


SD10:
Currently, there is an imbalance between the top labels and the bottom label. We have 2 top labels and only 1 bottom label. In JSQMVC they had two of both. We can afford another label because it will be sized statically.

SD10:
Closed by #920

PR #920, opened 2018-10-21
Enhancement/add another bottom label to the cell

maxxx777:
It adds cell bottom label into message content cell.

#769 

the screenshot of the result:
simulator screen shot - iphone x - 2018-10-22 at 00 29 12

Any other comments?
-------------------
-

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone X (Simulator)

**iOS Version:** 12.0

**Swift Version:** 4.2

**MessageKit Version:** 2.0.0-beta.1


Example/Sources/View Controllers/BasicExampleViewController.swift
```diff
@@ -34,6 +34,11 @@ final class BasicExampleViewController: ChatViewController {
         
         messagesCollectionView.messagesLayoutDelegate = self
         messagesCollectionView.messagesDisplayDelegate = self
+        
+        let layout = messagesCollectionView.collectionViewLayout as? MessagesCollectionViewFlowLayout
+        
+        layout?.setMessageIncomingAvatarPosition(AvatarPosition(vertical: .messageBottom))
```
nathantannar4: Isn't this the default position?

Sources/Layout/MessageSizeCalculator.swift
```diff
@@ -181,8 +188,23 @@ open class MessageSizeCalculator: CellSizeCalculator {
...
+    }
+    
+    open func cellBottomLabelAlignment(for message: MessageType) -> LabelAlignment {
+        let dataSource = messagesLayout.messagesDataSource
+        let isFromCurrentSender = dataSource.isFromCurrentSender(message: message)
+        return isFromCurrentSender ? outgoingCellBottomLabelAlignment : incomingCellBottomLabelAlignment
+    }
 
-    // MARK: - Bottom Label
+    // MARK: - Bottom message Label
```
nathantannar4: Capitalize `message`

nathantannar4 (changes requested):
Thanks so much for taking this on!! 🔥 💯

Example/Sources/View Controllers/BasicExampleViewController.swift
maxxx777: Yes, this is not the default position. I reverted back the change not to confuse users and not to complicate the example project.

Sources/Layout/MessageSizeCalculator.swift
maxxx777: Fixed

nathantannar4:
LGTM, @SD10

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -81,6 +89,7 @@ open class MessageContentCell: MessageCollectionViewCell {
     open func setupSubviews() {
         contentView.addSubview(accessoryView)
         contentView.addSubview(cellTopLabel)
+        contentView.addSubview(cellBottomLabel)
```
SD10: Nit: Can you add this subview _after_ the `messageBottomLabel`?

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -253,6 +268,18 @@ open class MessageContentCell: MessageCollectionViewCell {
         cellTopLabel.frame = CGRect(origin: .zero, size: attributes.cellTopLabelSize)
     }
     
+    /// Positions the cell's bottom label.
+    /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
+    open func layoutCellBottomLabel(with attributes: MessagesCollectionViewLayoutAttributes) {
+        cellBottomLabel.textAlignment = attributes.cellBottomLabelAlignment.textAlignment
+        cellBottomLabel.textInsets = attributes.cellBottomLabelAlignment.textInsets
+        
+        let y = messageContainerView.frame.maxY + attributes.messageContainerPadding.bottom + attributes.messageBottomLabelSize.height
```
SD10: For calculating the Y position, can we make it so the `cellBottomLabel` always starts after the `messageBottomLabel` max Y?

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -265,9 +292,9 @@ open class MessageContentCell: MessageCollectionViewCell {
         messageTopLabel.frame = CGRect(origin: origin, size: attributes.messageTopLabelSize)
     }
 
-    /// Positions the cell's bottom label.
+    /// Positions the message bubble's bottom label.
     /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
-    open func layoutBottomLabel(with attributes: MessagesCollectionViewLayoutAttributes) {
+    open func layoutMessageBottomLabel(with attributes: MessagesCollectionViewLayoutAttributes) {
```
SD10, line 297: This is an API breaking change. Let's make sure to add a CHANGELOG entry for this. Please refer to the other breaking change entries and follow that format.

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -46,6 +46,14 @@ open class MessageContentCell: MessageCollectionViewCell {
         return label
     }()
     
+    /// The bottom label of the cell.
+    open var cellBottomLabel: InsetLabel = {
```
SD10, line 50: Let's add a CHANGELOG entry for this new feature 👍

SD10 (changes requested):
Hey @maxxx777, thanks for taking this on 💯 Looks really solid to me. Just please take a look at my suggestions

SD10:
This has to go into 3.0 because it's API breaking and tbh a pretty big layout change. Let's also start using the API breaking label

Sources/Views/Cells/MessageContentCell.swift
maxxx777: I think it's better if the calculation of the layout of one cell does not depend on the layout of another cell. 
For example, the layout of `messageBottomLabel` is not calculated yet. if we gonna use it to calculate the layout of `cellBottomLabel`, we get the wrong layout of `cellBottomLabel`.
We have to agree either to use a certain order of the calculation of labels' layouts in `apply(_:)` function, or depend only on `messageContainerView`s layout in the calculations of labels' layouts.

SD10: Fwiw, the rest of the layout methods rely on a certain order of method calls. I think making them dependent is a more reliable and easier way to do the layout

maxxx777: Ok, did the fixes.

SD10 (approved):
Awesome work here @maxxx777. Sorry for the slow review time. Thanks so much for this 💯

Issue #400, opened 2017-12-13
collectionView's didSelectItemAt: not call when touching Cells


jdriselvato:
## General Information

* MessageKit Version: v0.11.0 (Pods)

* iOS Version(s): 11.1

* Swift Version: 4.0

* Devices/Simulators: Both
 
* Reproducible in ChatExample? (Yes/No):  YES

##  What happened?

- I tap a bubble, avatar, date, title and the respective MessageCellDelegate are called. 
- I tap a bubble, avatar, date, title and collectionView's `didSelectItemAt:` does not get called
- I tap outside of any of those items (the white space around it) `didSelectItemAt:` gets called.

## What did you expect to happen?

- I tap any part of a cell and collectionView's `didSelectItemAt:` should be called

## Why?
I need to identify the MessageType for selected cell.

I did find this: https://github.com/MessageKit/MessageKit/issues/154, which works if you do this:

```
func didTapMessage<T>(in cell: MessageCollectionViewCell<T>) {
        let indexPath = self.messagesCollectionView.indexPath(for: cell)
}
```
Is this the only expected way to get the index?

SD10:
I'm not sure about this one. @nathantannar4 @zhongwuzw do you think we should block the `UICollectionViewDelegate` method `didSelectItemAt:` or allow them to be recognized simultaneously? Have to look into this one

nathantannar4:
I'll look at this when I have a moment. Issue could be because we have our other tap gestures which don't allow for simultaneous recognition with the cell tap gestures. I think it would be ok to block the method as we already have more detailed tap delegate methods. I also don't think a user should be able to tap outside of the message bubble and get the same action as tapping the bubble (which would be the case if a developer placed the login in didSelectItemAt:)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

zhongwuzw:
@SD10 , I think we needn't support `didSelecItemAt:`, we have added a tap gesture on cell's `contentView`, maybe we can add a delegate method in `MessageCellDelegate`, like `didTapMessageCell(in cell: MessageCollectionViewCell, at position: CGPoint)`, and always call it in  `handleTapGesture:` method, like: 

```
    @objc
    open func handleTapGesture(_ gesture: UIGestureRecognizer) {
        guard gesture.state == .ended else { return }

        let touchLocation = gesture.location(in: self)

        switch true {
[...]

Issue #773, opened 2018-07-20
handleTapGesture needs to be moved to MessageCollectionViewCell


SD10:
This needs to be moved to the more general `MessageCollectionViewCell` class

https://github.com/MessageKit/MessageKit/blob/69c5414c224a98891e14984e83228740e4339219/Sources/Views/MessagesCollectionView.swift#L78-L87

maxxx777:
Hi @SD10,

I'd like to take the issue, but I need some clarification:

the idea is to remove the method from `MessagesCollectionView` to avoid calculation of gesture location and cell index path:

```
let touchLocation = gesture.location(in: self)
guard let indexPath = indexPathForItem(at: touchLocation) else { return }
```

either, the idea to move `handleTapGesture` method from `MessageContentCell` into `MessageCollectionViewCell` ?

nathantannar4:
Implemented in #950

PR #922, opened 2018-10-30
Added a `didTapBackground(in:)` method to MessageCellDelegate. 👻

kpennacchia:
Allows apps to detect when the user has tapped on the background of a cell, rather than its content.

The use case I have in an app I am working on is to dismiss the keyboard when the user taps on an empty space of the collection view. Typically I would do this by putting a gesture recogniser on the background view of the collection view, but because the message cells go full width, the touches are intercepted, preventing that approach from working.

#400 

N/A

Any other comments?
-------------------
N/A

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 6 device and iPhone 8 simulator.

**iOS Version:** iOS 11 and iOS 12

**Swift Version:** Swift 4.2

**MessageKit Version:** 2.0.0-Beta.1 (latest commit on the development branch as of this moment).


SD10:
I think this is related to #400. I haven't had time to review this PR. @MessageKit/contributors want to take a look?

kpennacchia:
That looks to be related, yes. I didn't see that when looking through the issues. I suppose the two solutions to choose from are:

1. Introduce a new method on `MessageCellDelegate` to indicate when the user has tapped on the cell and not its content (this pull request).
2. Or change the gesture recognition so that it no longer prevents touches on empty space being forwarded to `collectionView(_:didSelectItemAt:)` as @funct7 [has done](https://github.com/MessageKit/MessageKit/issues/400#issuecomment-427661969).

I think either solution is fine, though solution 1 may be easier for users to discover and reason about any possible side effects IMO.

Skornos:
And what if my `.custom` cell does not inherit from `MessageContentCell`? I am facing this issue right now but this tap gesture will return `nil` on the following condition:

```
    @objc
    open func handleTapGesture(_ gesture: UIGestureRecognizer) {
        ...
        let cell = cellForItem(at: indexPath) as? MessageContentCell
        ...
    }
```

kpennacchia:
This pull request won't solve your issue.

Not to go too far off-topic, and I haven't tried using a cell that doesn't subclass `MessageContentCell ` so I'm only guessing here, but from what I can tell you'll need to setup your own gesture recogniser on your custom cell and somehow prevent the gesture recogniser on the `MessagesCollectionView` from preventing your gesture recogniser on your custom cell from receiving touch events.

If possible, you'll be better off just subclassing `MessageContentCell`.

SD10:
Thanks for the review @Skornos, you bring up an interesting point. 

Maybe we need to move the tap gesture recognition code to `MessagesCollectionViewCell` and then override the behavior in `MessageContentCell`?

kpennacchia:
I think that is how it used to work, but was changed to the current behaviour with https://github.com/MessageKit/MessageKit/pull/417.

funct7:
@Skornos Have you tried [this](https://github.com/funct7/MessageKit/tree/enhancement/did_select)? I think it might be what you're looking for.

BTW, people should not be punished for not subclassing `MessageContentCell`. I'm currently using `MessageKit`, but seriously starting to wonder if this was a good choice. It's too bloated and tries to do too much by overriding too much while breaking the behavior of the parents. 

For one, `MessagesCollectionView` is based on `UICollectionView` but you can't even get the proper behavior when using a non-`MessageContentCell` `UICollectionViewCell` subclass in it. According to LSP, the `MessagesCollectionView` is **not** a subtype of `UICollectionView`. 

As long as this library keeps "adding" more to support the behavior of a "super-type" that is has broken by already adding something, it's doomed for some serious mess.

SD10:
@funct7 Some thoughts:

> BTW, people should not be punished for not subclassing MessageContentCell. I'm currently using MessageKit, but seriously starting to wonder if this was a good choice. It's too bloated and tries to do too much by overriding too much while breaking the behavior of the parents.

I agree, I think the fact that `MessagesCollectionViewCell` doesn't have some of this logic was an oversight when we refactored the cell class hierarchy.

> For one, MessagesCollectionView is based on UICollectionView but you can't even get the proper behavior when using a non-MessageContentCell UICollectionViewCell subclass in it. According to LSP, the MessagesCollectionView is not a subtype of UICollectionView. 

Any suggestions how you would fix this?

nathantannar4:
#773

funct7:
@SD10 Sorry. I'm currently busy working on a service, so I haven't been able to spend ample time to read the code thoroughly. I'll try and contribute on the way and after I finish my current project.

For now, I can only say [my solution](https://github.com/funct7/MessageKit/tree/enhancement/did_select) has served my purposes quite well. I used to mix `MessagesCollectionViewCell` with a subclass of a plain `UICollectionViewCell` and it worked okay as far as I'm concerned. After all, all I did is disable the gesture recognizer when a non-`MessagesCollectionViewCell` is tapped, so that the usual `UICollectionView` can do its thing; by "its thing", I mean a cell selection being handled in the `UICollectionViewDelegate` method `collectionView(_:didSelectItemAt:)`.

As you mentioned, moving the tap gesture handling to individual cells might be even better because that means there is no big brother tap gesture recognizer on the collection view overseeing it all. The `MessagesCollectionView` having a tap gesture recognizer also interfered with simple taps on the view itself, which is another behavior I tried to implement through `touchesBegan`. It would be great to see it go.

Skornos:
Appreciate all the ideas, I have solved the issues by subclassing `MessageContentCell ` and hiding all the unnecessary views.

nathantannar4:
@Skornos I have a PR open at the momemt to move the tap logic to the base cell

SD10:
@kpennacchia I'll accept this, please add a CHANGELOG entry for this change. It should go in the `Added` section and it is API breaking. Please take care to follow the format of the other entries

kpennacchia:
@SD10 Great, that's all done now.

SD10 (approved):

SD10:

PR #923, opened 2018-10-30
Removed repeated test

alephao:
👻 On MessagesViewControllerTests.swift there was a test verifying if the delegate and datasorce were the same object. A similar test can be found on MessagesViewControllerSpec.swift, so I removed it.

The similar test found on MessaesViewControllerSpec is:
https://github.com/MessageKit/MessageKit/blob/801b1e206e733d64c01d447da09cb8666a3005f6/Tests/ControllersTest/MessagesViewControllerSpec.swift#L95-L102


nathantannar4:
Seems similar to #924, can we keep all the test changes in one PR?

alephao:
This and #924 are similar in the sense that they both touch the same file, but they are not the same thing...

This one removes a test that seems unnecessary to me, since there is already a similar one on **MessagesViewControllerSpec.swift**

nathantannar4 (approved):

Issue #866, opened 2018-09-17
Add support another X devices


yauhenipakala:
Hi everyone!

Look at this:
https://github.com/MessageKit/MessageKit/blob/6374b87992447e203eb313fafc3c6d56722e4f62/Sources/Controllers/MessagesViewController%2BKeyboard.swift#L96-L98

New iPhone X devices have another display height:

iPhone XR - 1792px
iPhone XS Max - 2688px

For fix, look at this: https://stackoverflow.com/a/47067296

SD10:
@nathantannar4 More device bounds logic to haunt us :trollface:

austinwright:
Is this still a bug? I believe `iPhoneXBottomInset` was removed (in 2.0.0 ?).

nathantannar4:
I don't think it is

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

gwynantj:
This was fixed in: https://github.com/MessageKit/MessageKit/pull/787

PR #932, opened 2018-11-06
fix top layout constraint

### LINKED ISSUE #931: messagesCollectionView top layout constraint is not correct.
1.0.0
12.0.0
4.0
yes.
In some cases it can be really helpful to provide a short example of your code.
If so, please wrap these code blocks in backticks, like this:

```swift
// the problem is here, the topLayoutGuide.length is 0 in viewDidLoad, the value is set after viewDidAppear.
let top = messagesCollectionView.topAnchor.constraint(equalTo: view.topAnchor, constant: topLayoutGuide.length)
```

wangjiejacques:
it close #931

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -152,7 +152,7 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
     private func setupConstraints() {
         messagesCollectionView.translatesAutoresizingMaskIntoConstraints = false
         
-        let top = messagesCollectionView.topAnchor.constraint(equalTo: view.topAnchor, constant: topLayoutGuide.length)
+        let top = messagesCollectionView.topAnchor.constraint(equalTo: topLayoutGuide.bottomAnchor)
         let bottom = messagesCollectionView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
         if #available(iOS 11.0, *) {
             let leading = messagesCollectionView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor)


nathantannar4:
Does this allow for translucency with the navigation bar? I'm confused as to what the issue is?

SD10:
@wangjiejacques This is not a bug, its to support the behavior @nathantannar4 mentioned. Seems like there is a bug in your usage of MessageKit. You may want to take a look at this [medium article explaining extended layout in iOS](https://medium.com/@wailord/extended-layout-in-ios-pre-ios-11-5eff2debf28). Hope it helps!

wangjiejacques:
I just print `topLayoutGuide.length` in the view did load, the value is 0, and it becomes 64 in the view did appear. as the [doc](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621367-toplayoutguide) says **Query this property within your implementation of the viewDidLayoutSubviews() method.** so I don't know why you try to use the `topLayoutGuide.length` in the view did load.

@nathantannar4 @SD10 the issue is you use `topLayoutGuide.length` in viewDidLoad, sometimes this value is 0, we should use the length after `viewDidLayoutSubviews`, or maybe you want to use `topLayoutGuide.length` as 0 ?

SD10:
@wangjiejacques Yeah, it's very possible that we want to use `0` as the constant now. It could be leftover from the previous code. We've had so much trouble with the collectionView insets that I'm very hesitant to make any changes though. Though we still have to fix #866

wangjiejacques:
@SD10 OK, for the bottom, I think we can fix it by `let bottom = messagesCollectionView.bottomAnchor.constraint(equalTo: bottomLayoutGuide.topAnchor)`.
For the top, if you want to use it as 0, please refactor the code `let top = messagesCollectionView.topAnchor.constraint(equalTo: view.topAnchor, constant: topLayoutGuide.length)` to `let top = messagesCollectionView.topAnchor.constraint(equalTo: view.topAnchor)`. 
For me, I don't think this is a good idea to use it as 0, and it's doesn't work for me, but it's all on you.

nathantannar4:
@wangjiejacques Can you please show the bug as to why it's not working for you? Maybe a screenshot?

wangjiejacques:
@nathantannar4 
I have a message view with only one message, when the keyboard displayed, the message is under the navigation bar.

SD10:
> @SD10 OK, for the bottom, I think we can fix it by let bottom = messagesCollectionView.bottomAnchor.constraint(equalTo: bottomLayoutGuide.topAnchor).

We want to pin to the view not the layout guides

nathantannar4:
@wangjiejacques  can you reproduce it in the example? If so please list the steps. 

SD10:
@wangjiejacques We appreciate you taking the time to submit this PR, I'm going to close due to inactivity. Let us know in your issue if the problem persists

alobaili:
@SD10 @nathantannar4 I just tried MessageKit for the first time and faced this issue. This time I have a GIF:
ezgif-4-0c17aa2afbb4

I think the cause of this behavior is not pinning the `messagesCollectionView` to `safeAreaLayoutGuid.topAnchor`.

Here's the view hierarchy:
image

I think it's best to use `safeAreaLayoutGuid` for top and bottom constraints of the `messagesCollectionView`.

fury21:
Problem still actual

fukemy:
yes, confirm still actual

PR #944, opened 2018-11-17
Add Leading/Trailing padding to AvatarPosition

nathantannar4:
This PR adds the ability to set some horizontal padding to the AvatarPosition. @SD10 I'm not sure if this layout was what you were looking for.

### DIFF
--- a/Sources/Layout/MessagesCollectionViewLayoutAttributes.swift
+++ b/Sources/Layout/MessagesCollectionViewLayoutAttributes.swift
@@ -31,6 +31,7 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
 
     public var avatarSize: CGSize = .zero
     public var avatarPosition = AvatarPosition(vertical: .cellBottom)
+    public var avatarLeadingTrailingPadding: CGFloat = 0
 
     public var messageContainerSize: CGSize = .zero
     public var messageContainerPadding: UIEdgeInsets = .zero
@@ -56,6 +57,7 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
         let copy = super.copy(with: zone) as! MessagesCollectionViewLayoutAttributes
         copy.avatarSize = avatarSize
         copy.avatarPosition = avatarPosition
+        copy.avatarLeadingTrailingPadding = avatarLeadingTrailingPadding
         copy.messageContainerSize = messageContainerSize
         copy.messageContainerPadding = messageContainerPadding
         copy.messageLabelFont = messageLabelFont
@@ -76,7 +78,8 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
         // MARK: - LEAVE this as is
         if let attributes = object as? MessagesCollectionViewLayoutAttributes {
             return super.isEqual(object) && attributes.avatarSize == avatarSize
-                && attributes.avatarPosition == attributes.avatarPosition
+                && attributes.avatarPosition == avatarPosition
+                && attributes.avatarLeadingTrailingPadding == avatarLeadingTrailingPadding
                 && attributes.messageContainerSize == messageContainerSize
                 && attributes.messageContainerPadding == messageContainerPadding
                 && attributes.messageLabelFont == messageLabelFont
--- a/Sources/Views/Cells/MessageContentCell.swift
+++ b/Sources/Views/Cells/MessageContentCell.swift
@@ -183,13 +183,13 @@ open class MessageContentCell: MessageCollectionViewCell {
     /// - attributes: The `MessagesCollectionViewLayoutAttributes` for the cell.
     open func layoutAvatarView(with attributes: MessagesCollectionViewLayoutAttributes) {
         var origin: CGPoint = .zero
-        let padding = attributes.avatarPosition.padding
+        let padding = attributes.avatarLeadingTrailingPadding
 
         switch attributes.avatarPosition.horizontal {
         case .cellLeading:
-            origin.x = padding.left
+            origin.x = padding
         case .cellTrailing:
-            origin.x = attributes.frame.width - attributes.avatarSize.width - padding.right
+            origin.x = attributes.frame.width - attributes.avatarSize.width - padding
         case .natural:
             fatalError(MessageKitError.avatarPositionUnresolved)
         }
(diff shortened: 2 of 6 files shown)


Sources/Models/AvatarPosition.swift
```diff
@@ -73,16 +73,20 @@ public struct AvatarPosition {
     
     // The horizontal position
     public var horizontal: Horizontal
+
+    // The horizontal padding
+    public var padding: HorizontalEdgeInsets
```
SD10: I'm not sure if this should be a property or an associated value of `Horizontal` 🤔

SD10 (changes requested):
@nathantannar4 This is what I was looking for 👍 However, there's still work to be done here. The padding now impacts the size of the message container view

nathantannar4:
@SD10 I only adjusted the origin, not the size of the message container view. 

Sources/Models/AvatarPosition.swift
nathantannar4: I thought we should stick with the same pattern as LabelAlignment. Making it an associated value also makes it kind of a pain to read from. Are there any benefits in making it an associated value?

SD10: Maybe call it horizontal padding then 👍

SD10:
@nathantannar4 Padding reduces the amount of available size for the message label

Sources/Models/AvatarPosition.swift
nathantannar4: 👍🏻

nathantannar4:
So we should be effecting the size of the message container?

SD10:
Yeah @nathantannar4 you have to subtract the size of the padding from the available width of the container

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -246,7 +246,8 @@ open class MessageContentCell: MessageCollectionViewCell {
             fatalError(MessageKitError.avatarPositionUnresolved)
         }
 
-        messageContainerView.frame = CGRect(origin: origin, size: attributes.messageContainerSize)
+        let size = CGSize(width: attributes.messageContainerSize.width - avatarPadding.horizontal, height: attributes.messageContainerSize.height)
```
SD10: The frame can't be modified here, it needs to be correct inside of the `MessagesCollectionViewLayoutAttributes`

This is where you need to be making the change:

https://github.com/MessageKit/MessageKit/blob/master/Sources/Layout/MessageSizeCalculator.swift#L227-L233

SD10: It can't size the text properly if it doesn't know how much padding is consuming space

nathantannar4: Ah thats where it should be modified.

Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -235,11 +236,12 @@ open class MessageContentCell: MessageCollectionViewCell {
...
         }
 
+        let avatarPadding = attributes.avatarPosition.padding
         switch attributes.avatarPosition.horizontal {
         case .cellLeading:
-            origin.x = attributes.avatarSize.width + attributes.messageContainerPadding.left
+            origin.x = attributes.avatarSize.width + attributes.messageContainerPadding.left + avatarPadding.horizontal
         case .cellTrailing:
-            origin.x = attributes.frame.width - attributes.avatarSize.width - attributes.messageContainerSize.width - attributes.messageContainerPadding.right
+            origin.x = attributes.frame.width - attributes.avatarSize.width - attributes.messageContainerSize.width - attributes.messageContainerPadding.right - avatarPadding.horizontal
```
SD10: We use `horizontal` here because we need to offset it the total size to layout the container, right?

Sources/Models/AvatarPosition.swift
```diff
@@ -73,16 +73,20 @@ public struct AvatarPosition {
...
-    public init(horizontal: Horizontal, vertical: Vertical) {
+    public init(horizontal: Horizontal, vertical: Vertical, padding: HorizontalEdgeInsets = .zero) {
         self.horizontal = horizontal
         self.vertical = vertical
+        self.padding = padding
     }
 
-    public init(vertical: Vertical) {
-        self.init(horizontal: .natural, vertical: vertical)
+    public init(vertical: Vertical, padding: HorizontalEdgeInsets = .zero) {
```
SD10: Is this the current default in MessageKit? I can't remember

Sources/Models/AvatarPosition.swift
SD10: I feel like this property should still read `horizontalPadding`. What do you think? I know it reads `horizontalPadding.horizontal` then

SD10 (commented):
Just some questions, looking good

SD10:
@nathantannar4 I also wanted to ask if you see any benefit of not making this property part of `AvatarPosition`. Instead, match the same API as `accessoryViewPadding`? 

Better to ask now than change later

Sources/Models/AvatarPosition.swift
nathantannar4: Yes I matched the default

nathantannar4:
@SD10 it's currently apart of `AvatarPosition`

SD10:
@nathantannar4 I meant should it _*not*_ be to match the API of `accessoryViewPadding`?

nathantannar4:
Will refactor to only allow leading or trailing padding passed on horizontal position

nathantannar4:
simulator screen shot - iphone xs - 2018-12-09 at 16 34 43

nathantannar4:
@SD10 This is the result with leading/trailing padding set to 20. Thoughts on the overlap with the top/bottom labels. Do you think we should also use the padding to inset their layout? Or leave that to devs to account for when setting their padding.

SD10:
@nathantannar4 I think that we should leave that to the devs to account for 👍

nathantannar4:
Agreed 👍 Changelog added

SD10:
I'm 99% positive this change is ok but I'll wait to merge on the weekend when I can review in more depth

Sources/Layout/MessagesCollectionViewLayoutAttributes.swift
```diff
@@ -81,7 +83,8 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
         // MARK: - LEAVE this as is
         if let attributes = object as? MessagesCollectionViewLayoutAttributes {
             return super.isEqual(object) && attributes.avatarSize == avatarSize
-                && attributes.avatarPosition == attributes.avatarPosition
```
SD10, line 84: Good catch 😱

SD10 (approved):

SD10:
Thanks for taking care of this @nathantannar4 💯 🥇 Sorry for long review time once again 🙃 Last weekend was traveling for work

PR #955, opened 2018-12-16
add getting started section to README

TomWFox:
There is no reference to the documentation in the README so these changes help to make it quick and easy for people to get started with the library and point the right places for help.

no

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,13 @@ To integrate MessageKit using Carthage, add the following to your `Cartfile`:
 github "MessageKit/MessageKit"
 ````
 
+## Getting Started
+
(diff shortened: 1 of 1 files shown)


SD10 (approved):
@Tom-Fox1 Awesome! 💯 Thanks!!

SD10:

TomWFox:
I would be interested to join but I can't dedicate much if any time on a regular basis so it would just have to be as and when.

The other question I have is what restrictions would there be on what I could do? - I have a fair amount of experience using git/gitHub as an individual but not so much within an organisation so I would be worried about causing problems etc.

SD10:
@Tom-Fox1 

> I would be interested to join but I can't dedicate much if any time on a regular basis so it would just have to be as and when.

You should never feel any pressure to contribute to open source! We don't require anything from you 😃 Joining the org is more to reward you for contributing to the project and making contributing next time even easier

> The other question I have is what restrictions would there be on what I could do? - I have a fair amount of experience using git/gitHub as an individual but not so much within an organisation so I would be worried about causing problems etc.

This is fine! You can always ask questions if you're unsure of something. We have a lot of restrictions in place so you really can't do anything too harmful. And if a mistake happens, don't panic, we can fix it 👍

PR #1010, opened 2019-03-06
Update documentation for `MessageSizeCalculator.messageContainerSize(at:)'

whoyawn:
In the example app, only `sizeForItem(at:)` was being overridden. I was upgrading from 0.13 and I tried borrowing the example code, but my view wouldn't appear. Turns out `messageContainerSize` set my content view to 0. It wasn't clear what methods I had to override, and the original comment is hidden under the function instead of being part of the Xcode hover-documentation.


nathantannar4:
So this is more an issue with Xcode's documentation of functions. We do have docs for that method but its in the base class `CellSizeCalculator`.

PR #1027, opened 2019-03-30
Revive carthage compatibility

lunij:
It basically makes `master` compilable again.

In the case you never checked out MessageKit repository before, you have to run `carthage bootstrap`. This command fails because it cannot build _Nimble_ and _Quick_, so I just updated the versions of both. After that bootstrapping will succeed.

When you try to compile MessageKit using Xcode it will immediately fail because of two SwiftLint rules, 
`explicit_acl` and `explicit_top_level_acl` (SwiftLint version 0.31.0). Both are configured to _error_ in `.swiftlint.yml`. To respect the configuration, I just disabled them individually in code. There is no other way to fix them because Xcode is otherwise complaining about _redundancy_ or _protocol conformance_. And because Xcode/Swift feedback has always a higher priority than a linter, the most logical decision is to just disable the linting individually.

My recommendation however is to loosen the SwiftLint configuration for both rules because of the following reasons:

1. They are conflicting with Swift compiler warnings.
2. Configuring them to _error_ prevents compiling the project itself which can lead to Carthage incompatibility.
3. The `master` branch shows multiple occurrences of these linting errors and _no one cares_.

My second recommendation is to get rid of the two _avoidable_ dependencies **Nimble** and **Quick** to guarantee Carthage compatibility or you appoint a responsible person to take care of the dependencies.

Does not seem to be the case.

Where has this been tested?
---------------------------
**Devices/Simulators:** Xcode 10.2 Simulators

**iOS Version:** 12.2

**Swift Version:** 4.2

**MessageKit Version:** Seems to be 3.0.0-pre or something similar.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,2 +1,2 @@
-github "Quick/Quick" ~> 1.2.0
-github "Quick/Nimble" ~> 7.0.3
+github "Quick/Quick" ~> 2.0.0
+github "Quick/Nimble" ~> 8.0.1
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,3 +1,3 @@
 github "MessageKit/MessageInputBar" "0.4.0"
-github "Quick/Nimble" "v7.3.1"
-github "Quick/Quick" "v1.3.2"
+github "Quick/Nimble" "v8.0.1"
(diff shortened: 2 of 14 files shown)


nathantannar4 (approved):

SD10:

PR #911, opened 2018-10-15
Typing Indicator Support

nathantannar4:
This PR adds TypingIndicator support by adding a new supplementary view rather than reserving the last section (as done in #776). 

The default typing indicator is an iMessage style typing bubble out of the box but you can also implement your own typing indicator cells!

This implementation is still not complete but is in a functional state to compare against #776. 

The goal should be to determine which is a better way of adding this feature, reserving a section, through a supplementary view, or perhaps another method entirely.

PR #1038, opened 2019-04-03
Typing Indicator Support

nathantannar4:
Fresh PR continuing the work from #911


ahmedAlmasri (approved):

ahmedAlmasri:
@nathantannar4 it's working fine , But i have a question 

there are the two typing functions the one it's showing typing view and other not show

call direct func from MessagesViewController class like this 
 `setTypingIndicatorViewHidden(!isTyping, animated: true)` it's showing typing view

call from collection view like this `self.messagesCollectionView.setTypingIndicatorViewHidden(!isTyping)` typing view not showing

nathantannar4:
@ahmedAlmasri great point, I should restrict the access control so it cannot be called on the `MessagesCollectionView`

PR #1082, opened 2019-05-31
Pass back MessageLabel reference in MessageLabelDelegate methods

marcetcheverry:
—————————————————————————

It adds a reference to the `MessageLabel` in all `MessageLabelDelegate` methods. This is useful if you want to present a popover based on a data detector tap, you need a reference to a UIView's `frame`.

I think there can be improvements elsewhere for this style of passing back a reference, but for now I just need this.

Any other comments?
-------------------

Where has this been tested?
---------------------------

**iOS Version:** 12.2

**Swift Version:** Swift 5

 👻

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The changelog for `MessageKit`. Also see the [releases](https://github.com/MessageKit/MessageKit/releases) on GitHub.
 
+## Upcoming Release
+
+### Added
+
+- **Breaking Change** Pass back MessageLabel reference in MessageLabelDelegate methods. [#1082](https://github.com/MessageKit/MessageKit/pull/1082) by [@marcetcheverry](https://github.com/marcetcheverry)
+
 ## 3.0.0
 
 ### Dependency Changes
--- a/Example/Sources/View Controllers/ChatViewController.swift
+++ b/Example/Sources/View Controllers/ChatViewController.swift
@@ -258,35 +258,35 @@ extension ChatViewController: MessageCellDelegate {
 
 extension ChatViewController: MessageLabelDelegate {
     
-    func didSelectAddress(_ addressComponents: [String: String], in messageLabel: MessageLabel) {
+    func didSelectAddress(_ addressComponents: [String: String], in messageLabel: MessageLabel, touchLocation: CGPoint) {
         print("Address Selected: \(addressComponents)")
     }
     
-    func didSelectDate(_ date: Date, in messageLabel: MessageLabel) {
+    func didSelectDate(_ date: Date, in messageLabel: MessageLabel, touchLocation: CGPoint) {
         print("Date Selected: \(date)")
     }
     
-    func didSelectPhoneNumber(_ phoneNumber: String, in messageLabel: MessageLabel) {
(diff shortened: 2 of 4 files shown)


nathantannar4:
This is a pretty significant breaking change, as it would invalidate everyone's current delegate functions. We have to be careful with breaking changes, especially when they don't address a specific widespread issue. 

Do you have a specific usecase where you think this was needed? Perhaps it can be solved another way. 

marcetcheverry:
From the PR description:

`This is useful if you want to present a popover based on a data detector tap, you need a reference to a UIView's frame.`

Presenting pop overs (or an action sheet) from a tapped data detector on iPad require an anchor. This requires a touch location. A touch location CGPoint is useless without a reference to the view it happened in.

All delegate methods in Apple frameworks pass back a reference to the object that calls it, this is a widespread convention, see `UITableViewDelegate` for example. The delegates should have been designed like this from the start.

The breaking change is referenced in the `CHANGELOG`.

I am unsure why I am receiving such critical feedback of all the pull requests I am submitting without feedback on how else they could be achieved. Lots of practices in coding are left up to interpretation as to what is better, and MessageKit is full of those choices many of which are questionable. These solve real implementation problems we are facing and it would be nice to get them merged.

Alternative solutions: Pass an indexPath like UITableVIew does? Do you then have to fetch the cell, and the label, and assume that it was tapped in the label?. You still need to pass back the touch location.

Maybe a target/action subscription pattern like UIKit uses for UIControls? You'd still get back a reference to the control at least with this pattern.

As far as 'significant breaking change', it is literally a 1 minute task to copy paste a new method signature, even less so if these have default implementations. Swift sadly does not have optional protocol methods. Either way, this is Swift programming, everyone is used to breaking code changes, and this one enables a key feature, people should version pin their Pods and expect possible breaking changes in major releases. I understand this is a framework, so **we can always mark the existing methods as deprecated and include the new ones.**

marcetcheverry:
Here is an example of Apple's Messages.app on iPad showing a `UIActionSheet` in a popover style anchored to a tapped data detector in a message bubble

IMG_0084

nathantannar4:
Ok I support this, but we need to deprecate the old ones and still have them work

SD10:
I think this is a fine improvement to make here. I'm not concerned about a breaking change of this caliber. API changes will be inevitable, all I want to avoid is devastating API changes.

I think this is good to go but lets schedule it for a release 👍

marcetcheverry:
I think there is no way to deprecate protocol members in Swift without requiring our implementers to implement both versions of each method. If we keep both version of the method we will have to call both, as Swift has no runtime checking like Objective C does.

Everything that is stated in a protocol in Swift must be implemented, even if marked deprecated, I don't think Apple runs into this problem when they deprecate stuff because 99% of their frameworks are in Objective-C which is much more flexible with protocols.

If we make the protocol @objc then people will have to conform with @objc. So the transition won't be transparent.

What shall we do here, do you have any suggestions? Feel free to try out the project to come up with a solution. I've checked in some deprecation warnings so you can test them.

marcetcheverry:
Moving to private fork.

PR #1084, opened 2019-06-05
Add loading indicator to AudioMessageCell

marcetcheverry:
This adds a loading indicator to the audio cell that hides the play/pause button when made visible. This is what WhatsApp does for audio messages. **It is unlikely that in a real world app you would have local files ready to go for `AVAudioPlayer` as per the example implementation audio controller, these need to be loaded from the server as they are tapped.**

Of course you can override and return a custom cell for audio messages, but I think `AudioMessageCell` is decent enough, and the example audio controller can be updated to add a single `NSURLSessionDownloadTask`.

Sadly `AVAudioPlayer` is local only, and `AVPlayer` is quite a departure (see the ridiculous use of KVO and all the logic needed to handle buffering, been there, done that), so this PR can include in the future an updated BasicAudioController to handle remote audio files.

The only drawback to this PR is that if you do truly have local audio files only, `AudioMessageCell` creates an activity indicator for nothing, but that is a drawback of how cells are architected in `MessageKit`: `setupSubviews` is called without any context as to what type of message you are setting up for, and whether the audio URL is a `file:///` url or a remote asset.

Where has this been tested?

Devices/Simulators: iPhone XS

iOS Version: 12.2

Swift Version: Swift 5

MessageKit Version: 3.0.0-swif5

👻

### DIFF
--- a/Sources/Views/Cells/AudioMessageCell.swift
+++ b/Sources/Views/Cells/AudioMessageCell.swift
@@ -83,7 +83,6 @@ open class AudioMessageCell: MessageContentCell {
     open func setupConstraints() {
         playButton.constraint(equalTo: CGSize(width: 25, height: 25))
         playButton.addConstraints(left: messageContainerView.leftAnchor, centerY: messageContainerView.centerYAnchor, leftConstant: 10)
-        activityIndicatorView.constraint(equalTo: CGSize(width: 25, height: 25))
         activityIndicatorView.addConstraints(centerY: playButton.centerYAnchor, centerX: playButton.centerXAnchor)
         durationLabel.addConstraints(right: messageContainerView.rightAnchor, centerY: messageContainerView.centerYAnchor, rightConstant: 15)
         progressView.addConstraints(left: playButton.rightAnchor, right: durationLabel.leftAnchor, centerY: messageContainerView.centerYAnchor, leftConstant: 5, rightConstant: 5)
--- a/Sources/Views/Cells/AudioMessageCell.swift
+++ b/Sources/Views/Cells/AudioMessageCell.swift
@@ -47,7 +47,7 @@ open class AudioMessageCell: MessageContentCell {
         return durationLabel
     }()
 
-    private lazy var activityIndicatorView: UIActivityIndicatorView = {
+    public lazy var activityIndicatorView: UIActivityIndicatorView = {
         let activityIndicatorView = UIActivityIndicatorView(style: .gray)
         activityIndicatorView.hidesWhenStopped = true
         activityIndicatorView.isHidden = true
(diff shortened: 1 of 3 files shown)


Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -47,25 +47,51 @@ open class AudioMessageCell: MessageContentCell {
         return durationLabel
     }()
 
+    private lazy var activityIndicatorView: UIActivityIndicatorView = {
```
nathantannar4: I think we would be ok to keep this public, so others could change the style if they have a darker message bubble color

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -47,25 +47,51 @@ open class AudioMessageCell: MessageContentCell {
...
+        return activityIndicatorView
+    }()
+
     public lazy var progressView: UIProgressView = {
         let progressView = UIProgressView(progressViewStyle: .default)
         progressView.progress = 0.0
         return progressView
     }()
 
+    public var isLoadingIndicatorVisible: Bool {
```
nathantannar4: The naming of this variable doesn't line up with the value.
ex: Given `activityIndicatorView.isHidden = true`, `isLoadingIndicatorVisible` would return `true`, but that would indicate that `activityIndicatorView.isHidden` is false 

If we make `activityIndicatorView` `public`, I think we could actually move this to a function. Like

`startLoadingAnimation()` and `stopLoadingAnimation()` as in its current form I think its less obvious to expect that the `playButton` becomes hidden

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -47,25 +47,51 @@ open class AudioMessageCell: MessageContentCell {
...
+        }
+    }
+    
     // MARK: - Methods
 
     /// Responsible for setting up the constraints of the cell's subviews.
     open func setupConstraints() {
         playButton.constraint(equalTo: CGSize(width: 25, height: 25))
-        playButton.addConstraints(left: messageContainerView.leftAnchor, centerY: messageContainerView.centerYAnchor, leftConstant: 5)
+        playButton.addConstraints(left: messageContainerView.leftAnchor, centerY: messageContainerView.centerYAnchor, leftConstant: 10)
```
nathantannar4: What was the reason behind adjusting the left constant?

nathantannar4 (changes requested):
Thanks for taking the time to put this together! I just had a few suggestions/questions.

Also, would you mind making a entry in the `CHANGELOG.md` file in the `UPCOMING RELEASE` section at the top? You may need to add `## Upcoming Release` if its not there

Sources/Views/Cells/AudioMessageCell.swift
marcknock: OK.

marcetcheverry: OK.

Sources/Views/Cells/AudioMessageCell.swift
marcetcheverry: I generally like my approach, but I removed this since the `activityIndicatorView` is now public. Users can do whatever they want with it and I moved this approach into an extension in my codebase.

Sources/Views/Cells/AudioMessageCell.swift
marcetcheverry: To answer the question, probably a remnant of earlier attempts at positioning.

`UIActivityIndicatorView` is not generally a component that can be resized by changing the frame, so only certain styles of it will fit properly in this case. But in any case, users can subclass and override their layout.

nathantannar4 (approved):

nathantannar4:
Looks great! Thank you!

PR #1090, opened 2019-06-11
Support for tapping NSTextAttachments in MessageLabel

marcetcheverry:
—————————————————————————

In the case of a `MessageLabel` backed by an `NSAttributedString` this allows you to embed an `NSTextAttachment` (or a subclass) and to receive tap events from it.

A real life use case of this is adding file attachments to a text message. The screenshot below (pixelated some identifying information) shows an example of embedding an image based `NSTextAttachment` which can be tapped.

Screen Shot 2019-06-11 at 10 07 27

Any other comments?
-------------------

Where has this been tested?
---------------------------

**iOS Version:** 12.2

**Swift Version:** Swift 5

 👻

### DIFF
--- a/Sources/Views/MessageLabel.swift
+++ b/Sources/Views/MessageLabel.swift
@@ -459,12 +459,10 @@ open class MessageLabel: UILabel {
             }
         }
 
-        if NSTextAttachment.character == (textStorage.string as NSString).character(at: index) {
-            let attributes = textStorage.attributes(at: index, effectiveRange: nil)
-            if let attachment = attributes[NSAttributedString.Key.attachment] as? NSTextAttachment {
-                delegate?.didSelectTextAttachment(attachment, characterIndex: index, in: self, touchLocation: touchLocation)
-                return true
-            }
+        let attributes = textStorage.attributes(at: index, effectiveRange: nil)
+        if let attachment = attributes[NSAttributedString.Key.attachment] as? NSTextAttachment {
+            delegate?.didSelectTextAttachment(attachment, characterIndex: index, in: self, touchLocation: touchLocation)
+            return true
         }
 
         return false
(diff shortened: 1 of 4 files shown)


Sources/Views/MessageLabel.swift
```diff
@@ -446,7 +446,7 @@ open class MessageLabel: UILabel {
 
     }
 
-  open func handleGesture(_ touchLocation: CGPoint) -> Bool {
+    open func handleGesture(_ touchLocation: CGPoint) -> Bool {
```
marcetcheverry, line 449: Fixed indentation on this method

nathantannar4:
I think this could be a useful addition, and it doenst break anything. 

I'm regards to the delegate method, what is the rationale behind including the character index? Also I'm not sure how I feel about passing back a reference to the `MessageLabel`

marcetcheverry:
You either pass the characterIndex or the text attachment. One or the other, otherwise you can't act on it. There really isn't any cost to passing both. If you want to remove one or the other that is fine, I am happy to do it so we can get this merged.

Text attachments can be subclassed to include more information, such as a URL to act on. This is why it is passed back.

It passes back a `MessageLabel` because of the rationale behind https://github.com/MessageKit/MessageKit/pull/1082. There is absolutely no cost in passing back a MessageLabel.

`NSAttributedString`s can be expensive to create, and if they are simply created for passing it to a `MessageLabel` you may not have access to it elsewhere, with the reference passed back in the delegate you can access the attributedString that is already created and rendered, and work with the `characterIndex` from there. `NSTextAttachment`'s have no properties that reference their position in an attributed string, hence why you get both the attachment and the position, convenience to do whatever you want with that data.

nathantannar4:
Just have to test this

nathantannar4:
This looks good, because of the delegate pattern change will wait for #1082 first

marcetcheverry:
Moving to private fork.

fukemy:
hi, can u share the code how u put attachment image like this? Thanks
image

PR #1091, opened 2019-06-11
Add support for styling NSLinkAttribute with existing urlAttributes in MessageLabel

marcetcheverry:
—————————————————————————

Since `MessageLabel` supports handling `NSLinkAttribute` as tappable links, this add support for styling the links with the same styles as a URL data detector.

`UILabel` has no option to override the default styling that it applies to `NSLinkAttribute`, this faculty is reserved for `UITextView`'s `linkTextAttributes`.

The only way to imitate this behavior is to remove the `NSLinkAttribute` before drawing and to apply our own attributes (which have already been detected by the existing code in `MessageLabel`).

With the current behavior you just get a blue style which is the default `UILabel` style.

Below is a real world example of using `NSLinkAttribute` to highlight a filename which will open an url when tapped.

Screen Shot 2019-06-11 at 10 07 27

**A future PR may want to implement a highlight/touch down attributed style for these links**

Any other comments?
-------------------

Where has this been tested?
---------------------------

**iOS Version:** 12.2

**Swift Version:** Swift 5

 👻

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## 3.0.0
 
+## Upcoming Release
+
+### Added
+
+- Add support for styling NSLinkAttribute with existing urlAttributes in MessageLabel. [#1091](https://github.com/MessageKit/MessageKit/pull/1091) by [@marcetcheverry](https://github.com/marcetcheverry)
+
+## Upcoming Release
 ### Dependency Changes
 
 - **Breaking Change** The dependency `MessageInputBar` was replaced with `InputBarAccessoryView`. As `MessageInputBar` was previously a fork this means no functionality has been lost but improvements and bug fixes will be present. `InputBarAccessoryView` has more of a following outside of `MessageKit` making its development faster than `MessageInputBar`. Maintaining two versions only increased the workload. You can find the changelog for `InputBarAccessoryView` [here](https://github.com/nathantannar4/InputBarAccessoryView/blob/master/CHANGELOG.md).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,14 @@
 
 The changelog for `MessageKit`. Also see the [releases](https://github.com/MessageKit/MessageKit/releases) on GitHub.
 
-## 3.0.0
-
 ## Upcoming Release
 
 ### Added
 
 - Add support for styling NSLinkAttribute with existing urlAttributes in MessageLabel. [#1091](https://github.com/MessageKit/MessageKit/pull/1091) by [@marcetcheverry](https://github.com/marcetcheverry)
 
-## Upcoming Release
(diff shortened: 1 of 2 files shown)


nathantannar4:
Could you please elaborate on why we would want to remove the link attribute? It seems odd that if a user specifies an attributed string to have a link attribute that it gets removed. This could break people's code. 

If specific highlighting is wanted, is there a problem with a user just not using the `link` detector type?

marcetcheverry:
It isn't removed in the attributedString, but in the one that gets rendered through the text storage. It is the only way to actually make `UILabel` render a `NSLinkAttribute` with the attributes you want (in this case, MessageLabel parses that it was a NSLinkAttribute and takes over with its own attributes and link handling mechanisms). 

Otherwise if the NSLinkAttribute is present but you have attributes such as foregroundColor they will get ignored. You will get the default styling no matter what. This is why `UITextView` has that property to override link styling.

The link detector type only works with full URLs, and not arbitrary words marked as links, as in the screenshot shown. See "Attempts to locate URL links." in the documentation. Using a custom regex is not a proper solution.

This does not break people's code, links are still drawn and handled the same way as before, in fact, `MessageLabel` had existing code to detect the attribute and handle those taps. This piece of code just ensures that `NSTextStorage and NSLayoutManager` draws them the same as 'data detected' links, and not with the default UILabel style.

No data is lost, it is a rendering choice to use urlAttributes for NSLinkAttribute as well.

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ## 3.0.0
 
+## Upcoming Release
+
+### Added
+
+- Add support for styling NSLinkAttribute with existing urlAttributes in MessageLabel. [#1091](https://github.com/MessageKit/MessageKit/pull/1091) by [@marcetcheverry](https://github.com/marcetcheverry)
+
+## Upcoming Release
```
nathantannar4: Duplicate header here

Sources/Views/MessageLabel.swift
```diff
@@ -273,6 +273,11 @@ open class MessageLabel: UILabel {
             guard let rangeTuples = rangesForDetectors[detector] else { continue }
 
             for (range, _)  in rangeTuples {
+                // This will enable us to attribute it with our own styles, since `UILabel` does not provide link attribute overrides like `UITextView` does
+                if detector.textCheckingType == .link {
+                    mutableAttributedString.removeAttribute(NSAttributedString.Key.link, range: range)
+                }
```
nathantannar4, line 279: Ok looks good!

CHANGELOG.md
marcetcheverry: Yeah sorry about that, it is hard to keep track of all the CHANGELOGs with multiple branches and the one master integration branch I keep. It would be easier if the CHANGELOGs were only changed by maintainers on merge.

PR #1098, opened 2019-06-18
Make messageLabelInsets(for message: MessageType) open so it can be overriden

marcetcheverry:
—————————————————————————

This makes `messageLabelInsets(for message: MessageType)` an open function so subclasses of `TextMessageSizeCalculator` can override it.

While it is nice that the `incomingMessageLabelInsets` and `outgoingMessageLabelInsets` variables are open, sometimes you want to override insets beyond the basics based on a property of a `MessageType`. A real world use case is if you wanted to render an email message with different insets as opposed to a text message.

A sample implementation of this would involve subclassing this calculator to override the `messageLabelInsets(for message: MessageType)` function.

Any other comments?
-------------------

Where has this been tested?
---------------------------

**iOS Version:** 12.2

**Swift Version:** Swift 5

 👻

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Lazily initialize the MessageInputBar on MessagesViewController. [#1092](https://github.com/MessageKit/MessageKit/pull/1092) by [@marcetcheverry](https://github.com/marcetcheverry)
 
+- Make messageLabelInsets(for message: MessageType) open so it can be overriden in TextMessageSizeCalculator. [#1098](https://github.com/MessageKit/MessageKit/pull/1098) by [@marcetcheverry](https://github.com/marcetcheverry)
+
 ## 3.0.0
 
 ### Dependency Changes
--- a/Sources/Layout/TextMessageSizeCalculator.swift
+++ b/Sources/Layout/TextMessageSizeCalculator.swift
@@ -31,7 +31,7 @@ open class TextMessageSizeCalculator: MessageSizeCalculator {
 
     public var messageLabelFont = UIFont.preferredFont(forTextStyle: .body)
 
-    internal func messageLabelInsets(for message: MessageType) -> UIEdgeInsets {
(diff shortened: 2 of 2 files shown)


nathantannar4:
@marcetcheverry I understand what your trying to achieve here but I don't think this change is needed. I think the strategy you should follow is have a separate size calculator for each type, rather than using the wake size calculator to calculate sizes for multiple types. 

For example, we use 3 `TextMessageSizeCalculator`s, one being for emojis. You could add a 4th in your project for emails. The override `cellSizeCalculatorForItem` for your logic of determining which calculator to use. 

I think Steven made `messageLabelInsets` as more of helper function. 

Maybe I'm missing the larger picture of what you're trying to achieve, if so please let me know which part of my suggested strategy wouldn't work. 

marcetcheverry:
I think the idea is that we want all of the behavior of TextMessageSizeCalculator but just with custom insets. If we made another custom type it would require copying a ton of code rather than just subclassing and overriding one method. Either way you are using a custom size calculator, this just allows you to not have to copy the entire implementation on each MessageKit release.

nathantannar4:
You wouldn't have to copy code. Did you look at the example of the emoji message size calculator? You would just have to add a new one there, and you would be able to just modify the properties. The only thing that would need subclassing is the `MessageCollectionViewFlowLayout`

marcetcheverry:
Subclassing MessagesCollectionViewFlowLayout requires overriding cellSizeCalculatorForItem which means copying a certain amount of code to handle special cases inside of text/attributedText switch case to return my modified size calculator.

It can’t be overriden cleanly without duplicating some code as far as I know. 

Implementing a custom size calculator which means even deeper changes in my codebase and even deeper changes to TextMessageSizeCalculator to handle .custom wrapped objects.

nathantannar4:
@marcetcheverry it is an internal function that purely exists as a helper to determine which inset, incoming or outgoing to use. For that reason this should not be overridden to change that logic. 

From an API development standpoint, why have inset properties and also a function that returns insets? That makes the properties irrelevant. 

I recommend the strategy I listed above. Else if you want to go with overriding methods, I suggest overriding the methods that use `meseageLabelInsets(...)`

Issue #993, opened 2019-02-21
Bottom inset issue when (2nd) expected Keyboard Notification is not sent.


austinwright:
This is an odd issue / edge case related to https://github.com/MessageKit/MessageKit/pull/787:

Overview:

When stepping through the `MessagesViewController+Keyboard.handleKeyboardDidChangeState(_:)` function with a hardware keyboard attached, you can usually see two notifications trigger the function in succession, one of which is discarded by the changes in the above PR. Following these steps to reproduce, you will see that only the discarded notification is sent, thus leaving the bottom content inset unchanged causing an overlap:

Steps:

0. Pre-req - Make changes to the example app (see code snippets below). Modify the AdvancedExampleViewController class as follows:
0.1 Disable the mock socket.
0.2 Add a UIAlertController actionSheet to one of the buttons
0.3 Launch the chat example in the simulator with the hardware keyboard connected and the software keyboard toggled off. (edited) 
1. Go to the 'advanced' view.
2. Type a multi-line attributed text (without going over the limit or deleting a line) and hit 'send'
3. Tap the button to present the actionSheet
4. Dismiss/cancel the actionSheet
Actual Result: Only one keyboard frame change notification is sent, and it happens to be the one ignored by the `guard !keyboardStartFrameInScreenCoords.isEmpty else {` condition in `MessagesViewController.handleKeyboardDidChangeState(_:)`. This means that the insets stay at 0. The effect in the UI is that the scrollview content is obfuscated by the inputAccessoryView.
Expected Result: Insets are re-calculated correctly.

Screenshots:

Scrolled up & held to show content:
![screen shot 2019-02-06 at 2 53 02 pm](https://user-images.githubusercontent.com/1952159/53183677-c1f0e780-35c9-11e9-867e-f01b4b94b28d.png)

The resting bottom content offset position after ending touches on the scrollview:
![screen shot 2019-02-06 at 2 52 11 pm](https://user-images.githubusercontent.com/1952159/53183748-e947b480-35c9-11e9-9260-7ad9dc44970f.png)

Some debugging screenshots if you want to dig through the values displayed:

![screen shot 2019-02-04 at 5 11 27 pm](https://user-images.githubusercontent.com/1952159/53184108-77239f80-35ca-11e9-9000-0e6fdb7ac2d6.png)
[...]

PR #1101, opened 2019-06-18
Fix: Content offset wrong after dismissing modal view controller

whoyawn:
- Implement the fix via https://github.com/MessageKit/MessageKit/issues/993#issuecomment-479390478
- Force `isFirstLayout` to be true whenever the view appears and call `viewDidLayoutSubviews()`

#993 

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone XR Simulator + Device

**iOS Version:** 12.2, 12.3

**Swift Version:** 5

**MessageKit Version:** 3.0.0-swift5

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -103,6 +103,8 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
     open override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         isMessagesControllerBeingDismissed = false
+        isFirstLayout = true
+        viewDidLayoutSubviews()
     }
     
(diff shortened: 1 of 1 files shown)


magneticrob:
This does fix the issue somewhat and is a better overall experience, but there is still a visual extra scroll when dismissing.

nathantannar4:
@whoyawn Thanks for taking the time to submit this. I added a comment on #993 as I'm not sure this is the best way to fix the issue. 

nathantannar4:
@whoyawn maybe does just `adjustScrollViewTopInset()` need to be called in `viewDidAppear`? Or something like that?

whoyawn:
I'll try that later this week and see how that goes

austinwright:
Closing, as this PR is quite stale. There was no followup on the suggestion above. Feel free to re-open if there is still an issue, or if you have a better solution.

PR #1102, opened 2019-06-19
Add support for audio seeking in AudioMessageCell

marcetcheverry:
—————————————————————————
This replaces `UIProgressView` with `UISlider` in `AudioMessageCell` giving an audio seek capability. Messages and WhatsApp both have this ability, as I assume most messaging applications do. It seems reasonable to assume that most users would want this functionality and this is why it is being added in `AudioMessageCell` as the default.

This also makes the audio cells wider and taller and slightly adjusts the layout to compensate for the slider's thumb image.

The example app has a commented out implementation example to disable seeking by hiding the `UISlider`'s `thumbImage`. This will return it to the current style of a progress view.

Ideally this would be a non breaking change if Swift supported optional methods in protocols and runtime checking of actual implementation, as we could hide the `UISlider`'s `thumbImage` if `didSeekAudio` is not implemented.

As it stands users will have to implement `didSeekAudio` to get the slider to work.

**Question: Should we add something like:**

`func supportsAudioSeeking(in cell: AudioMessageCell) -> Bool`

So the slider is automatically configured to hide its thumb image and the cell is laid out accordingly? Perhaps an option in `AudioItem`?

`protocol AudioItem {
    var isSeekable: Bool { get }
}`

IMG_3310

Any other comments?
-------------------

Where has this been tested?
---------------------------

**iOS Version:** 12.2

**Swift Version:** Swift 5

 👻


Sources/Views/Cells/MessageContentCell.swift
```diff
@@ -35,6 +35,7 @@ open class MessageContentCell: MessageCollectionViewCell {
         let containerView = MessageContainerView()
         containerView.clipsToBounds = true
         containerView.layer.masksToBounds = true
+        containerView.isUserInteractionEnabled = true
```
marcetcheverry, line 38: `UIImageView` `isUserInteractionEnabled` is off by default, but in this case it is a container view.

Sources/Views/MessagesCollectionView.swift
```diff
@@ -85,19 +85,28 @@ open class MessagesCollectionView: UICollectionView {
     
     private func setupGestureRecognizers() {
         let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleTapGesture(_:)))
-        tapGesture.delaysTouchesBegan = true
```
marcetcheverry, line 88: If there is a need to keep this please let me know, but failing on the existing `UICollectionView` recognizers should be enough. This avoids a delay in dragging the `UISlider`.

Sources/Views/MessagesCollectionView.swift
```diff
@@ -85,19 +85,28 @@ open class MessagesCollectionView: UICollectionView {
...
+        tapGesture.delegate = self
+
+        tapGesture.require(toFail: self.panGestureRecognizer)
+        if let pinchGestureRecognizer = self.pinchGestureRecognizer {
+            tapGesture.require(toFail: pinchGestureRecognizer)
```
marcetcheverry, line 92: I did not add `directionalPressGestureRecognizer` since it crashes on iOS 12, and it is disabled by default anyway.

farrukhaskari1:
is it going to be merged,

nathantannar4:
This looks great! I'm on holiday right now so I won't be able to take a detailed look until I'm back

Sources/Views/MessagesCollectionView.swift
nathantannar4: This could be an issue removing it

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -54,12 +58,22 @@ open class AudioMessageCell: MessageContentCell {
...
+
+        slider.addTarget(self,
+                         action: #selector(sliderValueChanged(_:)),
+                         for: .valueChanged)
+        return slider
     }()
-    
+
+    // MARK: - Interface Actions
+    @objc func sliderValueChanged(_ slider: UISlider) {
```
nathantannar4: Can you type this as private?

marcetcheverry: Good catch

Sources/Views/MessagesCollectionView.swift
marcetcheverry: Please feel free to do testing with and without it, I found that it created issues with the slider. Why was it added in the first place?

nathantannar4:
Did some testing, the seeking does not work unless the play button is pressed first. Do you think that would be a nice to have? For something like this we need to know what features are supported or if things are considered bugs

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -54,12 +58,22 @@ open class AudioMessageCell: MessageContentCell {
         return activityIndicatorView
     }()
 
-    public lazy var progressView: UIProgressView = {
-        let progressView = UIProgressView(progressViewStyle: .default)
-        progressView.progress = 0.0
-        return progressView
+    public lazy var slider: UISlider = {
```
nathantannar4: This would be a breaking change, so we need to handle that migration.

- I think `slider` should be renamed `progressControl`
- We should use `@available(*, unavailable, message: "Renamed to `progressControl` of type `UISlider`")` for `progressView`

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -54,12 +58,22 @@ open class AudioMessageCell: MessageContentCell {
...
+        let slider = UISlider()
+        let thumbImage = UIImage.messageKitImageWith(type: .sliderThumb)
+        slider.setThumbImage(thumbImage?.withRenderingMode(.alwaysTemplate), for: .normal)
+
```
nathantannar4: Please remove extra space

Sources/Views/Cells/AudioMessageCell.swift
```diff
@@ -68,27 +82,37 @@ open class AudioMessageCell: MessageContentCell {
...
     }
 
+    open override func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+        let touchPoint = gestureRecognizer.location(in: messageContainerView)
+
+        if slider.frame.contains(touchPoint) {
+            return false
+        }
+
+        return super.gestureRecognizerShouldBegin(gestureRecognizer)
```
nathantannar4, line 119: Why were all these gesture recognizer functions required to be overridden?

nathantannar4 (changes requested):
There are a lot of core changes here to the tap handling, and I dont think its needed. `UISlider` already handles a pan gesture so why add another one? It would be nice to have this self contained to the cell.

marcetcheverry:
Responding to the top level comment:

`UISlider already handles a pan gesture so why add another one?`.

I am not adding any tap gesture recognizers anywhere in the PR. All the work that is done with the delegate methods is to work around an architectural decision `MessageKit` made to use a global tap gesture recognizer on the entire collection view which then gets forwarded to the cells, instead of letting cells do their own thing. All of the code is necessary to let `UISlider` smoothly do its thing without getting intercepted or slowed down by `MessageKit`'s gesture recognizers. Feel free to comment out some of the code and see how it works without it.

Sources/Views/Cells/AudioMessageCell.swift
marcetcheverry: Fixed

Sources/Views/Cells/AudioMessageCell.swift
marcetcheverry: Because in `MessagesCollectionView.swift` `func setupGestureRecognizers` there is a global tap gesture recognizer that we must punch through so `UISlider` gets clean touches.

Sources/Views/Cells/AudioMessageCell.swift
marcetcheverry: That is fine, should we include a `fatalError("Renamed to progressControl of type UISlider")` in the actual implementation, or keep the old one?

marcetcheverry:
> Did some testing, the seeking does not work unless the play button is pressed first. Do you think that would be a nice to have? 

This is up to the implementer, `BasicAudioController` handles disabling and enabling `isUserInteractionEnabled`. This is an example implementation, users can do whatever they want and `MessageKit` does not have any restrictions. The reason it was done this way is that `BasicAudioController` does not actually create a `AVAudioPlayer` until `playSound` is called, so there won't be anything to seek before playback has begun.

Sources/Views/Cells/AudioMessageCell.swift
marcetcheverry: I found that I wasn't getting clean sliding with `UISlider` if these were not overridden. 

Also otherwise just simply tapping the `UISlider` could count as a message tap/long press.

Finally it is good form to just let UISlider do its own thing and not be affected by any other gesture recognizers the user may place throughout the view hierarchy.

SD10:
I requested @moldovaniosif on this one 👀

marcetcheverry:
Moving to private fork.

ElrB:
What happened to this? Is audio seeking supported? I just think this is a necessary feature as a user would not like or want to wait for the entire audio to finish playing and restart just so he/she can listen again to a certain part of an audio. What if the audio is 10 minutes+ long, and user has to wait for all that 10 minutes to finish. Thanks.

Issue #1106, opened 2019-06-21
cellTopLabelAlignment not set in configure(attributes:)


magneticrob:
- What version of MessageKit are you using? 
3.1.0

- What version of iOS are you running on? 
iOS 12

- What version of Swift are you running on?
Swift 5

- What device(s) are you testing on? Are these simulators?
iPhone X 12.2, Sim

- Is the issue you're experiencing reproducable in the example app?
Yep

In [MessageSizeCalculator.swift](https://github.com/MessageKit/MessageKit/blob/master/Sources/Layout/MessageSizeCalculator.swift) there's a method where the attributes are configured. It looks like this:

```
attributes.avatarSize = avatarSize(for: message)
attributes.avatarPosition = avatarPosition(for: message)
attributes.avatarLeadingTrailingPadding = avatarLeadingTrailingPadding

attributes.messageContainerPadding = messageContainerPadding(for: message)
attributes.messageContainerSize = messageContainerSize(for: message)
attributes.cellTopLabelSize = cellTopLabelSize(for: message, at: indexPath)
attributes.cellBottomLabelSize = cellBottomLabelSize(for: message, at: indexPath)
attributes.cellBottomLabelAlignment = cellBottomLabelAlignment(for: message)
attributes.messageTopLabelSize = messageTopLabelSize(for: message, at: indexPath)
attributes.messageTopLabelAlignment = messageTopLabelAlignment(for: message)

attributes.messageBottomLabelAlignment = messageBottomLabelAlignment(for: message)
attributes.messageBottomLabelSize = messageBottomLabelSize(for: message, at: indexPath)

attributes.accessoryViewSize = accessoryViewSize(for: message)
attributes.accessoryViewPadding = accessoryViewPadding(for: message)
attributes.accessoryViewPosition = accessoryViewPosition(for: message)
```

How come you're setting everything except `cellTopLabelAlignment`? Had me pulling my hair out for a minute there. Want me to make a pull request?

It'd just be the below:

```
attributes.cellTopLabelAlignment = cellTopLabelAlignment(for: message)
```

nathantannar4:
Yes that looks like it was missed! PR would be great!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
This was resolved by #1113

PR #1113, opened 2019-06-26
Add missing cellTopLabelAlignment to MessageSizeCalculator

marcetcheverry:
See https://github.com/MessageKit/MessageKit/issues/1106


nathantannar4:
Thanks for taking the time to make another PR @marcetcheverry 😁

When I looked at this issue I noticed that `MessageContentCell.layoutCellTopLabel` also needs updates for it to even use the assigned layout attributes. 

Could that be added to this? Needs to look similar to `layoutCellBottomLabel`

mdescalzo:
@nathantannar4 I saw that in my poking around as well.  I'm setting up a PR for it now: PR #1117

nathantannar4:
@mdescalzo I think this should all be done in one PR, as it's for the same issue. I appreciate you taking the time to do so none the less. 

Issue #1107, opened 2019-06-24
The Cell structure diagram does not match behavior


mdescalzo:
- What version of MessageKit are you using?
3.0.0

- What version of iOS are you running on?
12.2

- What version of Swift are you running on?
4.2

- What device(s) are you testing on? Are these simulators?
Simulator

- Is the issue you're experiencing reproducable in the example app?
Yes

The frames of the `cellTopLabel`,` cellBottomLabel`, `messageTopLabel`, and `messageBottomLabel` extend the full width of the cell.  They are not constrained between the avatar and accessory views as represented in the CellStructure diagram in the README file.  The messageContentView is so constrained.  This is more representative of the frame borders:
![CellStructure copy](https://user-images.githubusercontent.com/1128650/60054285-45fdcd80-9697-11e9-8508-ca47c0ea839a.png)

You can see this with in the sample app by making setting the `layer.borderColor` and `layer.borderWidth` properties on the cell's subviews.

I suggest having the two message labels left and right contraints match those of `messageContentView` and the cell labels match the cell's left and right constraint.

nathantannar4:
I would have to double check but I think the current layout diagram is accurate. 

The actual layout will vary depending on how the size calculators are configured. For example, the positioning of the avatar view will change the layout. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mdescalzo:
I've played with the avatar position a bit (not exhaustive) and it has had no effect on the left or right bounds of the messageTopLabel or messageBottomLabel.  I'll mess with it some more tomorrow and get you some screenshots.

nathantannar4:
Changing the padding of the labels will effect the frame shown in the image 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mdescalzo:
Neither `MessagesCollectionViewLayoutAttributes` or `MessagesCollectionViewFlowLayout` contain padding properties for the top/bottom message labels.  There are message container padding properties, but these do not effect the frames of the top/bottom cell/message labels.

[...]

PR #1118, opened 2019-06-27
Add individual padding attributes for the cell and message labels

mdescalzo:
- Adds `cellTopLabelPadding`, `cellBottomLabelPadding`, `messageTopLabelPadding`, & `messageBottomLabelPadding` attributes to `MessagesCollectionViewLayoutAttributes`
- Adds the above attributes to layout calculations in `MessageSizeCalculator`
- Adds `cellTopLabelPadding`, `cellBottomLabelPadding`, `messageTopLabelPadding`, & `messageBottomLabelPadding` methods to the `MessageLayoutDelegate` protocol with defaults which return `.zero` insets.

This allows manipulation of the label frames which can be desirable when adding borders or backgrounds to these labels.

The default `.zero` insets means this should not break/change existing layout behavior.

This is, in part, to address issues discussed in #1107 

nope

Any other comments?
-------------------
nope

Where has this been tested?
---------------------------
**Devices/Simulators:** 
iPad sim, iPhone Xs, iPad mini

**iOS Version:** …
12.2

**Swift Version:** …
4.2 & 5

**MessageKit Version:**
3.0.0


Sources/Protocols/MessagesLayoutDelegate.swift
```diff
@@ -145,18 +185,33 @@ public extension MessagesLayoutDelegate {
         return 0
     }
     
+    func cellTopLabelPadding(for message: MessageType, at indexPath: IndexPath) -> UIEdgeInsets {
```
nathantannar4, line 188: We cannot set these properties through a delegate method. See #1107

Sources/Layout/MessagesCollectionViewLayoutAttributes.swift
```diff
@@ -40,15 +40,19 @@ open class MessagesCollectionViewLayoutAttributes: UICollectionViewLayoutAttribu
 
     public var cellTopLabelAlignment = LabelAlignment(textAlignment: .center, textInsets: .zero)
     public var cellTopLabelSize: CGSize = .zero
+    public var cellTopLabelPadding: UIEdgeInsets = .zero
```
nathantannar4, line 43: This is where we want the padding properties, so its confusing how there is a mixture of this and delegates. We should also have a convenience setter in `MessagesCollectionViewFlowLayout` to set this as we do with the other properties

nathantannar4 (changes requested):
I'm happy to add padding properties, but it should follow the same convention we have for the other layout attributes.

mdescalzo:
I'm pulling this back for a rework.  Thank you very much for the input.

nathantannar4:
Look forward to the new version!

PR #1136, opened 2019-07-10
Fix text clipping caused by MessageSizeCalculator when textInsets are zero in MessageLabel

marcetcheverry:
I don't know which branch to target anymore, 3.0.0-swift5 or development, please clarify.

If you set `MessageLabel` insets to `.zero`, text size calculation will not work because it is often off by 1 point in large wrapped strings. Adding `usesDeviceMetrics` to use image bounds fixes this easily.

Notice in the before and after picture how "END" does not appear in the first case. I turned off bubble style to make it clearer (bubble style has no effect on this issue).

Before:
before

After:
after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Lazily initialize the MessageInputBar on MessagesViewController. [#1092](https://github.com/MessageKit/MessageKit/pull/1092) by [@marcetcheverry](https://github.com/marcetcheverry)
 
+### Changed
+
+- Fix text clipping caused by MessageSizeCalculator when textInsets are zero in MessageLabel. [#1136](https://github.com/MessageKit/MessageKit/pull/1136) by [@marcetcheverry](https://github.com/marcetcheverry)
+
 ## 3.0.0
 
 ### Dependency Changes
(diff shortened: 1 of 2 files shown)


marcetcheverry:
Use this patch to test this PR.

[Test1136.patch.txt](https://github.com/MessageKit/MessageKit/files/3379535/Test1136.patch.txt)

This issue was found while designing a complex cell with multiple `MessageLabel`s that had `textInsets`  of `.zero`.

marcetcheverry:
Note that using usesDeviceMetrics introduces a regression with single line strings, cutting them shorter than they should be. It seems MessageLabel draws about 1.66 points offset from UILabel, which causes issues with sizing when using the bounding methods for size calculation.

The other approach is to do what #824 did and just avoid this method altogether and use NSLayoutManager.

marcetcheverry:
This bug is related to #897, he tried fixing it by overriding `intrinsicContentSize` to match the size calculator rather than use `UILabel`'s

nathantannar4:
Funny how such a small change can result in such a difference 😅

marcetcheverry:
Stand by, I will attempt to provide a fix that fixes both #897 and this.

marcetcheverry:
Moving to private fork.

hyouuu:
Hi @marcetcheverry found this PR from https://github.com/MessageKit/MessageKit/pull/824 - would you kindly provide what your final solution was to this issue? Thanks!

PR #1137, opened 2019-07-10
Fix implementation of sizeThatFits and improve implementation of intrinsicContentSize in MessageLabel

marcetcheverry:
Overriding `intrinsicContentSize` is a not a proper implementation for `UILabel` subclasses that want to add insets.

Instead the `textRect` methods of `UILabel` should be overridden, since they are called by both `intrinsicContentSize` and the older `sizeThatFits`.

This PR also makes sure to call `invalidateIntrinsicContentSize` when `textInsets` is changed.

Purpose of this PR: use `MessageLabel` inside of a `UIStackView` and to improve the overall correctness of the implementation.

Note for third parties: you will need to set a `preferredMaxLayoutWidth` if you want to use auto sizing, otherwise `sizeToFit` will size to a single line regardless of `numberOfLines`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Fix implementation of sizeThatFits and improve implementation of intrinsicContentSize in MessageLabel. [#1137](https://github.com/MessageKit/MessageKit/pull/1137) by [@marcetcheverry](https://github.com/marcetcheverry)
 
-
 ## 3.0.0
 
 ### Dependency Changes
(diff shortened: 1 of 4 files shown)


marcetcheverry:
Note the care taken care with `CGFLoat`/`Float.greatestFiniteMagnitude`. I don't know of a better implementation of this, and yes I am aware of float comparison pitfalls, but I am unsure if they apply to these sentinel values. The case for `Float.greatestFiniteMaginute` is that somewhere in `UILabel` (developed with the original iPhone OS) the framework is calling with 32 bit float max and not using `CGFLoat` to do constraint based layout.

marcetcheverry:
Note that using `usesDeviceMetrics` introduces a regression with single line strings, cutting them shorter than they should be. It seems `MessageLabel` draws about 1.66 points offset from UILabel, which causes issues with sizing when using the `bounding` methods for size calculation.

The other approach is to do what #824 did and just avoid this method altogether and use NSLayoutManager.

marcetcheverry:
Moving to private fork.

Issue #1149, opened 2019-07-26
Proposal: Move methods of MessageLabelDelegate to MessageCellDelegate so that selected cells can be acquired


hmuronaka:
In the current implementation of MessageLabelDelegate, when the messageLabel is tapped, the URL and Mention can be obtained, but the tapped cell is unknown.

Proposal: I want to move the method of MessageLabelDelegate to MessageCellDelegate so that selected cells can also be acquired.

We will send a pullreq, please refer to it if you like.

- What version of MessageKit are you using?
3.0.0

- What version of iOS are you running on?
iOS 12.1
- What version of Swift are you running on?
Swift 5

- What device(s) are you testing on? Are these simulators?
iPhone XR(simulator),  iPhone7Plus

PR #1150, opened 2019-07-26
update Proposal: Move methods of MessageLabelDelegate to MessageCellDelegate so that selected cells can be acquired

hmuronaka:
Please check issue #1149 I wrote the detail.


hyouuu:
`MessageCellDelegate` is already inheriting from `MessageLabelDelegate` - shouldn't you be able to directly use `MessageCellDelegate`?

austinwright:
After some offline discussion I think this can be done in a non-breaking manner.

Kaspik (changes requested):
As mentioned by Austin.

Kaspik:
Closing for no further activity as original issue was auto-closed as well.

PR #1180, opened 2019-09-15
Resolve issue with scrolling to bottom

grangej:

The content size is 0 when reading before performBatchUpdates is called
The content size is correct inside performBatchUpdates
Calling reloadData() before doesn't help.

Where has this been tested?
---------------------------
**Devices/Simulators:**

Tested on iOS 13 

**Swift Version:** …

Swift 5.1

**MessageKit Version:** …

3.0

### DIFF
--- a/Sources/Views/MessagesCollectionView.swift
+++ b/Sources/Views/MessagesCollectionView.swift
@@ -101,9 +101,10 @@ open class MessagesCollectionView: UICollectionView {
     }
 
     public func scrollToBottom(animated: Bool = false) {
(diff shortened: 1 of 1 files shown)


hyouuu:
Please point this to development and we can get it merged :)

hyouuu:
Actually I could change base - please resolve conflicts

hyouuu:
Looks like the major part of your change is already in `development`, and I've incorporated your `[weak self]` usage in https://github.com/MessageKit/MessageKit/pull/1247 - thanks for contributing!

Issue #807, opened 2018-08-11
How to set unread message status?


evgzor:
Could u please explain how to set read/unread message?

- version of MessageKit are you using : 1.0.0
- version of iOS are you running on: ios 11
-  version of Swift are you running on: Swift 4
- device(s) are you testing on? Are these simulators? : iOS simulator
- Is the issue you're experiencing reproducable in the example app?

nathantannar4:
Use the footer text of a MessageCell (which we show how to set in the example) or implement a custom cell. When to set a message as seen/unseen is the trickier part but that you would have to implement client side. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

SD10:
@evgzor In my app we use the footer text of the cell as @nathantannar4 has mentioned above. If we receive a notification that the message status has changed, we update the label. This is a very client specific issue

tobitech:
@SD10 Please how do you add status to the MessageType... is there a way to create a Message model that would have "status" that can be used by the MessagesViewController datasource?

nathantannar4:
You can use .custom

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #1188, opened 2019-09-30
Adding property status to MessageType protocol

Vlada31R:
There is necessity to track message status. In future cases "sending", "error" could be added to MessageStatus model.

### DIFF
--- a/Sources/Models/MessageStatus.swift
+++ b/Sources/Models/MessageStatus.swift
@@ -1,0 +1,15 @@
+//
+//  MessageStatus.swift
+//  MessageKit
+//
+//  Created by Vlada Radchenko on 7/24/19.
+//  Copyright © 2019 MessageKit. All rights reserved.
+//
+
+import Foundation
+
+public enum MessageStatus: String, Codable {
+
+    case delivered
+    case read
+}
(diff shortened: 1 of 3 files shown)


Sources/Models/MessageStatus.swift
```diff
@@ -0,0 +1,15 @@
...
+//  MessageKit
+//
+//  Created by Vlada Radchenko on 7/24/19.
+//  Copyright © 2019 MessageKit. All rights reserved.
+//
+
+import Foundation
+
+public enum MessageStatus: String, Codable {
+
```
MacMeDan, line 12: Shouldn't there be an `Error` or `Failed` status as well?

MacMeDan (commented):
Add a failed state.

Sources/Models/MessageStatus.swift
MacMeDan: You also need to update the `MockMessage` for tests to pass.

kinoroy: Should be of note that this would be a breaking change to the library, it would cause everybody who implements the MessageModel protocol (so everybody) to have their implementation break. We should carefully consider if it's worth it to break everybody's code for something that MessageKit would not even make use of internally. Seems like something that could just be specified in a concrete implementation on the user's side as opposed to requiring every consumer of the library to implement it.

MacMeDan: Ya I did not take that into consideration. This change wound need to be apart of a major release due to the breaking nature of it.

hyouuu: Could we add an `initial` status, and extend `MessageType` protocol to return initial so it works with existing clients?

kinoroy: That could work but IMHO, anything that's not used by the library itself probably doesn't need to be in the `MessageType` protocol. Since it's just a protocol, clients of the library already have the flexibility to add a `MessageStatus` field to their message struct/class if they desire.
````swift
struct Message: MessageType {
    // ... MessageType implementation
    let status: MessageStatus
}
````
This issue has come up before (#807) and it seems like Nathan and Steven believed this to be a more client side issue.

hyouuu: Makes sense. @Vlada31R what do you think? We would probably just close this and approach as kinoroy recommended above

alexmomotiuk:
I think it's no necessary for now, especially if it's not using by MessageKit

Kaspik:
Should we close this @alexmomotiuk / @austinwright ?

PR #1218, opened 2019-11-20
Fixed requiredInitialScrollViewBottomInset not returning the correct value when inputAccessoryView is nil

aabosh:
When disabling MessageKit's input bar by returning `nil` for `inputAccessoryView` in `MessagesViewController`, this causes the method `requiredInitialScrollViewBottomInset()` to always return `0`. This shouldn't the case as it causes the property `additionalBottomInset` to be completely ignored. I noticed this issue as I currently handle adding `InputBarAccessoryView` into view myself due to wanting tab bar support.

Not that I'm aware of.

Nope.

Any other comments?
-------------------
Nope.

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone XS, iPad Air 2

**iOS Version:** 12.1, 13.1

**Swift Version:** 4.2

**MessageKit Version:** 3.0.0


marcetcheverry:
This is a good catch, in my branch where I removed InputBarAccessoryView I also did something similar.

sgupta3 (approved):
This is a great catch. Thanks for making the change. Can we have this merged in?

austinwright:
@aabosh, I will review this more in-depth while awaiting merge privileges. In the meantime, this is what I have done in our specific case, not sure if it is the same issue you are experiencing but perhaps it may help without needing to point to a custom fork:

```swift
    override func viewSafeAreaInsetsDidChange() {
        NSLog("\(type(of: self)).\(#function) -> Checking if collectionView's bottom inset needs to be adjusted.")
        guard let inputAccessoryView = inputAccessoryView else {
            NSLog("\(type(of: self)).\(#function) -> inputAccessoryView was nil.")
            return
        }
        let requiredBottomInset = max(0, inputAccessoryView.frame.height - view.safeAreaInsets.bottom)
        if messagesCollectionView.contentInset.bottom != requiredBottomInset {
            NSLog("\(type(of: self)).\(#function) -> Updating bottom inset.")
            messagesCollectionView.contentInset.bottom = requiredBottomInset
            messagesCollectionView.scrollIndicatorInsets.bottom = requiredBottomInset
        }
    }
```

In our case we are hiding a tab bar and also sometimes making other UI changes that require adjusting the safe area insets. In your case above you may want to incorporate `additionalBottomInset` into your implementation. You may also need to override `viewDidLayoutSubviews` rather than `viewSafeAreaInsetsDidChange` depending on your specific situation.

Sources/Controllers/MessagesViewController+Keyboard.swift
```diff
@@ -122,8 +122,8 @@ internal extension MessagesViewController {
     }
 
     internal func requiredInitialScrollViewBottomInset() -> CGFloat {
-        guard let inputAccessoryView = inputAccessoryView else { return 0 }
-        return max(0, inputAccessoryView.frame.height + additionalBottomInset - automaticallyAddedBottomInset)
+        let inputAccessoryViewHeight = inputAccessoryView?.frame.height ?? 0
```
austinwright, line 130: Personally I'm not a fan of optional chaining anymore, it can and has caused soft-failures to slip past unnoticed. Usually in a guard let there is a log statement before returning so that we can be notified that the execution went down an 'unhappy path'. However, I'm not aware of any logging standards in the MessageKit framework thus far. What if we did it like this:

```swift
    internal func requiredInitialScrollViewBottomInset() -> CGFloat {
        var adjustedBottomInset = additionalBottomInset - automaticallyAddedBottomInset
        if let inputAccessoryView = inputAccessoryView, inputAccessoryView.frame.height > 0 {
            adjustedBottomInset += inputAccessoryView.frame.height
        }
        return max(0, adjustedBottomInset)
    }
```

This leaves room to log something within an 'else' statement in the future. Execution can continue if the conditional unwrapping fails, so we go with `if let` instead of `guard`. This also prevents us from having near-duplicate code in what would have been the guard's early return statement.

austinwright (approved):
These changes appear to be fine as-is, though I did leave some thoughts about how to avoid the optional chaining in favor of more maintainable (thought also more verbose) code.

I am concerned that this covers up what might be a larger issue with making UI calculations before, during, or in-between UIKit lifecycle events which may cause problems. See my comment about needing to recalculate the insets when the safe area / additional safe area change as the view is loaded/appearing.

hyouuu:
Thanks for reviewing! Due to others' similar needs this seems worthy merging in - @aabosh could you pull from develop to resolve conflicts?

aabosh:
@hyouuu Sure! That's been done now.

hyouuu:
Great thanks! Will merge in after ci

aabosh:
@hyouuu Looks like tests unrelated to this PR are failing:
```
AvatarViewTests
    ✓ testGetImage (1.220 seconds)
    ✗ testInitialsOnly, XCTAssertEqual failed: ("Optional(<UIDynamicModifiedColor: 0x7f8f34089970; contrast = normal, baseColor = <UIDynamicSystemColor: 0x7f8f340e54a0; name = systemGrayColor>>)") is not equal to ("Optional(UIExtendedGrayColorSpace 0.5 1)")
    ✗ testNoParams, XCTAssertEqual failed: ("Optional(<UIDynamicModifiedColor: 0x7f8f31d3f840; contrast = normal, baseColor = <UIDynamicSystemColor: 0x7f8f340e54a0; name = systemGrayColor>>)") is not equal to ("Optional(UIExtendedGrayColorSpace 0.5 1)")
    ✓ testRoundedCorners (0.001 seconds)
    ✗ testSetBackground, XCTAssertEqual failed: ("Optional(<UIDynamicModifiedColor: 0x7f8f342453f0; contrast = normal, baseColor = <UIDynamicSystemColor: 0x7f8f340e54a0; name = systemGrayColor>>)") is not equal to ("Optional(UIExtendedGrayColorSpace 0.5 1)")
    ✗ testWithImage, XCTAssertEqual failed: ("Optional(<UIDynamicModifiedColor: 0x7f8f340da530; contrast = normal, baseColor = <UIDynamicSystemColor: 0x7f8f340e54a0; name = systemGrayColor>>)") is not equal to ("Optional(UIExtendedGrayColorSpace 0.5 1)")
```

hyouuu:
CI failures are related to the color update - I'll update the tests later.

SD10:

PR #1249, opened 2020-02-19
Add messageStyleInsets

hyouuu:
Add messageStyleInsets delegate method to adjust messageStyle insets - currently with bubble style the left & right padding from text is quite large.

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 11 Pro

**iOS Version:** 13.3.1

**Swift Version:** 5.1

**MessageKit Version:** develop branch


Sources/Protocols/MessagesDisplayDelegate.swift
```diff
@@ -30,6 +30,17 @@ public protocol MessagesDisplayDelegate: AnyObject {
...
 
+    /// Specifies the `UIEdgeInsets` to be used for a `MessageContainerView`.
+    ///
+    /// - Parameters:
+    ///   - message: The `MessageType` that will be displayed by this cell.
+    ///   - indexPath: The `IndexPath` of the cell.
+    ///   - messagesCollectionView: The `MessagesCollectionView` in which this cell will be displayed.
+    ///
+    /// - Note:
+    ///   The default value returned by this method is `MessageStyle.bubble`.
```
alexmomotiuk: Typo, should be 

> The default value returned by this method is `UIEdgeInsets.zero`.

hyouuu: Good catch on the copy-pasta :p

alexmomotiuk (approved):

SD10:
Thanks for the PR 💯 

1. Can this not be achieved using the following API?

https://github.com/MessageKit/MessageKit/blob/5adb798612db8e430ab75a6240b2c5399b2dd6a5/Sources/Layout/MessageSizeCalculator.swift#L43-L44

2. Won't this affect cell sizing calculations?

hyouuu:
@SD10 thanks for checking!
1. No - it calculates the padding between the actual text boundary to left & right to e.g. avatar, but this is the bubble image behind it, which always had constant "padding" from the text boundary.
2. Since the messageTopLabel, cellTopLabel etc are positioned to the text boundary, this shouldn't affect sizing calculation most of the time as this is mostly cosmetic. I could be wrong though.

Sources/Controllers/MessagesViewController.swift
```diff
@@ -47,7 +47,7 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
     /// bottom whenever the `InputTextView` begins editing.
     ///
     /// The default value of this property is `false`.
-    /// NOTE: This calls scrollToBottome where as the above flag calls  scrollToLastItem - check methods for differences
+    /// NOTE: This calls scrollToBottom where as the above flag calls  scrollToLastItem - check methods for differences
```
austinwright: Can you update this comment to what I commented on your previous PR? (Fix extra space, other wording)

Sources/Views/MessageContainerView.swift
```diff
@@ -30,6 +30,8 @@ open class MessageContainerView: UIImageView {
 
     private let imageMask = UIImageView()
 
+    open var styleInsets: UIEdgeInsets = .zero
```
austinwright, line 33: Is this intended to affect the intrinsic content size of the Container, or just offset the image within the container? (I'm assuming it is the latter, just confirming)

Sources/Controllers/MessagesViewController.swift
hyouuu: Updated

Sources/Views/MessageContainerView.swift
hyouuu: Yes the latter - see here it's just insetting the image mask: https://github.com/MessageKit/MessageKit/pull/1249/files#diff-bac966c08f96c77572148721a36fd0f0R54

austinwright (changes requested):
I have similar concerns as @SD10's comment related to content size calculations - Can you test a few things out and report back?:

1. Set all of the other views in the container to `.zero`. Try different variations of un-balanced offsets, different length text messages (single line, multi-line), Different message types. 
2. Try using negative insets.
3. Set all of the other views to non-zero and re-test # 1 & 2.
4. Rotate the phone and/or go to settings and turn down/up the accessibility font size (or anything else that might dynamically change the frame size).

Sources/Views/MessageContainerView.swift
austinwright: Can you post a screenshot of what this looks like in use? We might also want to add some additional function documentation explaining how to use this and how it differs from `incoming/outgoingMessagePadding`

Kaspik:
Let's wait with this after 3.1.0 release?

Sources/Controllers/MessagesViewController.swift
```diff
@@ -40,14 +40,14 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource {
     /// last item whenever the `InputTextView` begins editing.
     ///
     /// The default value of this property is `false`.
-    /// NOTE: This calls scrollToLastItem where as the below flag calls scrollToBottom - check methods for differences
+    /// NOTE: This is related to `scrollToLastItem` whereas the below flag is related to `scrollToBottom `- check each function for differences
```
austinwright, line 43: Can you apply just these comment fixes to develop in a separate PR? I'd like to have them in place now rather than later.

hyouuu:
Will test out and update soon!

Sources/Views/MessageContainerView.swift
jakubwarchol: @austinwright just as a side note when I was starting it was confusing to me the amount of padding and insets and how they relate to each other. I think we could use a better documentation. Perhaps I can help with this?

hyouuu:
@austinwright here are the examples:

Base example:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-25 at 13 37 14

With setMessageIncomingMessagePadding(.zero) - you can see it only reduces padding between message & avatar:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-25 at 13 38 15

With slight reduced top, left & right inset for the bubble - messageStyleInsets UIEdgeInsets(top: 1, left: 2, bottom: 0, right: 2):
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-25 at 13 39 46

Same messageStyleInsets as above & zero incoming padding:
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-25 at 13 40 58

Sources/Views/MessageContainerView.swift
Kaspik: Sure you can, we always welcome PRs about documentation, tests and this kind of stuff! :)

Kaspik (approved):
Not sure if changes to `MessagesViewController` are expected? I think they shouldn't be here?

abdulrehman73331:
@hyouuu  hi bro you have posted the example pictures above in which I can see time is showing against each message can you please tell me how you are doing this? I tried in messageBottomLabelAttributedText function but it always shows at the end of last message of the sender only.

hyouuu:
@abdulrehman73331 I don't think I did anything about time display here - try the example project and it should show it automatically

Kaspik:
We are getting rid of develoment branch, thus this PR was autoclosed. Please open it again against master. :)

PR #1253, opened 2020-02-21
Update InputBarAccessoryView to 4.3.2, edit Cartfile to allow < v5.0.0

kinoroy:
- Removes the fixed version requirement of "4.3.0" for InputBarAccessoryView, and allows versions up to but not including version 5.0.0.
- Updates the dependency to the latest version (4.3.2) 

-----------------------------------------
No

Where has this been tested?
---------------------------
**Devices/Simulators:** … Simulator: iPhone 11 Pro

**iOS Version:** … 13.2

**Swift Version:** … 5.0

**MessageKit Version:** … N/A (unreleased)

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "nathantannar4/InputBarAccessoryView" "4.3.0"
+github "nathantannar4/InputBarAccessoryView" ~> 4.3.0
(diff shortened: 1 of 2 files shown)


kinoroy:
@austinwright It seems like for Carthage, `~> 4.3.0` allows up to 5.0.0. I think this is different behaviour than Cocoapods 
https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md

> Compatibility is determined according to Semantic Versioning. This means that any version greater than or equal to 1.5.1, but less than 2.0, will be considered “compatible” with 1.5.1.

Kaspik (approved):

Kaspik (approved):
Re-Approve with note: Approved if we are okay with versioning as described and we don't want to be more strict. :)

PR #1256, opened 2020-02-26
Update Circle CI PR tests to include cache for pods

Kaspik:
Change example build to use cache for pods, without fetching specs.

No

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
 #          command: curl -sS https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash
       - run:
           name: Update Pods
-              command: cd Example && pod install
+          command: cd Example && pod install
       - save_cache:
           name: Save Pods cache
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,19 +1,13 @@
-# .circleci/config.yml
-
-# Specify the config version - version 2 is latest.
+# Specify the config version
 version: 2.1
 
-# Define the jobs for the current project.
+# Define the jobs for the current project
 jobs:
-  build-and-test:
-
+  build_tests:
+    macos:
+      xcode: 11.3.1
     environment:
       - DESTINATION: "platform=iOS Simulator,name=iPhone 11 Pro"
-
-    # Specify the Xcode version to use.
-    macos:
-      xcode: "11.3.1"
-
     # Define the steps required to build the project.
     steps:
       # Get the code from the VCS provider.
@@ -71,6 +65,6 @@ jobs:
       #    destination: scan-logs
 
 workflows:
-  build-and-deploy:
-    jobs:
-      - build-and-test
+    build_and_tests:
+        jobs:
+            - build_tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,25 +30,22 @@ jobs:
       - run:
           name: Build and run tests
           command: xcodebuild test -project MessageKit.xcodeproj -scheme MessageKitTests -destination "$DESTINATION" CODE_SIGNING_REQUIRED=NO | xcpretty -c
-      - run:
-          name: Open Example
-          command: cd Example
       - restore_cache:
           name: Restore Pods cache
           keys:
-              - v1-pods-cache-{{ arch }}-{{ checksum "Podfile.lock" }}
+              - v1-pods-cache-{{ arch }}-{{ checksum "Example/Podfile.lock" }}
               - v1-pods-cache-{{ arch }}-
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1,22 +1,15 @@
...
-version: 2
-
-# Define the jobs for the current project.
+# Define the jobs for the current project
 jobs:
-  build-and-test:
-
+  build_tests:
+    macos:
+      xcode: 11.3.1
```
austinwright, line 8: Curious: can this be set using an opportunistic operator so that we get Xcode bugfix versions automatically?

.circleci/config.yml
```diff
@@ -25,18 +18,31 @@ jobs:
       - run:
           name: Bootstrap Carthage
           command: carthage bootstrap --platform ios
+      - run:
+          name: Pre-start simulator
+          command: xcrun instruments -w "iPhone 11 Pro (13.3) [" || true
```
austinwright, line 24: Can you add a comment above this line breaking down what this does, since the next person to maintain this config might not want to dig through man pages for `xcrun` and `instruments`. Something like:

> `xcrun` - runs a tool available within xcode from the command line
> `instruments -w`- use the instruments command-line tool to start up the simulator
> `"iPhone 11 Pro (13.3) ["` - The Device ID to simulate
> `|| true` - Continue executing the job's run commands if this command does not exit successfully

Questions: What is the trailing `[` in the device ID literal? Where do you find the list of valid device IDs to choose from?

austinwright, line 24: Can you add a comment above this line breaking down what this does, since the next person to maintain this config might not want to dig through man pages for `xcrun` and `instruments`. Something like:

> `xcrun` - runs a tool available within xcode from the command line
> `instruments -w`- use the instruments command-line tool to start up the simulator
> `"iPhone 11 Pro (13.3) ["` - The Device ID to simulate
> `|| true` - Continue executing the job's run commands if this command does not exit successfully

Questions: What is the trailing `[` in the device ID literal? Where do you find the list of valid device IDs to choose from?

.circleci/config.yml
```diff
@@ -25,18 +18,31 @@ jobs:
...
           name: Build and run tests
           command: xcodebuild test -project MessageKit.xcodeproj -scheme MessageKitTests -destination "$DESTINATION" CODE_SIGNING_REQUIRED=NO | xcpretty -c
-      - run: 
-          name: Fetch CocoaPods Specs
-          command: curl -sS https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash
+      - restore_cache:
+          name: Restore Pods cache
+          keys:
+              - v1-pods-cache-{{ arch }}-{{ checksum "Example/Podfile.lock" }}
+              - v1-pods-cache-{{ arch }}-
```
austinwright: should we add `{{ .Branch }}` in between these two?

.circleci/config.yml
```diff
@@ -62,7 +68,6 @@ jobs:
       #    destination: scan-logs
 
 workflows:
-  version: 2
-  build-and-deploy:
-    jobs:
-      - build-and-test
+    build_and_tests:
```
austinwright: Bit of a nitpick - and I could be wrong - but should this be indented 2 spaces instead of 4?

MessageKit.podspec
```diff
@@ -11,17 +11,11 @@ Pod::Spec.new do |s|
    s.source = { :git => 'https://github.com/MessageKit/MessageKit.git', :tag => s.version }
    s.source_files = 'Sources/**/*.swift'
 
-   s.pod_target_xcconfig = {
```
austinwright, line 14: Can you speak to these two removals and why they are not needed anymore?

austinwright (changes requested):

austinwright:
~Also, PR needs a change-log entry (right? 😅)~
Changelog entries not needed for CI changes which don't directly affect the Framework end-user

.circleci/config.yml
Kaspik: It's recommended to keep it specific as you need to update the simulator usually too.

.circleci/config.yml
Kaspik: Added info to the docu.

.circleci/config.yml
Kaspik: Added.

MessageKit.podspec
Kaspik: `s.requires_arc` is true by default - not necessary then.
`s.pod_target_xcconfig` SWIFT_VERSION` is legacy thing that was replaced by `s.swift_version` that we already have. More updates coming! :)

austinwright (approved):

PR #1261, opened 2020-02-28
Drop iOS 9 and iOS 10, update example project

Kaspik:
- drop support for iOS 9 and iOS 10, update layout to use safe area layout guide
- update copyright
- update Example project to support iOS 11+ too
- cleanup legacy `.swift-version` file
- update podspec to support iOS 11+ and to to count with 3.1.0 release already

Any other comments?
-------------------
This is breaking but expected change, [Vision](https://github.com/MessageKit/MessageKit/blob/master/VISION.md) says this library will support at least 3 versions of iOS - which is still true. Using safeAreaLayoutGuide is requirement.

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 11 Pro

**iOS Version:** iOS 13.3

**Swift Version:** 5.1

**MessageKit Version:** 3.1.0 - `development`


victormihaita:
Hi. When will this be ready for release?

Kaspik:
Hey, version 3.1.0 is going to be finally released this week, after that this will get merged and will be ready on development for release 3.2.0. Date of 3.2.0 is unknown for now. :)

austinwright (approved):
👍

austinwright (changes requested):
Need to fix CI caching...

Kaspik:
I believe I need to debug this on separate PR, and this PR has nothing to do with CI, so merging to develop and will start new branch and PR for that! (also better to split two things to two branches)

PR #1278, opened 2020-03-12
Fix image view tap gesture handler.

### LINKED ISSUE #1264: MessageCellDelegate didTapImage secret update
Hi everyone,

Nice repo. Thanks for all the efforts! It's nice to see we are having updates again.

Anyway, I am just wondering why the MessageCellDelegate didTapImage method was a 'secret update'...

It would have been nice to get any kind of mention that didTapMessage no longer triggered for photo messages.

Peace

bguidolim:
The touchLocation set before is a wrong calculation if you want to compare it with ImageView's frame. Tapping in the middle of the image usually works, but if you tap in the edges, one of them won't work, especially if you have an image with very different width and height.

I don't think so.

Where has this been tested?
---------------------------
**Devices/Simulators:** Simulator - iPhone 11 Pro

**iOS Version:** iOS 13

**Swift Version:** 5

**MessageKit Version:** 3.1.0-beta.1

### DIFF
--- a/Sources/Views/Cells/MediaMessageCell.swift
+++ b/Sources/Views/Cells/MediaMessageCell.swift
@@ -84,14 +84,13 @@ open class MediaMessageCell: MessageContentCell {
     
     /// Handle tap gesture on contentView and its subviews.
     open override func handleTapGesture(_ gesture: UIGestureRecognizer) {
-        let touchLocation = gesture.location(in: imageView)
-        
-        switch true {
-        case imageView.frame.contains(touchLocation):
-            delegate?.didTapImage(in: self)
-        default:
-            break
+        let touchLocation: CGPoint = convert(gesture.location(in: self), to: imageView)
+
+        guard imageView.bounds.contains(touchLocation) else {
+            super.handleTapGesture(gesture)
+            return
         }
+        delegate?.didTapImage(in: self)
     }
     
 }
--- a/Sources/Views/Cells/MediaMessageCell.swift
+++ b/Sources/Views/Cells/MediaMessageCell.swift
@@ -84,7 +84,7 @@ open class MediaMessageCell: MessageContentCell {
     
     /// Handle tap gesture on contentView and its subviews.
     open override func handleTapGesture(_ gesture: UIGestureRecognizer) {
-        let touchLocation = gesture.location(in: self)
+        let touchLocation = gesture.location(in: imageView)
         
         switch true {
         case imageView.frame.contains(touchLocation):


martinpucik:
Thank you for the PR @bguidolim 💪 
Could you please check if your changes fix #1264 ?

Kaspik (approved):
@austinwright Not sure if we want this to the 3.1.0 release or to the `development`.

bguidolim:
@Kaspik @austinwright tbh, the development process is a little bit confusing for me. Please let me know if I need to do something else.

Kaspik:
Nothing needed to be done here, everything is allright!

Sorry for confusion.

The main branch for development is `development`. After about year, we released 3.1.0-beta.1 by merging `development` to `master` and it's in beta for almost 2 weeks. Once we release 3.1.0 from master (this week, cc @austinwright ), we will continue development on `development`. Master is only for releases afaik right now.

The confusing part here is setup of this repo and setup of protected branches, but I can't update it and fix it yet. Will do once we talk with Austin about permissions!

austinwright:
Can you be more specific about `What does this implement/fix? Explain your changes.`?

> This PR fixes a wrong gesture handler.

Doesn't really explain anything.

Sources/Views/Cells/MediaMessageCell.swift
```diff
@@ -84,14 +84,13 @@ open class MediaMessageCell: MessageContentCell {
...
-        let touchLocation = gesture.location(in: self)
-        
-        switch true {
-        case imageView.frame.contains(touchLocation):
-            delegate?.didTapImage(in: self)
-        default:
-            break
+        let touchLocation: CGPoint = convert(gesture.location(in: self), to: imageView)
+
+        guard imageView.bounds.contains(touchLocation) else {
```
austinwright, line 89: I don't see the benefit of converting the touch location to imageView and checking bounds over getting the original touch location and checking frame - they produce the same result, and doing the conversion is extra work.

Sources/Views/Cells/MediaMessageCell.swift
```diff
@@ -84,14 +84,13 @@ open class MediaMessageCell: MessageContentCell {
...
+            super.handleTapGesture(gesture)
```
austinwright, line 90: Did you test all the cases associated with this? We might want to write some tests against this change

austinwright (changes requested):
see above comments

austinwright:
Should we implement `func cellContentView(canHandle touchPoint: CGPoint) -> Bool` instead of overriding `func handleTapGesture(_ gesture: UIGestureRecognizer)`? Just a thought. Otherwise I think the call to `super` is a needed change for 3.1.0.

Sources/Views/Cells/MediaMessageCell.swift
bguidolim: Actually, the touchLocation set before is a wrong calculation if you want to compare it with ImageView's frame. Tapping in the middle of the image usually works, but if you tap in the edges, one of them won't work, especially if you have an image with very different width and height.
I had the same problem when I implemented a custom cell to handle link preview, I've just applied the same solution for this case.

Sources/Views/Cells/MediaMessageCell.swift
bguidolim: I didn't test all the cases. I am assuming it's working properly on the superclass.

PR #1287, opened 2020-03-17
Prepare 3.1.0 final release #trivial

Kaspik:
- Update podspec file to final name 3.1.0
- Update Example project to use final version 3.1.0

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -2,8 +2,8 @@ PODS:
   - InputBarAccessoryView (4.3.2):
     - InputBarAccessoryView/Core (= 4.3.2)
   - InputBarAccessoryView/Core (4.3.2)
-  - MessageKit (3.1.0-beta.1):
-    - InputBarAccessoryView (~> 4.3.0)
+  - MessageKit (3.1.0):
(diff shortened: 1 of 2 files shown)


Example/Podfile.lock
```diff
@@ -3,7 +3,7 @@ PODS:
     - InputBarAccessoryView/Core (= 4.3.2)
   - InputBarAccessoryView/Core (4.3.2)
   - MessageKit (3.1.0):
-    - InputBarAccessoryView (~> 4.3.2)
+    - InputBarAccessoryView (~> 4.3.0)
```
austinwright, line 6: We want this at 4.3.2. The swift package manager stuff is on `master` after the 4.3.2 tag.

Kaspik: It points to 4.3.2 thanks to `~>` which means it's the highest `4.3.x`

Kaspik: As you can see above that btw. :)

austinwright: 🤦‍♂

austinwright (approved):
🎉

PR #1356, opened 2020-05-19
Add 3 new categorized issue templates #trivial

kinoroy:
Our current issue template is a one-size-fits-all approach that could be improved by addressing the different needs of the 3 typical types of issues being: 

1. Bug reports (this is the type of issue that the current template is most designed for)
2. Feature requests 
3. Questions/Support

This PR creates 3 separate issue templates, allowing the user to select the template they would like to use when opening up their issue. By selecting the correct template, Github will also tag the issue with the appropriate label (‘bug?’, ‘feature request’, or ‘question’). This will help us sort through new issues with less triage.

### DIFF
--- a/.github/ISSUE_TEMPLATE/--question-support.md
+++ b/.github/ISSUE_TEMPLATE/--question-support.md
@@ -1,0 +1,30 @@
+---
+name: "❓ Question/Support"
+about: Ask a question about how to use MessageKit
+title: ''
+labels: question
+assignees: ''
+
+---
+
+<!--
+🛑 Read this 🛑
+Before you open an issue. Please check the quick start guide, the FAQ, and the example project to see if your question is answered there (links in the README). Please also search through previous issues to see if your question was previously answered. 
+Also consider posting(/crossposting) your question to StackOverflow (with the tag 'messagekit').
+
+⚠️ If you need to include code, please wrap it in backticks so it will get Swift highlighting like this:
+```swift
+*your code goes here*
+```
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/--question-support.md
```diff
@@ -0,0 +1,30 @@
...
+about: Ask a question about how to use MessageKit
+title: ''
+labels: question
+assignees: ''
+
+---
+
+<!--
+🛑 Read this 🛑
+Before you open an issue. Please check the quick start guide, the FAQ, and the example project to see if your question is answered there (links in the README). Please also search through previous issues to see if your question was previously answered.
```
Kaspik: Should we have the links here too?

Kaspik (commented):
Looks gooood

.github/ISSUE_TEMPLATE/--question-support.md
kinoroy: I thought about links but because it's inside a comment, the link wouldn't be clickable because it wouldn't render anywhere. We could still put the link in there but they'd have to copy and paste it. If thats cool then sure

Kaspik: Yeah I'm just thinking from user perspecite - reading this - would you rather copy the link to new tab, or would you go multiple clicks to readme to find out it didn't help and then return here? I feel like without navigating anywhere users will use it more...

Kaspik (approved):

PR #1365, opened 2020-06-01
Use messages key instead of string  #trivial

austinwright:
See #1364.

### DIFF
--- a/Example/Sources/Settings+UserDefaults.swift
+++ b/Example/Sources/Settings+UserDefaults.swift
@@ -31,12 +31,12 @@ extension UserDefaults {
     // MARK: - Mock Messages
     
     func setMockMessages(count: Int) {
-        set(count, forKey: "mockMessages")
+        set(count, forKey: messagesKey)
         synchronize()
     }
     
     func mockMessagesCount() -> Int {
-        if let value = object(forKey: "mockMessages") as? Int {
+        if let value = object(forKey: messagesKey) as? Int {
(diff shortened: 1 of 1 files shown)


martinpucik (approved):
LGTM! 🚀

Example/Sources/Settings+UserDefaults.swift
```diff
@@ -31,12 +31,12 @@ extension UserDefaults {
     // MARK: - Mock Messages
     
     func setMockMessages(count: Int) {
-        set(count, forKey: "mockMessages")
+        set(count, forKey: messagesKey)
```
martinpucik:
```suggestion
        set(count, forKey: UserDefaults.messagesKey)
```

should fix the failing build

Example/Sources/Settings+UserDefaults.swift
```diff
@@ -31,12 +31,12 @@ extension UserDefaults {
...
         synchronize()
     }
     
     func mockMessagesCount() -> Int {
-        if let value = object(forKey: "mockMessages") as? Int {
+        if let value = object(forKey: messagesKey) as? Int {
```
martinpucik:
```suggestion
        if let value = object(forKey: UserDefaults.messagesKey) as? Int {
```

^^

austinwright: Yep - I was already making the change when you commented 😛

PR #1386, opened 2020-06-22
Convert internal, static colors into color assets.

### LINKED ISSUE #1301: Convert internal, static colors into color assets.
Xcode 9 introduced Color Assets. Now that MessageKit is targeting iOS 11+, we can make use of these color assets and gain their benefits - mainly that color asset (and system defined semantic color) instances automatically adjust to the current environment settings. So if an end-user has their phone set to automatically switch from light to dark mode, currently a framework client would need to refresh the view programatically. This means listening for notifications and other boilerplate code. Otherwise the end-user does not have a seamless experience and would need to make some manual interaction to reload/refresh the view.

Color Assets should have a light and dark mode, and should also provide high contrast variants of the light and dark mode colors.

Color Assets should be named in a way that conveys the intended use of the color, rather than specific color values.

### LINKED ISSUE #1321: How can i change [PlayButtonView] background?
How can I change the color of the play button in video messages according to the dark mode and light mode?

-> MessageKit (3.1.0)

-> Real Device: 13.4
-> Simulator: 13.4

-> Swift 5.0

- What device(s) are you testing on? Are these simulators?
-> Real Device: iPhone 7
-> Simulator: All Simulator

-> Yes, there is also this problem in the sample MessageKit project.

### LINKED ISSUE #1335: Video message playback icon bug in dark mode
The "video playback icon" that is shown on video messages is too dark when iOS is in dark mode.  This may be the same issue that is referred to in [#1321](https://github.com/MessageKit/MessageKit/issues/1321). See screenshots below from the Example app.

IMG_0139
IMG_0138

kinoroy:

- As described by @austinwright in issue #1301, this PR replaces the static `UIColor` constants with Xcode Color Assets, which are edited visually in Xcode, allowing you to combine appearances for dark/light mode, and high contrast in one described 'color.  These color assets are named for the use of the color rather than the color value itself.

- Also fixes an issue where the video play button was too dark in dark mode (Issue #1335). Specifically, replaces the background color of the video playback button with a `UIVisualEffectsView` which blurs the background similar to how it looks on iMessage. 

- Also fixes an issue where the video play triangle was not properly centered in the button view (no issue #).

Yes:
Closes #1301  
Closes #1335 
Closes #1321 

Any other comments?
-------------------
- I imagine there is a small performance impact of using a `UIVisualEffectsView` in the video play button as opposed to a plain color background however I quickly tested by profiling scrolling through 99 video messages on an iPhone 6 (a 6 year-old phone!) in the example app and I was able to scroll up to 60FPS.

- The addition of the `UIVisualEffectsView` could also be its own PR, but I added it since I was changing the background color there anyways and its a small change. 

- For now, I added static vars in the extension of the `UIColor` class to make the references to the color assets type-safe, but this is something we could maybe have automated for us like with [SwiftGen](https://github.com/SwiftGen/SwiftGen). That could be its own PR in the future.

Where has this been tested?
---------------------------
**Devices/Simulators:** … iPhone 11 Simulator, iPhone 6 Device

**iOS Version:** … iOS 12, and iOS 13

**Swift Version:** … Swift 5

**MessageKit Version:** … 3.1.0


kinoroy:
Screenshots showing fixes to the video play button:
Video Message Comparison

Kaspik (approved):
Love this, thanks!

PR #1410, opened 2020-07-18
Add SwiftUI view to example project #trivial

kinoroy:
👻

- Adds an SwiftUI view to the example project showing how to use MessageKit within a SwiftUI view

Any other comments?
-------------------
- Right now it just calls reloadData whenever the view state changes. It would be a better example to use the new UICollectionViewDiffableDataSource API here to calculate the diffs and only reload what's needed (its not too important in a small example but if you had a lot of messages it would be). Could add this as a future improvement to the example.

 - This example requires iOS 13 to run since it uses SwiftUI but I marked everything with `@available(iOS 13.0, *)` and weakly linked the SwiftUI library so it shouldn't be a problem compiling for earlier versions, they just wouldn't be able to see the SwiftUI example specifically.

- Added #trivial since this is only an example project change and not an MessageKit library change so doesn't require a CHANGELOG entry

Where has this been tested?
---------------------------
**iOS Version:** … 13 (Requires iOS 13 to run)
**Swift Version:** … 5
**MessageKit Version:** … 3.1.0


Example/Sources/Views/SwiftUI/MessagesView.swift
```diff
@@ -0,0 +1,122 @@
...
+//
+//  Created by Kino Roy on 2020-07-18.
+//  Copyright © 2020 MessageKit. All rights reserved.
+//
+
+import SwiftUI
+import MessageKit
+import InputBarAccessoryView
+
+class MessageSwiftUIVC: MessagesViewController {
```
Kaspik: final please :)

Example/Sources/Views/SwiftUI/MessagesView.swift
```diff
@@ -0,0 +1,122 @@
...
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        // Because SwiftUI wont automatically make our controller the first responder, we need to do it on viewDidAppear
+        becomeFirstResponder()
+        messagesCollectionView.scrollToBottom(animated: true)
+    }
+}
+
+@available(iOS 13.0.0, *)
```
Kaspik: can we have just `13.0` ?

Example/Sources/Views/SwiftUI/MessagesView.swift
```diff
@@ -0,0 +1,122 @@
...
+        var messages: Binding<[MessageType]>
+        init(messages: Binding<[MessageType]>) {
+            self.messages = messages
+        }
+        
+    }
+
+}
+
+@available(iOS 13.0.0, *)
```
Kaspik: ^^

Example/Sources/Views/SwiftUI/MessagesView.swift
```diff
@@ -0,0 +1,122 @@
...
+        return NSAttributedString(string: name, attributes: [NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .caption1)])
+    }
+    
+    func messageBottomLabelAttributedText(for message: MessageType, at indexPath: IndexPath) -> NSAttributedString? {
+        let dateString = formatter.string(from: message.sentDate)
+        return NSAttributedString(string: dateString, attributes: [NSAttributedString.Key.font: UIFont.preferredFont(forTextStyle: .caption2)])
+    }
+}
+
+@available(iOS 13.0.0, *)
```
Kaspik: ^^

Example/Sources/Views/SwiftUI/MessagesView.swift
```diff
@@ -0,0 +1,122 @@
...
+extension MessagesView.Coordinator: InputBarAccessoryViewDelegate {
+    func inputBar(_ inputBar: InputBarAccessoryView, didPressSendButtonWith text: String) {
+        let message = MockMessage(text: text, user: SampleData.shared.currentSender, messageId: UUID().uuidString, date: Date())
+        messages.wrappedValue.append(message)
+        inputBar.inputTextView.text = ""
+    }
+}
+
+@available(iOS 13.0.0, *)
```
Kaspik: ^^

Example/Sources/Views/SwiftUI/SwiftUIExampleView.swift
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+    
+    private func cleanupSocket() {
+        MockSocket.shared.disconnect()
+    }
+    
+}
+
+@available(iOS 13.0.0, *)
```
Kaspik: ^^

Kaspik (approved):
Just a nitpick but would be nice to update it :) Thanks!

PR #1427, opened 2020-08-05
Update to InputBarAccessoryView 5.0.0, update SPM support

Kaspik:
- Switch to IBAV 5.0.0 from main organization


kinoroy (approved):
Thanks @Kaspik !

PR #1442, opened 2020-08-27
Update IBAV to 5.1.0 with Dark Mode support #trivial

Kaspik:
Update IBAV dependency to 5.1.0 with dark mode support.


martinpucik (approved):
LGTM! 🚀

PR #1456, opened 2020-09-08
Fix compilation issues by importing UIKit

tikhop:
Just imported UIKit framework where it must be imported

Haven't found any

For example:
.../SourcePackages/checkouts/MessageKit/Sources/Protocols/LinkItem.swift:51:25: Cannot find type 'UIImage' in scope

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:**Simulator/Device

**iOS Version:**14.0

**Swift Version:**5.4

**MessageKit Version:** 3.3.0

### DIFF
--- a/Sources/Layout/LinkPreviewMessageSizeCalculator.swift
+++ b/Sources/Layout/LinkPreviewMessageSizeCalculator.swift
@@ -23,6 +23,7 @@
  */
 
 import Foundation
+import UIKit
 
 open class LinkPreviewMessageSizeCalculator: TextMessageSizeCalculator {
 
--- a/Sources/Models/LinkPreviewFonts.swift
+++ b/Sources/Models/LinkPreviewFonts.swift
@@ -23,6 +23,7 @@
  */
 
 import Foundation
+import UIKit
 
 public struct LinkPreviewFonts: Equatable {
     let titleFont: UIFont
--- a/Sources/Protocols/LinkItem.swift
+++ b/Sources/Protocols/LinkItem.swift
@@ -24,6 +24,7 @@
 
 import Foundation
 import CoreGraphics
+import UIKit
 
 /// A protocol used to represent the data for a link preview message.
 public protocol LinkItem {


martinpucik:
this is an interesting issue, our CI is able to build the framework and the example project without these imports 🤷‍♂️ 😹

tikhop:
@martinpucik yeah, it only happens when you import framework via SPM. 
I had the same issue with my lib, it was working well via cocoapods or manual install, but then I tried SPM and bumped into the same problem.

martinpucik:
SMP package from https://github.com/MessageKit/MessageKit/tree/spm-5_2 is working too without these imports, but I don't see a reason we should not add these. 

@Kaspik should this go to development branch first and the from dev to master or are we ok with separate PRs?

Kaspik:
Everything needs to go to development, no master please.

Also we could re-apply patches from spm_5_2 branch as well - but again - not sure why we would want to do it as SPM is not yet really supported. :)

Kaspik:
For this specific case (iOS 14, Swift 5.3) I would recommend pointing to that branch anyways.

Kaspik:
@tikhop Note on SPM: MessageKit has assets, and these are not supported by SPM before Swift 5.3 / iOS 14 - that's why we don't have SPM support. adding these imports will fix your compiler, but you won't be able to use ANY of cell types except custom. :) I would suggest pointing to the branch as mentioned by @martinpucik or wait with SPM itself few more days/weeks for final XCode 12 release. :)

tikhop:
@Kaspik Thanks for the note. I know about assets and spm and before making a PR, I've checked the contribution doc. Since this issue is just a bug in the code, where you are trying to access a class declaration from UIKit framework without even importing the framework and it breaks the compilation process I have decided that it perfectly suits the case from the contribution doc where you should make a pr to master and merge to dev. That's what I did. 

And it works well in my case with xcode 12...

tikhop:
@Kaspik I've just tried `spm-5_2` and, I'm able to compile, but later I get a crash when trying to access `MessageKitAssets`, master works in this case.

PR #1457, opened 2020-09-08
Fix compilation issues by importing UIKit (dev branch)

tikhop:
Please, find more info here: https://github.com/MessageKit/MessageKit/pull/1456


martinpucik (approved):
LGTM! 🚀
Looks like this may fix a compilation issue for some clients so let's merge it!

PR #1463, opened 2020-09-15
Fix module imports

### LINKED ISSUE #1462: Fail to build when using `MessageKit` as a Swift package
`MessageKit` fail to build when using it as a Swift package. This happens because In some source files the `UIKit` module is not explicitly imported.

karimhm:
Fixes  #1462


Kaspik:
This has been already addressed in https://github.com/MessageKit/MessageKit/pull/1457. Also, see discussion there about SPM.

We will introduce full SPM support one XCode 12 is finally out.

PR #1469, opened 2020-09-20
Fix casting MessageType to custom type for LinkPreview cells

### LINKED ISSUE #1468: ConcreteMessageType  seems to break existing code
Hey there,

In my app I have a custom Message class that implements the MessageType protocol. In some places I cast the MessageType to my custom class. The code is now breaking if the cell is a LinkPreview cell because LinkPreview is a ConcreteMessageType which  also extends MessageType but is internal. 

What was the reason to make that class internal ?  What is the best way to support ConcreteMessageType with my own class ? 

Cheers.

kinoroy:
Fixes an issue where the user's message object would be replaced with a ConcreteMessageType object within the LinkPreviewMessageCell configure method and LinkPreviewMessageSizeCalculator, which was causing issues for users who needed to cast the message object to their custom class within the MessageKit delegate methods. This cast would fail since MessageKit has replaced their message object behind the scenes, and this is unexpected. 

Closes #1468 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------

**iOS Version:** … 14

**Swift Version:** … 5.3

**MessageKit Version:** … 3.4.1


Kaspik (approved):
LGTM, thanks!

PR #1476, opened 2020-10-01
Remove keyboard observers in viewWillDisappear, not deinit

### LINKED ISSUE #1451: Keyboard observers persist after MessagesViewController is dismissed
I'm filing a bug for the same problem that was previously noted by this question: https://github.com/MessageKit/MessageKit/issues/1212

The MessagesViewController removes keyboard observers only in `deinit`, not in `viewWillDisappear`. This is a problem because other parts of the app should be able to use the keyboard without having to trigger things happening in MessageKit.

When `handleKeyboardDidChangeState` is called, it can end up triggering a call to `messageForItem`, but at this point, the dataSource could be cleared, which then causes a crash. I also wonder why `messageForItem` doesn't return an optional? That prevents us from returning nil. 

Here are the relevant pieces:

Inside `MessagesViewController.swift`
```swift
        // Hack to prevent animation of the contentInset after viewDidAppear
        if isFirstLayout {
            defer { isFirstLayout = false }
            addKeyboardObservers()
            messageCollectionViewBottomInset = requiredInitialScrollViewBottomInset()
        }
        adjustScrollViewTopInset()
 }

deinit {
        removeKeyboardObservers()
        removeMenuControllerObservers()
        removeObservers()
        clearMemoryCache()
}
```

and inside `MessagesViewController+Keyboard.swift`
```swift
func addKeyboardObservers() {
        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleKeyboardDidChangeState(_:)), name: UIResponder.keyboardWillChangeFrameNotification, object: nil)
        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.handleTextViewDidBeginEditing(_:)), name: UITextView.textDidBeginEditingNotification, object: nil)
        NotificationCenter.default.addObserver(self, selector: #selector(MessagesViewController.adjustScrollViewTopInset), name: UIDevice.orientationDidChangeNotification, object: nil)
    }

@objc
    private func handleKeyboardDidChangeState(_ notification: Notification) {
        guard !isMessagesControllerBeingDismissed else { return }

        guard let keyboardStartFrameInScreenCoords = notification.userInfo?[UIResponder.keyboardFrameBeginUserInfoKey] as? CGRect else { return }
        guard !keyboardStartFrameInScreenCoords.isEmpty || UIDevice.current.userInterfaceIdiom != .pad else {
        // [comments] 
            return
        }

        guard self.presentedViewController == nil else {
            // This is important to skip notifications from child modal controllers in iOS >= 13.0
            return
        }
        
        // [very long comment here, justifying this logic below is necessary]

        guard let keyboardEndFrameInScreenCoords = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect else { return }
        let keyboardEndFrame = view.convert(keyboardEndFrameInScreenCoords, from: view.window)

        let newBottomInset = requiredScrollViewBottomInset(forKeyboardFrame: keyboardEndFrame)
        let differenceOfBottomInset = newBottomInset - messageCollectionViewBottomInset

        if maintainPositionOnKeyboardFrameChanged && differenceOfBottomInset != 0 {
            let contentOffset = CGPoint(x: messagesCollectionView.contentOffset.x, y: messagesCollectionView.contentOffset.y + differenceOfBottomInset)
            messagesCollectionView.setContentOffset(contentOffset, animated: false)
        }

        messageCollectionViewBottomInset = newBottomInset
    }
```

A function for adjusting the collectionView's bottom inset is _irrelevant_ when the messagesViewController is not visible. Please fix this.

Latest version as of this writing (9/2/2020)

iOS 13.3

Swift 5

- What device(s) are you testing on? Are these simulators?
No. iPhone 7 and iPhone X

Yes.

kinoroy:
As suggested by @neugierige , cleans up the keyboard observers when MessagesViewController disappears from the view, instead of waiting for `deinit`. This prevents us from using resources to adjust the bottom offset when the MessagesViewController isn't on screen any longer to display those changes

Closes #1451 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------

**iOS Version:** … iOS 14

**Swift Version:** … 5

**MessageKit Version:** … 3.4.1

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -124,6 +124,13 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
         addObservers()
     }
     
+    open override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        if !isFirstLayout {
+            addKeyboardObservers()
+        }
+    }
+    
     open override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         isMessagesControllerBeingDismissed = false
@@ -132,6 +139,7 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
     open override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         isMessagesControllerBeingDismissed = true
+        removeKeyboardObservers()
     }
     
     open override func viewDidDisappear(_ animated: Bool) {
@@ -156,7 +164,6 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
     // MARK: - Initializers
 
     deinit {
-        removeKeyboardObservers()
         removeMenuControllerObservers()
         removeObservers()
         clearMemoryCache()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Fixes an issue with casting MessageType to a custom type when using LinkPreview cells [#1469](https://github.com/MessageKit/MessageKit/pull/1469) by [@kinoroy](https://github.com/kinoroy)
 
+- Fixes an issue where the MessagesViewController keyboard observers were not cleaned up when MessagesViewController was no longer visible on screen [#1476](https://github.com/MessageKit/MessageKit/pull/1476) by [@kinoroy](https://github.com/kinoroy)
+
 ## 3.4.0
 
 ### Changed


martinpucik (approved):
LGTM! 🚀

neugierige:
sweet! thank you so much!

Sources/Controllers/MessagesViewController.swift
```diff
@@ -124,6 +124,13 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
         addObservers()
     }
     
+    open override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        if !isFirstLayout {
+            addKeyboardObservers()
```
Kaspik, line 130: So addObservers doesn't do this?

kinoroy: @Kaspik No, addObservers only adds the observer for didReceiveMemoryWarningNotification, it doesn't call addKeyboardObservers

Kaspik: So `addKeyboardObservers` wasn't called at all? Okay.. I trust you anyways, so feel free to merge it, but probably to development?

PR #1479, opened 2020-10-05
Moved tests to XCTest without external dependencies

martinpucik:
- Removed Quick/Nimble dependencies for tests
- Removed .xcproject file for running tests & building framework

🚫 

🚫 

Any other comments?
-------------------
🚫 

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …


Package.swift
```diff
@@ -27,26 +27,22 @@ import PackageDescription
 
 let package = Package(
     name: "MessageKit",
-    platforms: [.iOS(.v11)],
+    platforms: [.iOS(.v12)],
```
Kaspik, line 30: lol, totally missed this, thanks!

Tests/MessageKitTests/MessageKitTests.swift
```diff
@@ -0,0 +1,33 @@
...
+ SOFTWARE.
+ */
+
+import Foundation
+import XCTest
+@testable import MessageKit
+
+final class MessageKitTests: XCTestCase {
+    static var allTests = [
+        ""
```
Kaspik, line 31: is this correct?

Kaspik (commented):
Nice job!

Tests/MessageKitTests/MessageKitTests.swift
martinpucik: nope, will update this 👍

GitHubActions/build.sh
```diff
@@ -35,15 +35,13 @@ MODE="$1"
 
 if [ "$MODE" = "tests" -o "$MODE" = "all" ]; then
   echo "Running MessageKit tests."
-  carthage bootstrap --platform ios
-  set -o pipefail && xcodebuild test -project MessageKit.xcodeproj -scheme MessageKitTests -destination "platform=iOS Simulator,name=iPhone 11 Pro" CODE_SIGNING_REQUIRED=NO | xcpretty -c
+  set -o pipefail && xcodebuild test -scheme MessageKit -sdk iphonesimulator -destination "platform=iOS Simulator,name=iPhone 11" | xcpretty -c
```
Kaspik, line 38: You need to switch to XCode 12 (non-default) for Swift 5.3 - https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

martinpucik: Fixed here 👍 
https://github.com/MessageKit/MessageKit/pull/1479/commits/37648f1dbe88525cc7b631aac95a1224f4a97dbd#

Tests/MessageKitTests/MessageKitTests.swift
martinpucik: but it shouldn't be a blocker, this is used for running tests on Linux 😹

Kaspik: Lol, don't ever do that. 😂

Kaspik (approved):
Do we need the carthage file itself etc? or can be removed also, not only the `.private` one?

kinoroy:
Can we switch IBAV to use SPM/Cocoapod dependencies and just be able to remove Carthage entirely from the repo?

The only issue I think with that would be that people installing the repo manually (not using SPM/Cocoa/Carthage) would need to also manually install IBAV instead of just running Carthage after cloning

Also people using Carthage would have to also list IBAV in their Cartfile in addition to MessageKit but this would actually give them more flexibility because it would no longer forcefully tie a specific IBAV version to a specific MessageKit version.  [Carthage officially recommends that dependencies not be nested](https://github.com/Carthage/Carthage/issues/844#issuecomment-147783855)

martinpucik:
Folks I have no idea how Carthage works and what files are necessary, so feel free to remove anything you think could be removed 😹

Kaspik:
@kinoroy Let's do it in another PR if you have a time for that. :) Merging this!

xcodefriend:
@martinpucik , @Kaspik Can we bring back `MessageKit.xcodeproj/project.pbxproj` file? I use it to build an XCFramework out of MessageKit or is there another way to have a XCFramework whithout `MessageKit.xcodeproj`?

Kaspik:
Maybe directly from built archive? https://github.com/bielikb/xcframeworks#2-create-xcframework-from-built-archives

PR #1483, opened 2020-10-16
Calling the delegate at the end of configure audio cell

vittoridavide:
This fix the fact that the duration text is being override if is also inserted in the delegate method, creating sometimes a glitch or not updating properly the cell.

Tested in iOS 14


Kaspik (approved):
Looks good, thanks!

Issue #187, opened 2017-09-28
Support GIF messages


SD10:
It would be nice to support GIF messages inside of the media cells. I'm not sure how we would go about this:
- A) A new `MessageData` type `case .gif(UIImage)` and a new `MessageCollectionViewCell` subclass
- B) Add a player view to the existing `MediaMessageCell` type. I think this would also allow us to play videos directly from the bubbles/cells without navigating to a player view controller.

@cwalo has already got this working so he may have some input on the path of least resistance

Related issues:
[Add new GIF media type #1869](https://github.com/jessesquires/JSQMessagesViewController/issues/1869)

Jeyhey:
Actually it's not difficult to do it with the current implementation by using this uiimage extension: https://github.com/bahlo/SwiftGif/blob/master/SwiftGifCommon/UIImage%2BGif.swift

cwalo:
I think it's pretty straightforward if you already have the image data. I think controlling playback will be pretty essential, even if it's just internal as a means to manage memory. We did run into issues when trying to play all gifs in a conversation, so we made them 'tap-to-play'. Ideally when gif cells went off screen, they'd stop playing and start when coming back on screen - that might be a little tough.

A quick aside:
What we're doing is technically 'unfurling' urls in the message body and that all happens asynchronously. We _do_ know immediately if there's something to unfurl, but we _don't_ know if we're looking at a gif, an image, or a webpage until we do a request for the content-type.

Currently accomplishing this with SwiftLinkPreview and YYWebImage.

cwalo:
@Jeyhey Sounds like SwiftGif has some memory leak issues with their implementation and that's definitely something we wouldn't want to inherit.
https://github.com/bahlo/SwiftGif/issues/6

Jeyhey:
@cwalo I wasn't suggesting to inherit anything from SwiftGif. I was saying that the current implementation is already enough to display gifs if the controller subclass passes a uiimage instantiated with the SwiftGif-extention to the message.

Thanks for pointing out the memory leak issues. That's too bad.

SD10:
I'm only willing to add 2 external dependencies to this project and this won't happen anytime soon:
1. AsyncDisplayKit / Texture
2. IGListKit

Anything else we have to do by hand 😃

PR #1499, opened 2020-11-13
support gif message

amirpirzad:
support gif messages (mute video)

#187

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 11

**iOS Version:** 14

**Swift Version:** 5

**MessageKit Version:** 3.4.2


Example/Sources/Data Generation/SampleData.swift
```diff
@@ -189,6 +190,9 @@ final internal class SampleData {
         case .Video:
             let image = messageImages.random()!
             return MockMessage(thumbnail: image, user: user, messageId: uniqueID, date: date)
+        case .Gif:
+            let image = messageImages.random()!
+            return MockMessage(thumbnail: image, user: user, messageId: uniqueID, date: date, streamingUrl: "https://www.radiantmediaplayer.com/media/big-buck-bunny-360p.mp4")
```
Kaspik, line 196: why does it have mp4 extension if it's supposed to be gif?

Sources/Views/Cells/MediaMessageCell.swift
```diff
@@ -40,19 +40,40 @@ open class MediaMessageCell: MessageContentCell {
...
+        view.backgroundColor = UIColor.black.withAlphaComponent(0.5)
+        view.layer.cornerRadius = 8
+        view.layer.masksToBounds = true
+        return view
+    }()
+
+    open var gifLabel: UILabel = {
+        let label = UILabel()
+        label.text = "GIF"
+        label.textColor = .white
```
Kaspik: Please verify dark mode

Kaspik (changes requested):
I'm going back na forth with this... it kind of looks the same as video functionality, but it isn't, and can be easily achieved with custom view. Good thing is that this PR is not adding any new dependency... nice!

Added some comments, also please can we get some screenshots (or better - gif) about how it looks like in example app, in normal and in dark mode? Currently don't have time to run it locally and test everything, sorry.... Thanks!

Example/Sources/Data Generation/SampleData.swift
amirpirzad: I think for better performance, we can use mute video.
This method does not require a **GIF** player and any new dependency, 
But if you think it must be **GIF** format, I will do it.

Kaspik: Well, then it's a question. Are we adding support for videos that will be muted, or for .gif images?

amirpirzad: We are supporting video formats.

Kaspik: Thanks for the info. :)

But we do have video type, right? Cell media type. And you can adjust it to be silent. I don't think we should merge this PR as gif support as it's not really that...

amirpirzad: :))) That's right ٫ so I will add a gif player instead of a video player to support the gif format

Kaspik: Before you junp on that (and thanks for the proactivity!), how do you want to do it?

I feel like the easiest way for users is to create new content media type and have their own cell with animated view (from kingfisher) or pinanimatedview (from pinremoteimage) thus this library doesn't have to do it. :)

Definitely not if it means adding new dependency.

amirpirzad: You're welcome,

I think it is good to use **PINRemoteImage**.

Kaspik: I think based on https://github.com/MessageKit/MessageKit/issues/187#issuecomment-333253813 we shouldn't really do that... as I said, adding gif support on project side is super easy, not sure if it's worth it doing it here in this framework. It would add dependency for everyone even tho only some users would need it...

PR #1556, opened 2021-03-17
Make messageTimestampLabel center on messageContainerView

kurtsequoia:
Currently the time stamps have their upper edge aligned vertically with the contentView center, but since the messageContainerView often extends outside of the contentView.  The vertical position of the time labels is very erratic and aligns differently for incoming vs outgoing messages, messages with different content, etc.

That makes it hard to associate the time labels with their messages, and is rather confusing.

This one line fix changes the layout so it vertically centers the upper extent of the label to the messageContainerView center.  That makes it track better visually and makes it much easier to associate time stamps with messages.

I have tested this in the example app In the simulator with all the different content types turned on.

I think it makes a big difference in the usability of the messageTimeStampLabel's.   

Here is the before.  Notice that the alignment is erratic and even when working as intended the text is below center:
timeLabelAlignementBefore

Here is the after:
timeLabelsAlignedAfter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 - Fixes an issue with Scroll problem on new messages with keyboard open [#1529](https://github.com/MessageKit/MessageKit/pull/1529) by [@politan8](https://github.com/politan8)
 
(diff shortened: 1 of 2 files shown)


Kaspik:
Please add screenshots before & after ideally and update changelog. :)

Kaspik (approved):
Awesome, thanks!

PR #1563, opened 2021-03-25
Fix typing indicator calulator based on delegate methods

### LINKED ISSUE #1562: Delegate func typingIndicatorViewSize(...) is not called when implemented
Describe the bug
The func typingIndicatorViewSize(...) delegate function is not called when implemented.

Environment

What version of MessageKit are you using? 3.5.1
What version of iOS are you running on? iOS 14
What version of Swift are you running on? 5
What device(s) are you testing on? Are these simulators? Simulator
Is the issue you're experiencing reproducible in the example app? yes

Additional context
This method is declared in the protocol but not used by the delegate at any time.

Kaspik:
Fix typing indicator size based on delegate methods.

Closes https://github.com/MessageKit/MessageKit/issues/1562

Any other comments?
-------------------
This is a breaking change as delegate method has changed, but it shouldn't break anything as default implementation was used in most of the cases.

Where has this been tested?
---------------------------
**Devices/Simulators:** Simulator

**iOS Version:** 14.4

**Swift Version:** 5.3

**MessageKit Version:** master

### DIFF
--- a/Sources/Layout/TypingIndicatorCellSizeCalculator.swift
+++ b/Sources/Layout/TypingIndicatorCellSizeCalculator.swift
@@ -26,18 +26,13 @@ import UIKit
 
 open class TypingCellSizeCalculator: CellSizeCalculator {
 
-    open var height: CGFloat = 62
-
     public init(layout: MessagesCollectionViewFlowLayout? = nil) {
         super.init()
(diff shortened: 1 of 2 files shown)


martinpucik (approved):
LGTM! 🚀

PR #1577, opened 2021-04-23
Added support to display custom cells for non MessageType.custom

jvigneshcs:
I have added enough data source and delegate methods to display customized UICollectionViewCell for `MessageType`s other than .custom

No

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** Device & Simulators

**iOS Version:** iOS 11.0.1, 11.3

**Swift Version:** 5.3.2

**MessageKit Version:** 3.1.0


ekazaev (approved):

Kaspik (changes requested):
Can we please also add a new controller to examples that demonstrates the usage and use case this could utilize? Not sure I understand the expected behavior. Thanks

jvigneshcs:
Simulator Screen Shot - iPhone 8 - 2021-05-01 at 21 32 30
> Can we please also add a new controller to examples that demonstrates the usage and use case this could utilize? Not sure I understand the expected behavior. Thanks

@Kaspik , I have added example view controller to demonstrate the usage.
Please find the attached screenshot for reference.

Kaspik (changes requested):
Awesome addition to the examples, thanks! Please - last thing - add an entry into the changelog before merging this :)

Kaspik (approved):

jvigneshcs:
@Kaspik , I have made these changes in my fork to support iOS 9 and above, can you please create a branch for iOS 9 above from tag 3.1.0?
So that I can merge these changes into that branch and start using it.

Kaspik:
I do not think that's a good idea, and I'm definitely not able to help you with that (we dropped CocoaPods for a reason, they are not really compatible with Apple Silicon that I / we use already, and 3.1.0 uses internally CocoaPods for Example project), but I've created a branch `3_1_1` for you so you can open a PR to that if you think you can do it.

PR #1579, opened 2021-04-23
(fix) restore an example file that was previously deleted

mredig:
used in `Documentation/MessageInputBar.md`

A previous commit (d16e5f8078c14cbdc70085d06bea5d9874b19c18) deleted this file, seemingly not intentionally. Documentation continues to link to this now deleted file, so I restored it to the state prior to deletion.

…

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …


mredig:
Upon further inspection, it kind of looks like another png that was added in that same commit (Assets/InputBarAccessoryViewLayout.png) was intended to replace the original file, but the markdown file wasn't updated. @nathantannar4 Can you confirm?

Kaspik:
Yes I believe we replaced MessageInputBar with the InputBarAccessoryView and the image was supposed to be replaced. :)

Kaspik:
Are you planning to use the replacement image?

Kaspik:
Fixed.

PR #1640, opened 2021-11-18
Add overridable UIScrollView Delegate

mehmetbaykar:
It has scroll view delegate functions to listen to the scroll action of the message view controller. This will help developers override any action they want in one of these delegate functions.


Kaspik:
I would really appreciatw screenshot and video of the behavior a d how it works. Thanks! :)

mehmetbaykar:
Hi @Kaspik 
If someone is interested in adding new capabilities to messageCollectionView, they may need its scrolling callback. So adding scroll delegates to MessagesViewController will fix this issue.

Kaspik:
I understand that, but this PR also adds some button and stuff so I would like to see that. :)

mehmetbaykar:
Sorry again. The second commit should've been here as well. I will make a new PR

Issue #1633, opened 2021-11-02
Dismissing the keyboard does not work properly


mehmetbaykar:
When dismissing the keyboard with swiping from top to bottom on the accessory view, the layout of the collection view changes. I believe it has to do with InputAccessoryView and [MessagesViewController-Keyboard](https://github.com/MessageKit/MessageKit/blob/master/Sources/Controllers/MessagesViewController%2BKeyboard.swift)

Please check the video attached.

https://user-images.githubusercontent.com/20362928/139854671-f117de7e-3d09-494f-a8e6-546b1568c3e4.mov

mehmetbaykar:
@Kaspik  any update on that?

mehmetbaykar:
Found the issue will make a PR.

Kaspik:
How does it look with the PR you closed?

mehmetbaykar:
Sorry. I realised that this PR has an extra commit that should've been. I will make a new PR when I can.

PR #1641, opened 2021-11-18
Fix Scroll issue when dismissing the keyboard

mehmetbaykar:
Fix the scrolling issue [here](https://github.com/MessageKit/MessageKit/issues/1633)


Kaspik (commented):
Just making sure if this works on iOS 12 - iOS 15? If you could add tested devices to the PR description + add changelog entry, that would be awesome. Thanks!

PR #1658, opened 2021-12-20
Improvement in logic for scrollToLastItem

Ewarrender:
The default implementation of scrollToLastItem only considers the last section and therefore doesn't work if that section is empty. This change now iterates through the sections until it finds the actual last valid item index path. The `scrollToLastItem` function now also uses the `indexPathForLastItem` property instead of duplicating logic (it also wasn't being used).

No

N/A

Any other comments?
-------------------
N/A

Where has this been tested?
---------------------------
**Devices/Simulators:** N/A

**iOS Version:** iOS 14 and 15

**Swift Version:** Swift 5.1

**MessageKit Version:** Tested from 3.6.1 up to 3.7.0

### DIFF
--- a/Sources/Views/MessagesCollectionView.swift
+++ b/Sources/Views/MessagesCollectionView.swift
@@ -46,9 +46,14 @@ open class MessagesCollectionView: UICollectionView {
     internal var showMessageTimestampOnSwipeLeft: Bool = false
 
     private var indexPathForLastItem: IndexPath? {
-        let lastSection = numberOfSections - 1
-        guard lastSection >= 0, numberOfItems(inSection: lastSection) > 0 else { return nil }
-        return IndexPath(item: numberOfItems(inSection: lastSection) - 1, section: lastSection)
+        guard numberOfSections > 0 else { return nil }
+        
+        for offset in 1...numberOfSections {
+            let section = numberOfSections - offset
+            let lastItem = numberOfItems(inSection: section) - 1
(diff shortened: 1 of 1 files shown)


Sources/Views/MessagesCollectionView.swift
```diff
@@ -46,9 +46,14 @@ open class MessagesCollectionView: UICollectionView {
...
     private var indexPathForLastItem: IndexPath? {
-        let lastSection = numberOfSections - 1
-        guard lastSection >= 0, numberOfItems(inSection: lastSection) > 0 else { return nil }
-        return IndexPath(item: numberOfItems(inSection: lastSection) - 1, section: lastSection)
+        guard numberOfSections > 0 else { return nil }
+        
+        for offset in 1...numberOfSections {
+            let section = numberOfSections - offset
+            let lastItem = numberOfItems(inSection: section) - 1
+            if lastItem >= 0 { return IndexPath(item: lastItem, section: section) }
```
Kaspik: Can we put the return on new line here please? :)

Kaspik (changes requested):
One nitpick!

Sources/Views/MessagesCollectionView.swift
Ewarrender: @Kaspik no problem, that has been changed now.

Kaspik (approved):

PR #1670, opened 2022-01-26
Changes in gesture recognizer delegate

Romulomga:
- Adding shouldRecognizeSimultaneouslyWithGestureRecognizer method
- Making gesture recognizer methods do public, so they can be overridable

This implementation adds support for developers to override implemented gesture recognizer delegates in MessagesViewController on their own subclass.

No

No

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …

### DIFF
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -518,12 +518,17 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
     // MARK: - UIGestureRecognizerDelegate
            
     /// Check Pan Gesture Direction:
-    open func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+    public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
         guard let panGesture = gestureRecognizer as? UIPanGestureRecognizer else {
             return false
         }
         let velocity = panGesture.velocity(in: messagesCollectionView)
         return abs(velocity.x) > abs(velocity.y)
+    }
+    
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        
+        return true
     }
 }
 
--- a/Sources/Controllers/MessagesViewController.swift
+++ b/Sources/Controllers/MessagesViewController.swift
@@ -518,7 +518,7 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
     // MARK: - UIGestureRecognizerDelegate
            
     /// Check Pan Gesture Direction:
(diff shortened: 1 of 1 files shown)


Sources/Controllers/MessagesViewController.swift
```diff
@@ -518,13 +518,18 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
     // MARK: - UIGestureRecognizerDelegate
            
     /// Check Pan Gesture Direction:
-    open func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+    public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
```
Kaspik: This should stay open so apps can override it.

Sources/Controllers/MessagesViewController.swift
```diff
@@ -518,13 +518,18 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
...
         guard let panGesture = gestureRecognizer as? UIPanGestureRecognizer else {
             return false
         }
         let velocity = panGesture.velocity(in: messagesCollectionView)
         return abs(velocity.x) > abs(velocity.y)
     }
+    
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
```
Kaspik: The same, this should be open.

Kaspik (changes requested):
I believe you misunderstood public vs open. You can't override public, you can override open. :)

Kaspik (approved):

Sources/Controllers/MessagesViewController.swift
```diff
@@ -525,6 +525,11 @@ UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, UIGestureRecogni
         let velocity = panGesture.velocity(in: messagesCollectionView)
         return abs(velocity.x) > abs(velocity.y)
     }
+    
+    open func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
```
Kaspik, line 529: What exactly this method does btw. ? :)

Romulomga: For MessageKit it self, nothing changes because the default return of this method is true. I just add this, because im trying to implement some feature with multiple touch events(what this method is related of) and this method kind of help me(maybe for other to). :)

Kaspik: Well, this is not even override, is it? So what would change if you return false here, if its not used?

Kaspik (changes requested):
Waiting for more info.

Kaspik:
Closing for no activity. Feel free to reopen again with more info. :)

PR #1671, opened 2022-02-03
Added new methods in `MessagesLayoutDelegate` for adjusting alignment of message top and bottom labels 

martinpucik:
- Adds an option to use `MessagesLayoutDelegate` methods for adjusting message top/bottom label alignments without the need to subclass MessageSizeCalculators
- Chat example not used files cleanup

N/A


Kaspik (changes requested):
I know it will be small but could you please create super small V4 Migration info section into Readme file for people that upgrade from V3?

martinpucik:
There are no breaking changes yet in 4.0. I'd add the migration doc when there is something to migrate 😁

Kaspik:
Well you have to add new functions when upgrading to it, they are not optional, right? 😅

martinpucik:
There are default implementations in `MessagesLayoutDelegate` extension so clients don't have to add them 🙌

Kaspik (approved):

PR #1704, opened 2022-05-04
Move `messageInputBar` from `inputAccessoryView` to a subview in `MessagesViewController`

### LINKED ISSUE #120: Move keyboard handling logic into KeyboardTracker object
The current way of keyboard event handling by adjusting the `contentInset` of the `MessagesCollectionView` is a bit limiting. 

Since we pin the `MessagesCollectionView` to fill the superview -- we currently add `top` and `bottom` insets to account for the size of potential `NavigationBar`s, `TabBar`s, and of course the `InputAccessoryView`.

This blocks #117 because you can't adjust the `top` `contentInset` of the view to adjust for other potential views. Which is a significant limitation.

We could add an API that allows users to add an additional header to the `MessagesCollectionView`. Then handle the layout/contentInset based on the size of the provided view. I think this is avoiding the underlying problem and is less flexible.

Instead:
- We should move back to using constraints to manage the position of the view instead of adjusting the `contentInset`. This means no longer pinning to the `superView`.

- We should refactor all this keyboard logic into a `KeyboardTracker` object. Keyboard handling is proving to be a pain point and we are refactoring it every other version. It's time for a dedicated solution.

### LINKED ISSUE #296: Support undocked/split keyboard on iPad

### LINKED ISSUE #345: MessageInputBar overlaps navigationBar in landscape orientation
## General Information

* MessageKit Version: 0.10.1

* iOS Version(s): 11.0.3

* Swift Version: 4

* Devices/Simulators: iPhone 7 and simulator

##  What happened?

    1. Launch Example project in portrait
    2. Start typing and press return on the keyboard to create multiple lines
    3. Rotate to landscape
The MessageInputBar overlaps the navigationBar

I expected the MessageInputBar to stop growing when it reaches the navigationBar

32680108-17a8903a-c661-11e7-951e-9144b1b439c4

### LINKED ISSUE #737: MessageInputBar not showing when MessagesViewController subclass used as ChildViewController
MessageKit Version: 1.0.0
Swift Version: 4.1
Test Device: iPad 12.9(inch)

When adding messageViewController on container view(red color background), seems like inputAccesoryView doesn't show.

screen shot 2018-06-22 at 13 56 55

I tried to use becomFirstResponder in viewdidAppear but the inputAccesoryView seems to take width of the parentViewController (here you can see the code): 

```
override func viewDidLoad() {
    super.viewDidLoad()
    setupNavigationBar(actionsTarget: self)
    setupLocalization()
    
    viewController = ChatViewController()
    addChildViewController(viewController!)
    viewController?.view.frame = containerView.bounds
    containerView.addSubview((viewController?.view)!)
    viewController?.didMove(toParentViewController: self)
     
  }
  
  override func viewDidAppear(_ animated: Bool) {
    super.viewDidAppear(animated)
    viewController?.becomeFirstResponder()
  }
```

screen shot 2018-06-22 at 13 57 34

Is there a way to make inputAccesoryView to be same width as containerView ?

martinpucik:
- Moving the inputBar to subview will solve the bar to be the width of the owning view on iPads
- inputBar should show on MacCatalyst
- inputBar should work with undocked/floating keyboard

- This should close #120 
- This should close #296 
- This should close #345 
- This should close #737 
 
Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

**iOS Version:** …

**Swift Version:** …

**MessageKit Version:** …


Package.swift
```diff
@@ -32,7 +32,8 @@ let package = Package(
         .library(name: "MessageKit", targets: ["MessageKit"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/nathantannar4/InputBarAccessoryView", .upToNextMajor(from: "5.5.0"))
+//        .package(url: "https://github.com/nathantannar4/InputBarAccessoryView", .upToNextMajor(from: "5.5.0"))
+        .package(url: "https://github.com/nathantannar4/InputBarAccessoryView", .branch("master"))
```
Kaspik: Should cut release of IBAV before merging this probably? And drop iOS 12 too there

martinpucik: Sounds good 👍

Kaspik: We can now point to 6.0.0 release

Kaspik (approved):

PR #1715, opened 2022-05-10
Fix iOS 13 depraction, fix Swiftlint and Danger on PRs

Kaspik:
- add valid swiftlint config
- fix Danger and swiftlint on PRs
- fix iOS 13 showMenu depraction


martinpucik (approved):
LGTM! 🚀

PR #1737, opened 2022-08-29
DocC Documentation

maxxfrazer:
- Adds auto-deployment of documentation on a branch `gh-pages` root when pushed to master.

No

Any other comments?
-------------------
The way DocC is set up, the new homepage of the docs would need to be `https://messagekit.github.io/documentation/messagekit`.
index.html would need to be modified with a redirect to avoid this, but it would be a hack as that's not how DocC currently works.
The automation also pushes InputAccessoryBarView's docs to the following URL:
`https://messagekit.github.io/inputbaraccessoryview/documentation/inputbaraccessoryview/`

Where has this been tested?
---------------------------
Locally for this repository, also on another project I work on, [VideoUIKit-iOS](https://github.com/AgoraIO-Community/VideoUIKit-iOS). This is probably already known, but as the owner and repo in MessageKit's instance are the same, the url base does not need the repo name after github.io.

To test it out locally, run the commands found in `.github/workflows/deploy_docs.yaml`, then launch the server with python for example: `python3 -m http.server`. And open `http://localhost:8000/documentation/messagekit/` in any browser.

Screenshot 2022-08-29 at 10 50 41

Once the next release of DocC and Xcode is out and available via GitHub Actions, this action can be adapted to use it with ease.

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -1,0 +1,32 @@
+name: Deploy DocC
+
+on:
+  push:
+    branches:
+      - "main"
+permissions:
+  contents: write
+jobs:
+  build_docs:
+    runs-on: macos-12
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v3
+
+      - name: Build DocC
+        run: |
+          swift package resolve
+          xcodebuild docbuild -scheme MessageKit -derivedDataPath 'docc' -destination 'generic/platform=iOS';
+          $(xcrun --find docc) process-archive \
+            transform-for-static-hosting docc/Build/Products/Debug-iphoneos/MessageKit.doccarchive \
+            --hosting-base-path '' \
+            --output-path docs
+          $(xcrun --find docc) process-archive \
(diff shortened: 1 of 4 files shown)


Kaspik (approved):

Kaspik:
Wanna add a link from Readme too?

maxxfrazer:
@Kaspik sure will do! Just so you're aware:
- The changes will only take effect if the repo GitHub pages setting is updated as the image below
- And any current links to `https://messagekit.github.io` will go to a blank page because of how DocC works

Screenshot 2022-08-29 at 11 38 42

Kaspik:
We want to get rid of what currently exists on https://messagekit.github.io anyways cause it's not being updated. I'll update the settings for sure.

maxxfrazer:
@Kaspik amazing, thanks!

Kaspik (approved):

Kaspik:
Seems like the URL is different: https://messagekit.github.io/MessageKit/

And it seems like the page doesn't load correctly. 😄

maxxfrazer:
aha, i thought that it had nothing appended after github.io because of the name. need to prepend hosting base path with "MessageKit/" and then it'll come out. I can do it as soon as I'm home (~2hrs)

PR #1792, opened 2023-04-30
Fix typing indicator in advanced example controller

### LINKED ISSUE #1787: Advanced Example crashes when typing indicator is hidden
👻
**Describe the bug**
Advanced Example crashes when typing indicator is hidden

**To Reproduce**
Steps/code to reproduce the behavior:
1. Open the example app
2. Navigate to Advanced Example
3. Typing indicator shows. Upon indicator hide, the app crashes.

**Expected behavior**
The indicator should not crash on hide.

**Screenshots**
Screen Shot 2023-04-10 at 4 23 57 PM

**Environment**
- What version of MessageKit are you using? latest commit
- What version of iOS are you running on? iOS 16.4.1
- What version of Swift are you running on? Swift 5.8
- What device(s) are you testing on? Are these simulators? On Device
- Is the issue you're experiencing reproducible in the example app? Yes, it crashes on the example app

**Additional context**
The problem is partially solved when animation is set to `false`. However, other cases of the app crashing due to improper index still occurs (e.g. sending a message when indicator is visible).

Kaspik:
- Don't animate hiding typing indicator if also inserting a message

Seems like there is a race condition where `hideIndicator` calls `performBatchUpdates` and the update inside calls `insertMessage` which does another `performBatchUpdates` and that crashes.

Fixes https://github.com/MessageKit/MessageKit/issues/1787


martinpucik (approved):
LGTM! 🚀

PR #1793, opened 2023-05-15
Note-Taking Example using MessageKit

vidur-subaiah:
Implements an example project that uses MessageKit to create a note taking application. Showcases how MessageKit can be used for use cases other than Chat UI's. 
…

No.
…

No.

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 14 Pro Simulator

**iOS Version:**  iOS 16.4

**Swift Version:** Swift 5

**MessageKit Version:** MessageKit 4.1.1


Kaspik:
Would love to see a video of the example so I don§t have to checkout :) Also, gitignore should be in the root already, is this another one necessary?

Issue #1788, opened 2023-04-14
Consistent "Invalid number of sections" and "Index out of range" errors when using typing indicator


monstermac77:
**Describe the bug**
All attempts to add a typing indicator to our MessagesViewController have resulted in fatal crashes. We get "Invalid number of sections" or "Index out of range" errors. It seems a number of people have had similar issues with the typing indicator (https://github.com/MessageKit/MessageKit/issues/1387, https://github.com/MessageKit/MessageKit/issues/1787) , despite us all basing our code off of the working example, `AdvancedExampleViewController`.

We are using a custom cell implementation, we believe we have taken the additional steps necessary to handle typing indicators for custom cells ([found here](https://github.com/MessageKit/MessageKit/blob/main/Documentation/CUSTOM_CELLS.md)), namely adding code to return `typingIndicatorSizeCalculator` in our custom cell size calculator and the typing indicator cell itself in the `cellForItemAt` function. 

**To Reproduce**
Take an existing, currently functional, implementation of MessageKit with a custom cell. Then add the following at some point in the view's lifecycle (which attempts to add the typing indicator to the view): 
```
self.setTypingIndicatorViewHidden(false, animated: true)
```
Running it at this stage will obviously crash, as you haven't updated the necessary functions to handle the fact that there is now an extra section that will appear in certain cases. So update these functions:

While the class you've inherited has the `isTypingIndicatorHidden` case in `numberOfSections` handled already, you would have had to implement it in your view controller as well in order to conform to `MessagesDataSource`, so rather than just returning `messages.count`, you'll need to add 1 in the case that the typing indicator is being shown. 
```
func numberOfSections(in messagesCollectionView: MessageKit.MessagesCollectionView) -> Int {
   let sections = messages.count
   return isTypingIndicatorHidden ? sections : sections + 1
}
```

As mentioned above, you'll also need to make sure you return the typing indicator cell when it's being shown, rather than trying to paint a message cell. 
```
override func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
        guard let dataSource = messagesCollectionView.messagesDataSource else{fatalError("No data for messages.")} 
        if isSectionReservedForTypingIndicator(indexPath.section) {
            return dataSource.typingIndicator(at: indexPath, in: messagesCollectionView)
        }

      ....

}
```

[...]

PR #1804, opened 2023-07-10
Fix hiding typing indicator crash.

Zandor300:
👻

Set typing indicator hidden inside performBatchUpdates to prevent flow layout from requesting message type from DataSource for typing indicator section.

Showing the typing indicator works fine, but when hiding it, a crash occurs in the following function of `MessagesDataSource` with message `index out of bounds`:
```swift
    func messageForItem(at indexPath: IndexPath, in messagesCollectionView: MessagesCollectionView) -> MessageType {
        return messages[indexPath.section]
    }
```

The stacktrace indicates this is coming from the following function in `MessagesCollectionViewFlowLayout`:
```swift
  open func cellSizeCalculatorForItem(at indexPath: IndexPath) -> CellSizeCalculator {
    if isSectionReservedForTypingIndicator(indexPath.section) {
      return typingIndicatorSizeCalculator
    }
    let message = messagesDataSource.messageForItem(at: indexPath, in: messagesCollectionView) // Here
    switch message.kind {
// ...
```

This is because `messagesCollectionView.performBatchUpdates()` causes the `MessagesCollectionViewFlowLayout` to update. But because we have already set the typing indicator to `isHidden = true` before calling `performBatchUpdates()`, the call to `isSectionReservedForTypingIndicator` in flow layout will return false. This causes a call to `messageForItem` for a section dedicated to the typing indicator, which causes the `index out of bounds` crash since we are not expecting the library to request anything for this section.

When I made the change in this PR to my local codebase, it fixed the issue and the typing indicator hiding worked properly. This was tested within the app I am using MessageKit.

I've seen the example app crash a couple of times aswell and I think this might be because of the same issue. Specifically the advanced example.

Found this crash myself while developing an app that is using MessageKit.

Possible fix for https://github.com/MessageKit/MessageKit/issues/1788

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:**
- iPhone 11 Pro
- iPhone 12 Pro

**iOS Version:** 
- iOS 15.4
- iOS 16.0

**Swift Version:** Latest

**MessageKit Version:** 4.1.1

### DIFF
--- a/Sources/Controllers/MessagesViewController+TypingIndicator.swift
+++ b/Sources/Controllers/MessagesViewController+TypingIndicator.swift
@@ -50,14 +50,15 @@ extension MessagesViewController {
     }
 
     let section = messagesCollectionView.numberOfSections
-    messagesCollectionView.setTypingIndicatorViewHidden(isHidden)
 
     if animated {
       messagesCollectionView.performBatchUpdates({ [weak self] in
+        self?.messagesCollectionView.setTypingIndicatorViewHidden(isHidden)
         self?.performUpdatesForTypingIndicatorVisability(at: section)
         updates?()
       }, completion: completion)
     } else {
+      messagesCollectionView.setTypingIndicatorViewHidden(isHidden)
       performUpdatesForTypingIndicatorVisability(at: section)
       updates?()
       completion?(true)


Zandor300:
@Kaspik Are you the right person to tag for this? If so, could you take a look at my PR?

Kaspik (approved):

Issue #1734, opened 2022-08-10
InputBarAccessoryView slides to top on device with SwiftUI blocking messages


e1block:
**Describe the bug**

When you come to enter a message, the `InputBarAccessoryView` pushes up so that the keyboard can be shown. This works correctly in the Simulator, but not on device. The messages get blocked in behind. See screenshot attached.

**To Reproduce**

Run the Example app on device, select **SwiftUI Example**, try to send a message.

**Expected behavior**

The messages are shown _above_ the `InputBarAccessoryView` like they do with the other examples or when running in the Simulator.

**Screenshots**

![Screen Shot 2022-08-10 at 12 46 37 PM](https://user-images.githubusercontent.com/104395039/183981210-56f7a400-5f96-4d33-9f61-0981f18f8f86.png)

**Environment**
- What version of MessageKit are you using? **Latest main branch** - https://github.com/MessageKit/MessageKit/commit/14dd6aa335e76c46a06db7e9ee5c1d62a0ccca6a
- What version of iOS are you running on? **15.4.1**
- What version of Swift are you running on? **Swift 5**
- What device(s) are you testing on? Are these simulators? **iPhone 11 Pro (device), iPhone 13 Pro (simulator)**
- Is the issue you're experiencing reproducible in the example app? **Yes**

e1block:
It actually occurs in the simulator too when you have the software keyboard actually showing (setting). From what I can tell, the issue is in `MessagesViewController.updateMessageCollectionViewBottomInset()`. In the case of Swift, the `collectionViewHeight` needs to be the parent view since there's this wrapper structure between SwiftUI & UIKit. Still debugging.

martinpucik:
Thanks for raising this issue @e1block and trying to debug this. SwiftUI support is experimental and the main focus is for UIKit based applications.

e1block:
For sure. Nonetheless, any insights would be appreciated.

e1block:
Looks like the root of the bug is in [InputBarAccessoryView](https://github.com/nathantannar4/InputBarAccessoryView). The SwiftUI example there also has an incorrect offset as compared to the UIKit examples.

Kaspik:
Closing in favor of the same issue in IBAV.

e1block:
Figured it out.

[...]

PR #1807, opened 2023-08-23
Fix for SwiftUI example IBAV position issues (example app)

Janneman84:
This fixes IBAV position issues in the SwiftUI example of the example app by adding `ignorersSafeArea` modifier to the body.

I also fixed a small warning in the example app while I was at it.

https://github.com/MessageKit/MessageKit/issues/1734 and a couple of others probably

Any other comments?
-------------------
The fix also needs IBAV 6.3.0 to work properly, which was added to MessageKit today. Hence I send this PR today.

Also iOS 13 users are out of luck..., but it forced me to do a iOS version check which made things a bit ugly.

Where has this been tested?
---------------------------
iPhone and iPad simulator and iPhone Xs with iOS 16
MessageKit 4.2.0

### DIFF
--- a/Example/Sources/View Controllers/ChatViewController.swift
+++ b/Example/Sources/View Controllers/ChatViewController.swift
@@ -333,7 +333,7 @@ extension ChatViewController: InputBarAccessoryViewDelegate {
 
       let substring = attributedText.attributedSubstring(from: range)
       let context = substring.attribute(.autocompletedContext, at: 0, effectiveRange: nil)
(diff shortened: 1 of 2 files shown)


Janneman84:
Script says it can't find my commit, did I do something wrong?

`RuntimeError: Commit 7809a5b8 doesn't exist.`

Kaspik (approved):

Tomas-Shao:
not works from my side. 
see the issue i created today: https://github.com/MessageKit/MessageKit/issues/1861

PR #1821, opened 2023-11-14
Added ability to specify additionalBottomSpace for keyboardManager

Almaz5200:
Creates an open method that allows to specify withAdditionalBottomSpace argument for InputBarAccessoryView's bind method. This is necessary when MessageView does not occupy full screen and there is something on the bottom (in our case banner view)

I couldn't find any

Where has this been tested?
---------------------------
**Devices/Simulators:** iPhone 15 Pro, iPhone 15 Pro simulator

**iOS Version:** 17

**Swift Version:** 5.9

**MessageKit Version:** main branch

### DIFF
--- a/Sources/Controllers/MessagesViewController+Keyboard.swift
+++ b/Sources/Controllers/MessagesViewController+Keyboard.swift
@@ -31,7 +31,10 @@ extension MessagesViewController {
   // MARK: - Register Observers
 
   internal func addKeyboardObservers() {
-    keyboardManager.bind(inputAccessoryView: inputContainerView)
+    keyboardManager.bind(
+      inputAccessoryView: inputContainerView, 
+      withAdditionalBottomSpace: { [weak self] in self?.inputBarAdditionalBottomSpace() ?? 0 }
+    )
     keyboardManager.bind(to: messagesCollectionView)
 
     /// Observe didBeginEditing to scroll content to last item if necessary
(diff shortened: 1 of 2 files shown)


fukemy (approved):
look good

Kaspik (approved):

vvyskocil:
It seems there is a issue with additionalBottomSpace for keyboardManager, I override the method to take reserve some space for a information banner at the top of the MessageView list, when I start typing text in the inputBar the virtual keyboard is raised and the inputBar is going up with it, this is ok. But as soon as I touch the message list to scroll it, the inputBar disappear and the bottom of the message list is under the virtual keyboard.

IMG_0685

After scrolling the message list:

IMG_0686

@Almaz5200 I'm missing something ?

Almaz5200:
If your message is at the top I don't think you need to override that method, if I understand your case correctly. I'd try just removing this override and checking how is it going to behave afterwards, both with keyboard open and closed

Almaz5200:
My initial motivation for this was reserving space at the bottom of the screen, bellow the inputBar

vvyskocil:
@Almaz5200 thank you for the quick response! I removed the override, then the inputBar is always hidden by the keyboard... When the override is in place at least the inputBar is at the right place until I touch the message list.

Almaz5200:
In that case I am not sure what could be the reason, I don't think it is related to that override. In our case we too reserve some space at the top, but it didn't require that override to accomplish

Issue #1872, opened 2024-11-22
Crash in LocationMessageCell when input a invalid  CLLocation


xiaoxiaowesley:
**Describe the bug**
Crash in LocationMessageCell when input a invalid  CLLocation

**To Reproduce**
Steps/code to reproduce the behavior:

Take the "Advanced Example" for example.
1. if I force to set a invalid CLLocation (latitude: 120.20000, longitude: -122.0312)

SampleData.swift
```swift 
    let locations: [CLLocation] = [
        CLLocation(latitude: 120.20000, longitude: -122.0312),
    ]
```

2. It will crash when build  LocationMessageCell 
![image](https://github.com/user-attachments/assets/caf4a1ee-bc55-46fd-b291-24333fc0ae5c)

here is the log

```
2024-11-21 20:47:00.428457+0800 ChatExample[2237:229178] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid Region <center:+120.20000000, +30.26667000 span:+10.00000000, +10.00000000>'
*** First throw call stack:
(0x19f51d9d8 0x1b3886b54 0x19f42e6e4 0x1aeb025f4 0x102a0822c 0x1029c7780 0x1029c7d2c 0x1a1506b24 0x1a1522fd0 0x1a151fee4 0x1a15266d4 0x1a22837b4 0x1a1525714 0x1a151ecbc 0x1a1527970 0x1a1527da4 0x1a1527ba4 0x1a1527b10 0x1a1527a94 0x1a15279d0 0x1028ad74c 0x1028abedc 0x1028bb954 0x1028bc714 0x1a0814430 0x19f49e3e0 0x19f49dfe4 0x19f49d4c4 0x19f497850 0x19f496ba0 0x1b61d4598 0x1a1d863d8 0x1a1d8b958 0x1028c34fc 0x19f175568)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

**Expected behavior**

No crash 

**Environment**
- What version of MessageKit are you using?

main branch

- What version of iOS are you running on?

18.1

- What version of Swift are you running on?

swift 5

- What device(s) are you testing on? Are these simulators?

iphone12 pro

- Is the issue you're experiencing reproducible in the example app?

Yes. But need to modify to fore it reproducible

**Additional context**

Kaspik:
Can you please elaborate how you even get this to crash in real life?

xiaoxiaowesley:
[...]

PR #1873, opened 2024-11-22
Throw fatalError when the coordinate is invalid

xiaoxiaowesley:

Throw fataError when the  coordinate is invalid in LocationMessageCell

YES .  https://github.com/MessageKit/MessageKit/issues/1872

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Devices/Simulators:** …

iPhone12 pro

**iOS Version:** …

18.1

**Swift Version:** …

swift 5 

**MessageKit Version:** …

main branch

### DIFF
--- a/Sources/Views/Cells/LocationMessageCell.swift
+++ b/Sources/Views/Cells/LocationMessageCell.swift
@@ -72,6 +72,8 @@ open class LocationMessageCell: MessageContentCell {
       in: messagesCollectionView)
 
     guard case .location(let locationItem) = message.kind else { fatalError("Configuring LocationMessageCell with wrong") }
+      
+    guard CLLocationCoordinate2DIsValid(locationItem.location.coordinate) else { fatalError("Coordinate is not valid") }
 
     activityIndicator.startAnimating()
 


Sources/Views/Cells/LocationMessageCell.swift
```diff
@@ -72,6 +72,8 @@ open class LocationMessageCell: MessageContentCell {
       in: messagesCollectionView)
 
     guard case .location(let locationItem) = message.kind else { fatalError("Configuring LocationMessageCell with wrong") }
+      
+    guard CLLocationCoordinate2DIsValid(locationItem.location.coordinate) else { fatalError("Coordinate is not valid") }
```
Kaspik, line 76: How is this different if it also crashes?

xiaoxiaowesley: There really is no difference. I didn't find any similar error View in the project.
So I want to discuss with you whether we should add a View to display this kind of error.

Kaspik: Solved in my Swift6 PR - https://github.com/MessageKit/MessageKit/commit/3a0f20b96a0a91ac51fb9b1249636816a4da08db

PR #1915, opened 2026-08-28
Build the example app in the Swift 6 language mode

martinpucik:
> Stacked on #1914 → #1913 → #1911 → #1910 → #1909. Base is `chore/remove-dead-files`.

## Summary

The package declares `swiftLanguageModes: [.v6]`, but the example project still set `SWIFT_VERSION = 5.0`. CI built the example with no strict concurrency checking, so it validated none of the surface a consumer migrating to Swift 6 actually meets.

## It found a real library defect

Three of MessageKit's delegate protocols carry `@MainActor`:

| Protocol | `@MainActor` before | after |
| --- | --- | --- |
| `MessagesDataSource` | yes | yes |
| `MessagesLayoutDelegate` | yes | yes |
| `MessagesDisplayDelegate` | yes | yes |
| **`MessageCellDelegate`** | **no** | **yes** |
| **`MessageLabelDelegate`** | **no** | **yes** |

Both missing ones exist purely to handle taps inside cells, and the library only ever calls them from UIKit classes that are already main-actor isolated (`MessageLabel`, `MessagesCollectionView`, `MessageContentCell`, `TextMessageCell`). Every consumer building in Swift 6 had to work around this with `@preconcurrency`.

I annotated the protocols rather than papering over it in the example — a `@preconcurrency` there would have hidden the exact problem that building the example in Swift 6 is meant to surface.

## The example app's own migration

- **`AlertService` and `SampleData` become `@MainActor`.** One presents alerts, the other holds `lazy var` state that only view controllers read.
- **InputBarAccessoryView conformances marked `@preconcurrency`,** matching the existing treatment of `AVAudioPlayerDelegate` in `BasicAudioController`. Those protocols carry no isolation of their own, so this is the right tool there — unlike MessageKit's own protocols, which I fixed at the source.
- **`DispatchQueue.global` → `DispatchQueue.main` hops become `Task`** in the three example view controllers. They were passing non-Sendable arrays between queues for no benefit: the mock data generator is synchronous, and the hops existed only to fake latency, which `Task.sleep` expresses without crossing an isolation boundary.
- **`ChatExampleTests` becomes `@MainActor`,** since it reads `SampleData`.

## The deployment target is deliberately left at 15.0

I tried lowering it to 14.0 to match `Package.swift`. It does not build:

```
The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 14.0,
but the range of supported deployment target versions is 15.0 to 27.0.x.
```

Dependencies are not the blocker — Kingfisher 8.11.0 supports iOS 13 and InputBarAccessoryView 7.0.6 supports iOS 14. The current Xcode simply will not build an app below iOS 15.

**This raises a question worth deciding separately:** `Package.swift` advertises `platforms: [.iOS(.v14)]`, but no app on a current toolchain can target iOS 14, so that claim cannot be exercised by the example or by CI. Whether to raise the package minimum to iOS 15 is a breaking change and your call — I have not touched it.

## Verification

Builds clean for testing, zero errors. The only warnings are two pre-existing XCTest linkage notes about iOS 15 vs 17, unrelated to this change and present before it. `make lint` passes.

## One cosmetic note

SwiftFormat's `markTypes` rule regenerates `// MARK:` headers from the conformance list, so `@preconcurrency` now leaks into four section comments, e.g. `// MARK: @preconcurrency AttachmentManagerDelegate`. That is the formatter's canonical output — hand-editing fails `make lint`. Suppressing it means changing the SwiftFormat config, which I left alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Fixed
 
+- **Breaking Change** Annotate `MessageCellDelegate` and `MessageLabelDelegate` with `@MainActor`. Their sibling delegate protocols `MessagesDataSource`, `MessagesLayoutDelegate` and `MessagesDisplayDelegate` already carried the annotation, so consumers building in the Swift 6 language mode had to work around these two with `@preconcurrency` by [@martinpucik](https://github.com/martinpucik)
 - Fix `make format`, `make lint` and the pre-commit hook, which still called the removed SwiftPM plugins by [@martinpucik](https://github.com/martinpucik)
 - Adopt the scene lifecycle in the example app, which crashed on launch on iOS 26 and later by [@martinpucik](https://github.com/martinpucik)
 - Add the missing camera and photo library usage descriptions to the example app, which crashed when you opened the camera by [@martinpucik](https://github.com/martinpucik)
@@ -23,6 +24,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 ### Changed
 
 - Enforce `make lint` in CI and reformat the sources it flagged by [@martinpucik](https://github.com/martinpucik)
+- Build the example app in the Swift 6 language mode, so that CI exercises the same strict concurrency checking as the library, and replace its `DispatchQueue` hops with structured concurrency by [@martinpucik](https://github.com/martinpucik)
 
 - Migrate Danger from Ruby to Danger Swift and drop the `Gemfile` by [@martinpucik](https://github.com/martinpucik)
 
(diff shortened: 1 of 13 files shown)


martinpucik:
Pushed `0cae27d`: the UI test target also needed `@MainActor`.

`SWIFT_VERSION` is set once at the project level, so raising it to 6.0 applied to `ChatExampleUITests` too, and XCUITest's API is main-actor isolated — every `XCUIApplication` call failed to compile.

My earlier "builds clean" claim was scoped too narrowly: I built the `ChatExample` scheme, which covers the app and its unit tests but **not** the UI test target. That lives in the separate `ChatExampleUITests` scheme, which `make test_example` runs. CI caught what my local check could not reach. Both schemes now build for testing locally.

Kaspik (approved):

PR #1917, opened 2026-08-29
Add tests for MessageSizeCalculator

martinpucik:
> Stacked on #1915 → #1914 → #1913 → #1911 → #1910 → #1909. Base is `chore/example-matches-library`.

## Summary

`MessageSizeCalculator` (356 lines) decides how tall every cell is and how wide its container may be, and **no test referred to it**. A change to that arithmetic could only be caught by looking at the example app.

This is the first slice of the test-coverage item, aimed at the largest untested area rather than at raising a coverage number.

## What is covered

Only the parts that branch — not the parts that read back a stored property:

- **Natural avatar position resolution.** Resolves to trailing for the current sender, leading for everyone else, and an explicit horizontal position survives the resolution untouched.
- **Avatar size precedence.** Falls back to the incoming or outgoing size by sender, and the layout delegate overrides both.
- **Container maximum width.** Subtracts the avatar, container padding, accessory view, accessory padding and avatar padding from the item width.
- **Cell height.** Takes the tallest of avatar, labels and accessory view, and counts the vertical container padding.
- Container padding and cell label alignment follow the sender; the base calculator reports an empty container.

## These tests were checked for detection power

A passing test proves nothing on its own, so I ran them against a deliberately broken calculator:

| Mutation | Result |
| --- | --- |
| Invert the natural position resolution (`isFromCurrentSender ? .cellLeading : .cellTrailing`) | 2 tests fail |
| Drop `avatarLeadingTrailingPadding` from the max-width sum | 1 test fails |

Both mutations were caught, then reverted. The assertions detect regressions rather than only describing current behaviour.

## Verification

Full suite passes: **72 tests, 0 failures**, up from 60. `make lint` passes.

## Note on the fixtures

`StubLayoutDelegate` only stubs `avatarSize`; everything else comes from the protocol's own defaults, so the tests exercise the real fallback paths rather than a fully mocked delegate. `SUT` is a `@MainActor` struct because `MockMessagesDataSource` is main-actor isolated.

## Still uncovered

`TypingIndicator` (162 lines) also has no tests, and the other calculators (`Text`, `Media`, `Location`, `Audio`) inherit from this one without direct coverage. Worth a follow-up if you want to keep going.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 ### Added
 
 - Add tests for `MessageSizeCalculator`, which covers the avatar position resolution, the sender specific avatar sizes and paddings, the container maximum width and the cell height, and had no tests before by [@martinpucik](https://github.com/martinpucik)
+- Add tests for `TypingIndicator` and `TypingBubble`, which cover the dot layout and spacing, the animation state flags, the animation layers and the pulse layers, and had no tests before by [@martinpucik](https://github.com/martinpucik)
 ### Fixed
 
 - **Breaking Change** Annotate `MessageCellDelegate` and `MessageLabelDelegate` with `@MainActor`. Their sibling delegate protocols `MessagesDataSource`, `MessagesLayoutDelegate` and `MessagesDisplayDelegate` already carried the annotation, so consumers building in the Swift 6 language mode had to work around these two with `@preconcurrency` by [@martinpucik](https://github.com/martinpucik)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The changelog for `MessageKit`. Also see the [releases](https://github.com/Messa
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


martinpucik:
Pushed `58b40d4`: added **`TypingIndicator` and `TypingBubble` tests** to this PR. Neither had any test.

**Covered:** three dots in a horizontal stack, `dotColor` repainting every dot, the stack filling its bounds, dots spaced only once the view has a width, the animating flag on start/stop, stopping a view that never started being a no-op, the fade and bounce animation layers and their wiring to `bounceOffset`, and the bubble pulsing all three bubbles and clearing them on stop.

**Mutation-checked**, same as the calculator tests. Making the dot spacing unconditional, dropping the sign flip in the bounce animation, and clearing only the content bubble on stop each fail a test.

**One thing worth recording.** I expected `stopAnimating()` to be defeated by its own pending work — `startAnimating()` schedules three `asyncAfter` blocks at 0s, 0.33s and 0.66s, and none of them re-checks `isAnimating`, so stopping immediately should leave two dots animating. I probed it and **it does not reproduce**: all three dots are clean afterwards. Core Animation does not keep an animation on a layer with no window once the transaction commits, which also means `animationKeys()` cannot be used to assert the indicator's animations after a delay. That is why those assertions go through the factory properties instead.

Whether it misbehaves on-screen, where the layers are rendered, I have not verified — it would need a hosted UI test. Flagging rather than claiming either way.

Suite is now **87 tests, 0 failures**, up from 60 at the start.

PR #1920, opened 2026-09-03
Normalize the copyright headers

martinpucik:
## Summary

The copyright header had drifted into six variants:

| Header | Files |
| --- | --- |
| `2017-2019` | 68 |
| `2017-2020` | 34 |
| `2017-2022` | 17 |
| `2017-2018` | 3 |
| `2017-2026` | 1 |
| `2017` | 1 |

Three example app files also carried a stray leading `//`, and `Package.swift` and `Dangerfile.swift` had lost two separator lines from the middle of the licence text.

**Nothing enforced the header**, which is how it drifted: the `fileHeader` rule is absent from `.swiftformat`'s `--rules` list, and SwiftLint excludes `Tests` and `Example` — where most of the variants sat.

This sets the **115 files that carry MessageKit's own MIT header** to one header, byte for byte, dated `2017-2026` after `Dangerfile.swift`, the most recently added file.

## Ten files are deliberately left alone

They credit someone other than MessageKit:

- `Example/Tests/ChatExampleTests.swift` and `Example/UITests/ChatExampleUITests.swift` carry `Copyright (c) 2016-present Jesse Squires`, from JSQMessagesViewController, the project MessageKit grew out of.
- Eight example app files carry Xcode template headers naming their own author and year (`Copyright © 2021 MessageKit. All rights reserved.` and similar).

The MIT licence requires the copyright notice to travel with the code, so these are **attribution, not formatting**.

## Why not SwiftFormat's `fileHeader` rule

This was the obvious way to enforce it, and it is unsafe here. Pointing `--header` at the MIT text rewrites every file the rule sees. I ran it locally against a copy of one of the example app headers:

```
# before
//  Thing.swift
//  Created by Some Contributor on 01/05/21.
//  Copyright © 2021 MessageKit. All rights reserved.

# after --header "// MIT License\n//\n// Copyright (c) 2017-2026 MessageKit"
// MIT License
//
// Copyright (c) 2017-2026 MessageKit
```

The author credit is gone. Only `--exclude` protects a file, and that drops it from every other SwiftFormat rule as well. (A top-of-file `// swiftformat:disable fileHeader` does work, but it leaves a directive as the first line of ten files.)

## Enforcement through Danger instead

`Dangerfile.swift` now checks that a **newly added** Swift file has a copyright line somewhere in its first 25 lines — and never what that line says. A contributor keeping their own header passes; a file with no header at all gets flagged. Danger already runs on every PR, only warns, and cannot rewrite anything.

## Verification

- **Every line in the diff is a comment.** `git diff -U0 | grep -E '^[+-]' | grep -vE '^[+-]\s*//'` returns nothing but the CHANGELOG and Dangerfile additions.
- None of the 10 attribution files appear in the diff.
- All 115 normalized headers hash identically.
- `Package.swift` keeps `// swift-tools-version:6.0` as its first line.
- `make lint` passes. **87 tests pass.**

## Note

This overlaps #1919 only in spirit, not in content — that branch's 9 new files do not exist on `main`, and they already carry `2017-2026`, so the two merge in either order without conflict.


martinpucik:
## Enforcement moved to SwiftFormat

Replaced the Danger check with SwiftFormat's `fileHeader` rule in `5c50be864`.

**The Danger check was broken anyway.** Danger Swift's `warn` needs the `message:` label when given a file and line, so the Danger job failed above. `swiftc -parse` missed it because parsing never resolves the `Danger` module.

SwiftFormat is the better tool here regardless: it enforces the **exact text** rather than merely the presence of the word "copyright", and it repairs a wrong header instead of asking the contributor to.

### Three findings worth recording

1. **The licence text breaks the config parser.** It contains `(the "Software")` and `"AS IS"`, and a quoted config value ends at the first inner quote. Unescaped, `--header` truncates mid-licence and then *every* file fails the rule — including files whose header is already correct. The inner quotes are escaped as `\"`.
2. **`// swiftformat:disable fileHeader` only works on line 1.** Placed below the header, the header is replaced anyway. Verified both ways.
3. **A comment separated from the header by a blank line survives the rule.** This is what makes it possible to keep an author credit.

### What that allows

The 8 example app files with Xcode template headers now carry the licence, with their `Created by` line kept as a comment beneath it:

```swift
// MIT License
// ...
// SOFTWARE.

// Created by Vignesh J on 01/05/21.

import UIKit
```

Their copyright line already read `Copyright © 2021 MessageKit. All rights reserved.` — the holder was never the individual, and `All rights reserved` contradicted the licence the project ships under.

The 2 files whose notice reads `Copyright (c) 2016-present Jesse Squires` are genuinely third-party. They keep it, with the directive on line 1 and a comment explaining why.

### Verified by hand

| Check | Result |
| --- | --- |
| A file with no header | fails `make lint` |
| `make format` on the tree | 0/124 changed — idempotent |
| The 10 credited files | all still name who they credit |
| Source changes in the new commit | every line is a comment |
| Tests | 87 pass |

The `fileHeader` year is hardcoded rather than SwiftFormat's `{year}` token on purpose — `{year}` would rewrite every file each January and break `make lint` until someone ran `make format`.

